PR #24, opened 2022-06-28
Support subdirectories

ggjp:
Summary:
- In `denote-directory-files`, I replaced `directory-files` with `directory-files-recursively`.
- The new function `denote--file-relative-to-denote-directory` is used to fix the backlinks buffer and other parts of the code. You should now be able to see files like "subdir/20220628T111111--Note__tag.org" in the backlinks buffer.

Tests I have made:
- Backlinks buffer displays links in subdirectories. Clicking on them works.
- `denote-link` allows navigation in subdirectories (and only displays actual notes like before).
- Renaming a note still works. Inferred tags are all found.

Still to do:
- Test other parts of the code, notably `denote-link-ol-export` and `denote-link--ol-resolve-link-to-target`.
- Fix fontification for the backlinks buffer. There is a comment in the code saying that enabling `dired-denote-mode` might not be the best way to enable fontification. I don't know how it works. I have just commented that code. A temporary solution to not break existing setups is to enable `dired-denote-mode` only if there are no subdirectories and defer that feature for later.
- Review `denote-link--expand-identifiers`. There might be a better way to do this.

Question:
- I have been thinking: What would a user option `denote-excluded-subdirectories` actually be useful for? I mean the command `denote-link` already ignores anything that is not a note. I expect others commands to behave similarly (ignore non-notes). So, excluded subdirectories are already supported!

(I will notify you when I get the final confirmation from the FSF.)

### DIFF
--- a/denote-link.el
+++ b/denote-link.el
@@ -312,9 +312,13 @@ format is always [[denote:IDENTIFIER]]."
 
 (defun denote-link--expand-identifiers (regexp)
   "Expend identifiers matching REGEXP into file paths."
-  (delq nil (mapcar (lambda (i)
-                      (file-name-completion i (denote-directory)))
-                    (denote-link--collect-identifiers regexp))))
+  (let ((files (denote--directory-files))
+        (found-files))
+    (dolist (file files)
+      (dolist (i (denote-link--collect-identifiers regexp))
+        (if (string-prefix-p i (file-name-nondirectory file))
+            (push file found-files))))
+    found-files))
 
 (defvar denote-link--find-file-history nil
   "History for `denote-link-find-file'.")
@@ -442,16 +446,16 @@ Use optional TITLE for a prettier heading."
                   (l (length heading)))
         (insert (format "%s\n%s\n\n" heading (make-string l ?-))))
       (mapc (lambda (f)
-              (insert (file-name-nondirectory f))
+              (insert f)
               (make-button (point-at-bol) (point-at-eol) :type 'denote-link-backlink-button)
               (newline))
             files)
-      (goto-char (point-min))
+      (goto-char (point-min)))
       ;; NOTE 2022-06-15: Technically this is not Dired.  Maybe we
       ;; should abstract the fontification into a general purpose
       ;; minor-mode.
-      (when denote-link-fontify-backlinks
-        (denote-dired-mode 1)))
+      ;(when denote-link-fontify-backlinks
+        ;(denote-dired-mode 1)))
     (denote-link--display-buffer buf)))
 
 ;;;###autoload
--- a/denote-retrieve.el
+++ b/denote-retrieve.el
@@ -90,7 +90,8 @@ Optional GROUP is a regexp construct for
 
 (defun denote-retrieve--read-file-prompt ()
   "Prompt for regular file in variable `denote-directory'."
-  (read-file-name "Select note: " (denote-directory) nil nil nil #'denote--only-note-p))
+  (read-file-name "Select note: " (denote-directory) nil nil nil
+                  #'(lambda (f) (or (denote--only-note-p f) (file-directory-p f)))))
 
 (defun denote-retrieve--files-in-output (files)
   "Return list of FILES from `find' output."
@@ -109,14 +110,14 @@ The xrefs are returned as an alist."
 Parse `denote-retrieve--xrefs'."
   (sort
    (mapcar (lambda (x)
-             (file-name-nondirectory (car x)))
+             (denote--file-name-relative-to-denote-directory (car x)))
            xrefs)
    #'string-lessp))
 
 (defun denote-retrieve--proces-grep (identifier)
   "Process lines matching IDENTIFIER and return list of files."
   (let* ((default-directory (denote-directory))
-         (file (file-name-nondirectory (buffer-file-name))))
+         (file (denote--file-name-relative-to-denote-directory (buffer-file-name))))
     (denote-retrieve--files-in-output
      (delete file (denote-retrieve--files-in-xrefs
                    (denote-retrieve--xrefs identifier))))))
--- a/denote.el
+++ b/denote.el
@@ -310,6 +310,13 @@ FILE is relative to the variable `denote-directory'."
        (string-match-p (concat "\\b" denote--id-regexp) file)
        (not (string-match-p "[#~]\\'" file))))
 
+(defun denote--file-name-relative-to-denote-directory (file)
+  "Return file name of FILE relative to the variable `denote-directory'.
+FILE must be an absolute path."
+  (if (and (file-name-absolute-p file)
+           (string-prefix-p (denote-directory) file))
+    (substring-no-properties file (length (denote-directory)))))
+
 (defun denote--current-file-is-note-p ()
   "Return non-nil if current file likely is a Denote note."
   (and (or (string-match-p denote--id-regexp (buffer-file-name))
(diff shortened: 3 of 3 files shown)


protesilaos:
Just merged your PR into the `subdir-support` branch.  Makes it easier
for me to test and iterate on it.  Because this is a dev branch, we do
not have to wait for the FSF copyright assignment: we just need it when
we want to merge into `main`.

https://github.com/protesilaos/denote/tree/subdir-support

protesilaos:
Hello again!

I made some tweaks on top of your changes.  Check the commits in the
`subdir-support` branch.  Some comments:

> The new function `denote--file-relative-to-denote-directory`

I rewrote it to (i) call the `(denote-directory)` only once, (ii) return
the file unchanged if it is not absolute instead of returning nil.  This
made the backlinks' buffer work with my existing notes, which are in a
flat directory.

> Backlinks buffer displays links in subdirectories. Clicking on them
> works.

Can confirm.

> `denote-link` allows navigation in subdirectories (and only displays
> actual notes like before).
>
> Renaming a note still works. Inferred tags are all found.

I have not tested those yet, but expect them to work, given the current
state of the code.  Will try them later.

> Test other parts of the code, notably `denote-link-ol-export` and
> `denote-link--ol-resolve-link-to-target`.

On my to-do as well.

> Fix fontification for the backlinks buffer.

I think it is time to abstract the `denote-dired-mode`.  Or, at the very
least, tweak the fontification rules to work with subdirectories as
well.

> Review `denote-link--expand-identifiers`. There might be a better way
> to do this.

All I did to it was replace the `if` with a `when`.  The former is not
wrong, but the latter is easier to read.

I suspect there is a more efficient way to write this, but it works for
now.

> I have been thinking: What would a user option
> `denote-excluded-subdirectories` actually be useful for? I mean the
> command `denote-link` already ignores anything that is not a note. I
> expect others commands to behave similarly (ignore non-notes). So,
> excluded subdirectories are already supported!

I tweaked `denote--only-note-p` to match files ending in `org`, `md`,
`txt`.  I think this should suffice now and we do not need to exclude
anything else.  We can keep this idea in our mind though, in case some
user shows a concrete case where it is needed.

> (I will notify you when I get the final confirmation from the FSF.)

Looking forward to it!

ggjp:
> (ii) return
the file unchanged if it is not absolute instead of returning nil. This
made the backlinks' buffer work with my existing notes, which are in a
flat directory.

I am confused. For me, the `file` argument looks like `/home/user/denote/some-file.org` and `(denote-directory)` returns `/home/user/denote/`, which means the else part can never be reached and `substring` always returns non-nil. I am wondering what is happening on your side?

Also, I have pushed a commit on my fork above your changes to fix `denote-link--ol-resolve-link-to-target`. See my other pull request.

protesilaos:
> I am confused.

Maybe I made a mistake: you are welcome to review it.  Before, the `denote-retrieve--files-in-xrefs` would throw me an error.  With the change I made, things work the way they did before, with the addition of the subdir support.  This still needs to be tested further though.

Now I will check the fontification part in the backlinks' buffer.

> See my other pull request.

Thank you!  I merged it.

protesilaos:
> Now I will check the fontification part in the backlinks' buffer.

A status update on this:

* I put all the faces in denote-faces.el.
* Tweaked the fontification rules.
* Streamlined how `denote-dired-mode` and the backlinks' buffer use
  those rules.

ggjp:
Nice! Thanks.

ggjp:
I am letting you know that I have not heard from the FSF yet. I expect it will not be before Tuesday.

protesilaos:
> I am letting you know that I have not heard from the FSF yet. I expect
> it will not be before Tuesday.

Thank you!  We will have to wait a bit longer then.  That's okay.

By the way, I got some feedback about the branch we are working on:
https://lists.sr.ht/~protesilaos/denote/%3C8735fk4y1w.fsf@hallac.net%3E#%3Cm2czeo9n6e.fsf@fastmail.com%3E

Will check those issues later today.

ggjp:
Ok, I have checked the comments. The only issue I was able to reproduce is the one regarding `id:` notes. Clicking on them does not work for me either. I will check this tomorrow!

protesilaos:
You fixed that one.  The other about renaming, I cannot reproduce.  Probably the file was not saved before, so only the buffer existed.

protesilaos:
The fix for the backlinks' buffer inside a subdir note is probably this, but I am waiting for confirmation on the mailing list:

```diff
 denote-link.el | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/denote-link.el b/denote-link.el
index b1c9795..5de8dab 100644
--- a/denote-link.el
+++ b/denote-link.el
@@ -468,7 +468,7 @@ (defun denote-link-backlinks ()
 default, it will show up below the current window."
   (interactive)
   (let* ((default-directory (denote-directory))
-         (file (file-name-nondirectory (buffer-file-name)))
+         (file (buffer-file-name))
          (id (denote-retrieve--filename-identifier file))
          (title (denote-retrieve--value-title file)))
     (if-let ((files (denote-retrieve--proces-grep id)))
```

ggjp:
I have received confirmation from the FSF! I did not expect it on a Saturday!
You can close this pull request (since it references an old branch) and merge your subdir-support branch when you want.

Also, I agree with your fix about the backlinks!

ggjp:
Also, just a comment to tell you that currently the `denote` command to create new notes creates them in the root of the `denote-directory`. There are many options if we want to change this, but I am not sure which one would be the best. I think we can keep it unchanged for now, but it might be worth mentionning it in the doc of the `denote` command.

shrysr:
I was just thinking about this today and agree with you. I suppose for those who would like to use subdirectories, it may be nice to have starter examples or outlines of possible approaches. 

> Also, just a comment to tell you that currently the denote command to
> create new notes creates them in the root of the denote-directory. There
> are many options if we want to change this, but I am not sure which one
> would be the best. I think we can keep it unchanged for now, but it
> might be worth mentionning it in the doc of the denote command.

protesilaos:
> I have received confirmation from the FSF!

Excellent!  I merged the branch and we can now operate directly on `main`.

> Also, I agree with your fix about the backlinks!

Done!

> Also, just a comment to tell you that currently the denote command to create new notes creates them in the root of the denote-directory. There are many options if we want to change this, but I am not sure which one would be the best. I think we can keep it unchanged for now, but it might be worth mentionning it in the doc of the denote command.

Let's start a new issue about it.  Some ideas that come to mind:

* We already have in the manual samples of `let` binding the `denote-directory`: <https://protesilaos.com/emacs/denote#h:f34b172b-3440-446c-aec1-bf818d0aabfe>.  We can update/repurpose/copy those for a new section that talks about subdirectory support.

* Have a new user option, which will be off by default, that makes all our file-creating commands use the current directory if it is a subdir of `denote-directory`.

protesilaos:
About this PR: I have merged the code on my end and pushed it.  Your commits appear on the `main` branch: <https://github.com/protesilaos/denote/commits/main>.  But I notice that here the web UI does not consider it "merged".  I will see if I can do something about it.

ggjp:
This is referencing an old branch (`subdirectories`) which must not be merged. You don't need to merge anything now. This pull request can be closed.

protesilaos:
> This is referencing an old branch (`subdirectories`)

Gotcha!

> This pull request can be closed.

Thanks!

PR #25, opened 2022-06-29
Fix denote-link--ol-resolve-link-to-target

ggjp:
Clicking on a link in a note should open the linked note, even if it
is in a parent directory or a subdirectory.


protesilaos:
Thank you!

PR #26, opened 2022-06-30
Fix links in non org notes

ggjp:
This fixes links in non org notes.

### DIFF
--- a/denote-link.el
+++ b/denote-link.el
@@ -386,11 +386,12 @@ format is always [[denote:IDENTIFIER]]."
 
 (defun denote-link--find-file-at-button (button)
   "Visit file referenced by BUTTON."
-  (let ((id (denote-link--id-from-string
-             (buffer-substring-no-properties
-              (button-start button)
-              (button-end button)))))
-    (funcall denote-link-buton-action (file-name-completion id (denote-directory)))))
+  (let* ((id (denote-link--id-from-string
+              (buffer-substring-no-properties
+               (button-start button)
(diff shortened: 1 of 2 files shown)


protesilaos:
Merged.  Thank you!

protesilaos:
A general comment here.  The creation of `denote--get-note-path-by-id` is a pattern I prefer over `let` binding the function.  I find it easier to work with.

ggjp:
Yes, I wanted to reuse that code elsewhere, so I extracted it in a function. I think it might also be a useful function in the future as well.

Just so I can learn, how did you know that you needed `(eval-when-compile 'cl-lib)`? From where do you get compiler warnings?

protesilaos:
I use `flymake` (built-in) but the `flycheck` package also does the same.

This is my flymake configuration: https://protesilaos.com/emacs/dotemacs#h:b8bfcc05-c0d3-4093-b3fe-f06187d22c6a

And this is the glue code needed for linting Elisp packages: https://protesilaos.com/emacs/dotemacs#h:d03342da-3035-409f-a5e7-5df1e614d507

The `eval-when-compile` part is something I have learnt through experience.  As far as I can tell, it is only needed at compile time because the `;;;###autoload` does not actually load the file at that time.  At runtime, we do not need `(require 'cl-lib)` because the autoloads work.  Maybe I am wrong though, so please double-check.

ggjp:
Ok, thank you! I will check this out!

PR #28, opened 2022-07-02
Set org-id-extra-files to all files in all subdirectories.

ggjp:
Clicking on an id: link to a note in a subdirectory should work.


protesilaos:
Thank you!

PR #30, opened 2022-07-03
Set default-value to previous title in denote--title-prompt

ggjp:
I have been holding one last pull request until all other would be closed.

When I rename a note, it is often just to fix a typo in the title or change the wording. I rarely have to modify the whole title. This pull request just sets the `DEFAULT-VALUE` parameter of `read-string` in `denote--title-prompt` to the previous title instead of `nil` when using the command `denote-dired-rename-file`.

An empty input will just use the previous title. This would be useful if you just want to update the tags.
When prompted for a title, using `M-n` allows for editing of the previous title.

### DIFF
--- a/denote-dired.el
+++ b/denote-dired.el
@@ -231,10 +231,11 @@ This command is intended to (i) rename existing Denote
 notes, (ii) complement note-taking, such as by renaming
 attachments that the user adds to their notes."
   (interactive
-   (list
-    (denote-dired--rename-file-is-regular (denote-dired--rename-dired-file-or-prompt))
-    (denote--title-prompt)
-    (denote--keywords-prompt)))
+   (let ((file (denote-dired--rename-file-is-regular (denote-dired--rename-dired-file-or-prompt))))
+     (list
+      file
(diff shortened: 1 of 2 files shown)


protesilaos:
Makes sense.  Thank you!

shrysr:
Thank you! This is an excellent addition, and I can confirm it works for
me as described.

In the same spirit, I thought it would also be convenient to have
existing tags for the note pre-selected? This would help with quick tag
addition or deletion. If you like, I can open this as a separate issue.

ggjp:
I considered tags too, but I assumed it would be more complex because they use `completing-read-multiple`. I have not really tried, to be honest.

protesilaos:
> I considered tags too, but I assumed it would be more complex [...]

We have `denote-last-keywords`, so maybe that will do the trick?  If there is one tag, it is a string.  If there are mutliple, they are a list of strings.

protesilaos:
No, actually ignore what I wrote.  We need to read from the underlying file.  Sorry!

PR #33, opened 2022-07-04
Add denote-prompts option and refactor the denote command

jeanphilippegg:
I suggest this refactor of the `denote` command (which uses your most recent commit (`denote--subdirs-prompt`)).

My idea is to have a new user option `denote-prompts` defaulting to `'(title keywords)`. As can be guessed, this new user option specifies the prompts of the `denote` command in the order desired by the user (as a bonus).

The commands `denote-date`, `denote-type`, and `denote-subdirectory` could be deprecated because we can just let-bind the `denote-prompts` variable to get the desired prompts. I suggest we keep them as convenience functions, though. But they should be reimplemented to just call the new `denote` command.

Here are some advantages of this approach:
- It gives more flexibility to the user:
  - A "flat-structure" user would be fine with `'(title keywords)` (the default).
  - A "subdirs" user would like `'(subdirectory title keywords)`.
  - A "no-tags" user will be fine with just `'(title)`.
  - Any other combination in any order is possible!
- It reduces the number of functions to maintain in the code.

I wanted to keep this pull request short, so I have not yet refactored the convenience functions as stated above. Also, I believe I have not changed any existing behavior of denote, except that I did not want to overload the `denote` function with optional arguments to provide defaut values. Instead, I created the function `denote-no-prompt` with optional arguments. The few examples in README.org will have to be modified accordingly.

What do you think?

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -162,6 +162,21 @@ the appropriate list of strings."
   :group 'denote
   :type 'boolean)
 
+(defcustom denote-prompts '(title keywords)
+  "Specify the prompts of the `denote' command.
+
+Valid values are:
+  date: The date of the note.
+  file-type: The filetype of the note (md, txt, org).
+  subdirectory: The subdirectory in which to create the note.
+  title: The title of the note. If unspecified, the title is
+         \"unspecified\".
+  keywords: The keywords of the note.
+
+The prompts will happen in the given order."
+  :group 'denote
+  :type '(repeat symbol))
+
 (defcustom denote-sort-keywords t
   "Whether to sort keywords in new files.
 
@@ -651,26 +666,37 @@ Optional DEFAULT-TITLE is used as the default value."
 ;;;;; The `denote' command
 
 ;;;###autoload
-(defun denote (title keywords)
-  "Create new note with the appropriate metadata and file name.
+(defun denote (&optional title keywords type date subdir)
+  "Create a new note with the appropriate metadata and file name.
 
-This command first prompts for a file TITLE and then for one or
-more KEYWORDS (separated by the `crm-separator', typically a
-comma).  The latter supports completion though any arbitrary
-string can be inserted.
+When called interactively, the metadata and file name are prompted
+according to the value of `denote-prompts'.
 
-Completion candidates are those of `denote-known-keywords'.  If
-`denote-infer-keywords' is non-nil, then keywords in existing
-file names are also provided as candidates.
-
-When `denote-sort-keywords' is non-nil, keywords are sorted
-alphabetically in both the file name and file contents."
-  (interactive
-   (list
-    (denote--title-prompt)
-    (denote--keywords-prompt)))
-  (denote--prepare-note title keywords)
-  (denote--keywords-add-to-history keywords))
+Optional TITLE should be a string or nil.
+Optional KEYWORDS should be a list of strings.
+Optional TYPE should be a symbol among 'org, 'markdown-toml,
+'markdown-yaml or 'text.
+Optional DATE should be a string representing a date, or nil.
+Optional SUBDIR should be the path string of the subdir in
+which to create the note, or nil."
+  (interactive)
+  (when (called-interactively-p 'any)
+    (dolist (prompt denote-prompts)
+      (pcase prompt
+        ('title (setq title (denote--title-prompt)))
+        ('date (setq date (denote--date-prompt)))
+        ('file-type (setq type (denote--file-type-prompt)))
+        ('subdirectory (setq subdir (denote--subdirs-prompt)))
+        ('keywords (setq keywords (denote--keywords-prompt))))))
+  (let* ((denote-file-type (denote--file-type-symbol (or type denote-file-type)))
+         (date (if (or (null date) (string-empty-p date))
+                   (current-time)
+                 (denote--valid-date date)))
+         (id (format-time-string denote--id-format date))
+         (denote-directory (or subdir (denote-directory))))
+    (denote--barf-duplicate-id id)
+    (denote--prepare-note (or title "") keywords nil date id)
+    (denote--keywords-add-to-history keywords)))
 
 (defalias 'denote-create-note (symbol-function 'denote))
 
@@ -698,14 +724,13 @@ here for clarity."
    (t (user-error "`%s' is not a symbol or string" filetype))))
 
 ;;;###autoload
-(defun denote-type (filetype)
-  "Like `denote' but with FILETYPE for `denote-file-type'.
-In practice, this command lets you create, say, a Markdown file
-even when your default is Org.
+(defun denote-type ()
+  "Convenience function to have `denote' prompt for a date.
 
-When called from Lisp the FILETYPE must be a symbol."
-  (interactive (list (denote--file-type-prompt)))
-  (let ((denote-file-type (denote--file-type-symbol filetype)))
+This is equivalent to calling `denote' when `denote-prompts' is set to
+'(file-type title keywords)."
+  (interactive)
+  (let ((denote-prompts '(file-type title keywords)))
     (call-interactively #'denote)))
 
 (defalias 'denote-create-note-using-type (symbol-function 'denote-type))
@@ -741,30 +766,14 @@ NO-CHECK-CURRENT passes the appropriate flag to
     t))
 
 ;;;###autoload
-(defun denote-date (date title keywords)
-  "Like `denote', but create new note for given DATE.
+(defun denote-date ()
+  "Convenience function to have `denote' prompt for a date.
 
-DATE can either be something like 2022-06-16 or that plus time:
-2022-06-16 14:30.
-
-The hour can be omitted, in which case it is interpreted as
-00:00.  Beware that you might create files with non-unique
-identifiers if they both have the same date and time.  In such a
-case, Denote will refrain from creating the new note.  Try with
-another DATE value where, for instance, a different time is
-specified.
-
-The TITLE and KEYWORDS arguments are the same as with `denote'."
-  (interactive
-   (list
-    (denote--date-prompt)
-    (denote--title-prompt)
-    (denote--keywords-prompt)))
-  (when-let ((d (denote--valid-date date))
-             (id (format-time-string denote--id-format d))
-             ((denote--barf-duplicate-id id)))
-    (denote--prepare-note title keywords nil d id)
-    (denote--keywords-add-to-history keywords)))
+This is equivalent to calling `denote' when `denote-prompts' is set to
+'(date title keywords)."
+  (interactive)
+  (let ((denote-prompts '(date title keywords)))
+    (call-interactively #'denote)))
 
 (defalias 'denote-create-note-using-date (symbol-function 'denote-date))
 
@@ -803,22 +812,14 @@ The TITLE and KEYWORDS arguments are the same as with `denote'."
     (denote--subdirs-completion-table dirs)))
 
 ;;;###autoload
-(defun denote-subdirectory (directory title keywords)
-  "Like `denote' but ask for DIRECTORY to put the note in.
+(defun denote-subdirectory ()
+  "Convenience function to have `denote' prompt for subdirectory.
 
-The DIRECTORY is either the variable `denote-directory' or a
-subdirectory of it.  The TITLE and KEYWORDS are the same as for
-the `denote' command.
-
-Denote does not create subdirectories."
-  (interactive
-   (list
-    (denote--subdirs-prompt)
-    (denote--title-prompt)
-    (denote--keywords-prompt)))
-  (let ((denote-directory directory))
-    (denote--prepare-note title keywords)
-    (denote--keywords-add-to-history keywords)))
+This is equivalent to calling `denote' when `denote-prompts' is set to
+'(subdirectory title keywords)."
+  (interactive)
+  (let ((denote-prompts '(subdirectory title keywords)))
+    (call-interactively #'denote)))
 
 (defalias 'denote-create-note-in-subdirectory (symbol-function 'denote-subdirectory))
 


protesilaos:
> My idea is to have a new user option `denote-prompts` defaulting to `'(title keywords)`. [...] The commands `denote-date`, `denote-type`, and `denote-subdirectory` could be deprecated

In principle, I like this direction!  Keeps things in check.

> Here are some advantages of this approach: [...] A "flat-structure" user would be fine with [...] A "subdirs" user would like [...]

What happens if the user wants a dynamic approach?  For example, I normally want a note in the root dir, but now I need to record this new entry in a subdir.  Or think about the `denote-date` workflow, which is useful when someone writes, e.g., a daily diary but missed yesterday's entry.

Is this approach possible?  Or is modifying a variable a prerequisite?

Substantively, I am interested in the implications of the `defcustom` you propose, especially as concerns the "no tags" user.  Currently this puts us in uncharted territory as the filename is hardcoded.  There is a thread on the mailing list about it and I am not against the idea, though it needs to be carefully considered: <https://lists.sr.ht/~protesilaos/denote/%3C8D392BC3-980A-4E5B-9480-D6A00BE8279F@baty.net%3E>.

> I wanted to keep this pull request short, so I have not yet refactored the convenience functions as stated above.

No worries about that.  It is good to get the discussion going.

A maintenance note here: I was thinking about releasing version `0.2.0` to wrap up the progress we have made thus far.  There have been some bug fixes and there was the major change for the `id:` links.  What do you think about putting this on hold for a few days so that we give it as much time as it needs for `0.3.0`?  For now I want to test everything we have and release `0.2.0` on Tuesday or Wednesday if no major problems exist.  What do you think?

[ Also because the GNU ELPA package was not actually built, as I explained on the mailing list: <https://lists.sr.ht/~protesilaos/denote/%3C9d600ff0-4fed-2ad7-5dbc-5a194639a045%40gmail.com%3E> ]

ggjp:
> What happens if the user wants a dynamic approach? For example, I normally want a note in the root dir, but now I need to record this new entry in a subdir. Or think about the `denote-date` workflow, which is useful when someone writes, e.g., a daily diary but missed yesterday's entry.
> 
> Is this approach possible? Or is modifying a variable a prerequisite?

Yes, without convenience functions, this would look like this:

    (global-set-key (kbd "C-c n n") #'denote)
    (global-set-key (kbd "C-c n N") (lambda () (let ((denote-prompts '(subdirectory title keywords))) (denote)))
    (global-set-key (kbd "C-c n s") (lambda () (let ((denote-prompts '(subdirectory title keywords))) (denote)))
    (global-set-key (kbd "C-c n d") (lambda () (let ((denote-prompts '(date title keywords))) (denote)))

But I think we should continue to provide the convenience functions to wrap this simple code.

> Substantively, I am interested in the implications of the `defcustom` you propose, especially as concerns the "no tags" user. Currently this puts us in uncharted territory as the filename is hardcoded.

It seems to work fine without keywords. The file name ends with double _, but it is not so bad. The metadata header for `md-yaml` files also contains double _ which is a minor bug, I think.

> What do you think about putting this on hold for a few days so that we give it as much time as it needs for `0.3.0`? For now I want to test everything we have and release `0.2.0` on Tuesday or Wednesday if no major problems exist. What do you think?

Good with me! I will go ahead and refactor the convenience functions in my branch in the meantime to give you a better idea of what I meant. I think that even if we don't end up making `denote-prompts` available to users as a defcustom, it could be a simpler way to handle the various functions, internally.

protesilaos:
> Yes, without convenience functions, this would look like this: [...]
> But I think we should continue to provide the convenience functions to
> wrap this simple code.

Agreed!

> It seems to work fine without keywords. The file name ends with double
> _, but it is not so bad. The metadata header for md-yaml files also
> contains double _ which is a minor bug, I think.

I think we should iron those out, such as by tweaking
`denote--format-file` to not insert the `__` when no keywords are
supplied.  In general, things should work as we only really need the
identifier in the file name for the links.  Though further tweaks might
be necessary.

> Good with me! I will go ahead and refactor the convenience functions
> in my branch in the meantime to give you a better idea of what I
> meant. I think that even if we don't end up making `denote-prompts`
> available to users as a defcustom, it could be a simpler way to handle
> the various functions, internally.

Good!  Internal refinements are important.  I will now focus on
preparing version `0.2.0`.

jeanphilippegg:
I have refactored the convenience functions.

Notes:
- The new convenience functions do not accept any arguments anymore. Though we never used them in denote's code, this will break the code of a user that was using them with arguments in their config file. The fix is to use the unified `denote-no-prompt` function that will take the supplied arguments from the user.
  - Additional comment here: I think it is a good thing to remove arguments from those functions. I find the current state sometimes confusing. For example, the `denote-date` accepts `date title keywords` while `denote-type` only accepts `filetype`.
- As of now, I have not changed the current behavior of denote. I have just consolidated the convenience functions into `denote-no-prompt` respecting the way they were implemented before. So the keywords argument, for example, must be a string, maybe empty and not nil. I have extended the docstring with the expected values of the arguments.

Todo:
- The current info in the docstrings of the conveniences functions should not stay there. I suggest moving it in the `denote-prompts` variable or directly in the `denote` command.
- In `denote-prompts`, do you prefer `filetype` or `file-type`?
- Update the README.

protesilaos:
> - The new convenience functions do not accept any arguments
>   anymore. Though we never used them in denote's code, this will break
>   the code of a user that was using them with arguments in their
>   config file. The fix is to use the unified `denote-no-prompt`
>   function that will take the supplied arguments from the user.

Okay.  I will test the code later today.  Looking at the `denote`
command, it seems we can retain the arguments inside the `interactive`
spec.  The other commands can be refactored without them.  That way we
don't break people's configs while also streamlining the code.  What do
you think?

> - Additional comment here: I think it is a good thing to remove
>   arguments from those functions.  I find the current state sometimes
>   confusing. For example, the `denote-date` accepts `date title
>   keywords` while `denote-type` only accepts `filetype`.

Agreed.

> - As of now, I have not changed the current behavior of denote. I have
>   just consolidated the convenience functions into `denote-no-prompt`
>   respecting the way they were implemented before. So the keywords
>   argument, for example, must be a string, maybe empty and not nil. I
>   have extended the docstring with the expected values of the
>   arguments.

We can interpret a nil value for the keywords as an empty string in the
relevant private functions, though that is an implementation detail.
Keeping it in the command is okay for now as it is easier for testing.

> - The current info in the docstrings of the conveniences functions
>   should not stay there. I suggest moving it in the `denote-prompts`
>   variable or directly in the `denote` command.

Fine.

> - In `denote-prompts`, do you prefer `filetype` or `file-type`?

If it is user-facing `file-type`, otherwise `filetype` is okay.

> - Update the README.

This can be done last.  No need to worry about it now.

jeanphilippegg:
> Okay. I will test the code later today. Looking at the `denote` command, it seems we can retain the arguments inside the `interactive` spec. The other commands can be refactored without them. That way we don't break people's configs while also streamlining the code. What do you think?

I agree. I added this to my comment while you were replying:

> EDIT: There is something I don't like about having separate `denote` and `denote-no-prompt`. I will try to refine this later.

I will work on it!

protesilaos:
Okay!  A maintenance note here: earlier I pushed a fix for a bug and tagged version `0.2.1` to trigger a rebuild of the package. Before merging new functionality, I say we way a bit longer (2-3 days) in case any obvious bug is reported.  That way I can bump the version without having to produce a changelog.

jeanphilippegg:
I have got something cleaner, but I would like to put this on hold. I would like to make some other modifications before. You can check my other pull request.

protesilaos:
> I have got something cleaner, but I would like to put this on hold.

Okay.  I merged your other PR.  Thanks!

jeanphilippegg:
You can check the (short) implementation of the new `denote` command in its current form. I think I got something simple. Here is how it works:
- It accepts optional arguments (title, keywords, type, date, subdir). When used in a script, it creates a note according to these arguments.
- When used interactively, the necessary information is prompted according to the `denote-prompts` variable. (For example, a value of `'(title keywords date)` for `denote-prompts` would prompt for this info in that order.)

In order to get this, I used the function `called-interactively-p`. Please check the warning in its documentation and assess if this applies in this case. I am not sure. Thanks!

protesilaos:
Thank you!  I will fetch it and test it in ~20 minutes.  Since some of the commits are not relevant, like [Remove denote-no-prompt](https://github.com/protesilaos/denote/pull/33/commits/07e7c1870e83a288e2fa650b8cdbc93c455390e5), do you think you can squash everything into a single commit?  If it is too much trouble, just leave it as-is.

jeanphilippegg:
> Thank you! I will fetch it and test it in ~20 minutes. Since some of the commits are not relevant, like [Remove denote-no-prompt](https://github.com/protesilaos/denote/pull/33/commits/07e7c1870e83a288e2fa650b8cdbc93c455390e5), do you think you can squash everything into a single commit? If it is too much trouble, just leave it as-is.

It is easy, but wouldn't it be preferable to keep all commits and their commit messages in the history to better understand the evolution of the changes if we look at it in the future? What is not relevant in commit 07e7c187?

protesilaos:
I mean from the perspective of the `main` branch, as `denote-no-prompt` did not exist there.  So removing it is an internal operation within the confines of your change.

protesilaos:
Just to note a conflict caused by commit `16867ec` on the `main` branch:

```diff
<<<<<<< HEAD
  (mapcar (if denote-allow-multi-word-keywords
              #'denote--sluggify
            #'denote--sluggify-and-join)
          keywords))
||||||| constructed merge base
  (cond
   ((listp keywords)
    (if denote-allow-multi-word-keywords
        (mapcar #'denote--sluggify keywords)
      (mapcar #'denote--sluggify-and-join keywords)))
   (t
    (if denote-allow-multi-word-keywords
        (denote--sluggify keywords)
      (denote--sluggify-and-join keywords)))))
=======
  (if denote-allow-multi-word-keywords
      (mapcar #'denote--sluggify keywords)
    (mapcar #'denote--sluggify-and-join keywords)))
>>>>>>> Refactor to make the keywords argument always a list of strings
```

jeanphilippegg:
I have squashed my 5 commits!

> Just to note a conflict caused by commit `16867ec` on the `main` branch:

You are looking at some old commits which do not exist anymore (even before the squash). I remember this conflict. I fixed everything and rebased my branch on top of main. Maybe my workflow is not appropriate: Before a pull request is merged onto main, I rebase my branch in order to keep a clean history. Maybe I should not do that once a pull request is submitted...

jeanphilippegg:
I just updated the convenience functions to use `(call-interactively #'denote)` instead of `(denote)`. Sorry about that. I had not do tested them after my latest modifications. They are fixed now if you pull again.

protesilaos:
> You are looking at some old commits which do not exist anymore

Mea culpa!  I tried to apply another patch...  It now applies cleanly.  Will check and report back to you.

jeanphilippegg:
> Mea culpa!

No worry!

protesilaos:
I incorporated your change here.  Though I am making some further edits and adding them to the `refactor-denote-prompts` branch: https://github.com/protesilaos/denote/tree/refactor-denote-prompts.  More to follow.

protesilaos:
I will squash everything later.  This is for you to check what I did.

jeanphilippegg:
This is very nice!

Some questions:
- Are you okay with the use of `called-interactively-p`, given the warning in its doc?
- Why don't we need `denote--prompts` because of the new `denote--format-file`?

About supporting less fields in the file name, I think this is a good direction! I had checked that before, so I will share what I had found.

- It looks as though the file name syntax is very precise and enforced with one big regexp. This rexgexp is used in some places. For example, when validating that a file is a note in `denote-only-note-p`. Also, the fontification is relying on this regexp and its groups. I have not tested, but I expect some things to break.
- I think some more refactor need to happen to support "no keywords" and "no title". We could stop relying on one big regexp. For example, we could use smaller regexps:
  - A valid note starts with an ID (and possibly ends with a supported extension).
  - To find the title in the file name, we can only check for a match starting with "--" and ending when an invalid character for a title is found.
  - To find the keywords in the file name, we can only check for a match starting with double _ and ending when an invalid character for keywords is found.

I really like this direction. To keep things simple, I would have suggested to keep it as a separate issue/branch, but you see a connection between the two as stated in commit `5a4167d`. Can you clarify this?

Thanks!

jeanphilippegg:
> What do you think?

I agree with everything you said. Yesterday (for me), I was tired and somehow forgot that it was the reason I wanted to hold this branch. So I sent my message. But I am glad we had this discussion.

I suggest the following course of action:
1. Create a new branch on `main` and get your modifications of `denote--format-file` there.
2. Rework the way we use regexps on that branch and merge it in `main` when ready. Then let a few days pass to check if any issues arise.
3. Come back to this current branch after all this.

protesilaos:
> Create a new branch on `main` and get your modifications of
> `denote--format-file there`.

Sure!  Just to be clear because now I am the one who is tired :) I will
keep the `refactor-denote-prompts` branch at commit `95feac8` and move
commits `2c17400`, `5a4167d` to a new branch that starts from `main`,
right?

protesilaos:
Okay, I think I did everything correctly (first time I had a real-world scenario of cherry-picking a commit).  The new branches:

- `file-name-permutations` contains the change to `denote--format-file`.
- `refactor-denote-prompts` is what you were working on plus the tweaks I made on top.

As always, you are welcome to make any changes. 

Will get back to this tomorrow.  Local time is ~22:30, but I've been awake since ~05:30.

jeanphilippegg:
>     1. Create a new branch on `main` and get your modifications of `denote--format-file` there.
>     2. Rework the way we use regexps on that branch and merge it in `main` when ready. Then let a few days pass to check if any issues arise.
>     3. Come back to this current branch after all this.

Maybe we could move on with step 3? We can merge `main` into `refactor-denote-prompts` without conflict.

protesilaos:
> Maybe we could move on with step 3?

Done!

PR #35, opened 2022-07-04
Remove "/" from denote-directory path in denote--subdirs-prompt

ggjp:
We need to remove the trailing "/" from the denote-directory path in `denote--subdirs-prompt` or else it cannot be selected.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -813,7 +813,7 @@ The TITLE and KEYWORDS arguments are the same as with `denote'."
 
 (defun denote--subdirs-prompt ()
   "Handle user input on choice of subdirectory."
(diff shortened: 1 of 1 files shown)


protesilaos:
Thanks!  I am curious why that happens.  Maybe this is specific to the
completion framework?  I am using vertico right now and the trailing
slash does not prevent the selection of the root dir.  Trying this in
the scratch buffer:

```elisp
(denote--subdirs-prompt)
;; => "/home/prot/Documents/notes/"
```

protesilaos:
Gotcha!  Merged it.

PR #36, opened 2022-07-06
Clean keywords usage in order to support no keywords

jeanphilippegg:
This is meant as a first step in supporting Denote with no keywords.

In various places, the `keywords` argument can be a single string or a list of strings. There are `stringp` checks and other conditions in various places of the code to handle this.

Now that we also want to support no keywords, I think it would be cleaner if `keywords` would always be a list of strings (at least internally in the code), whether it contains 0, 1 or multiple keywords.

This pull request just removes the previous conditions and checks. Proper support for the use case of no keywords would come later.

I have done some quick tests and it does not seem to have broken anything.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -316,15 +316,9 @@ trailing hyphen."
 
 (defun denote--sluggify-keywords (keywords)
   "Sluggify KEYWORDS."
-  (cond
-   ((listp keywords)
-    (if denote-allow-multi-word-keywords
-        (mapcar #'denote--sluggify keywords)
-      (mapcar #'denote--sluggify-and-join keywords)))
-   (t
-    (if denote-allow-multi-word-keywords
-        (denote--sluggify keywords)
-      (denote--sluggify-and-join keywords)))))
+  (if denote-allow-multi-word-keywords
+      (mapcar #'denote--sluggify keywords)
+    (mapcar #'denote--sluggify-and-join keywords)))
 
 (defun denote--file-empty-p (file)
   "Return non-nil if FILE is empty."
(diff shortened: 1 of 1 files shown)


protesilaos:
I just tested it.  Everything seems to work fine.  Merging it.

protesilaos:
By the way, you may be interested in having more informative diff
headings (the text on the line with `@@`):
https://protesilaos.com/codelog/2021-01-26-git-diff-hunk-elisp-org/

For example, with my settings, I get this first line:

```diff
@@ -445,34 +439,19 @@ (defun denote--keywords-prompt ()
 output is sorted with `string-lessp'."
   (let ((choice (denote--keywords-crm (denote-keywords))))
     (setq denote-last-keywords
-          (cond
-           ((null choice)
-            "")
-           ((= (length choice) 1)
-            (car choice))
-           ((if denote-sort-keywords
-                (sort choice #'string-lessp)
-              choice))))))
+          (if denote-sort-keywords
+              (sort choice #'string-lessp)
+            choice))))
```

Whereas your patch's was:

```diff
@@ -445,34 +439,19 @@ In the case of multiple entries, those are separated by the
 output is sorted with `string-lessp'."
   (let ((choice (denote--keywords-crm (denote-keywords))))
     (setq denote-last-keywords
-          (cond
-           ((null choice)
-            "")
-           ((= (length choice) 1)
-            (car choice))
-           ((if denote-sort-keywords
-                (sort choice #'string-lessp)
-              choice))))))
+          (if denote-sort-keywords
+              (sort choice #'string-lessp)
+            choice))))
```

You don't need any of this, of course.  Just in case you are interested.

jeanphilippegg:
> You don't need any of this, of course. Just in case you are interested.

I am interested! But I am not sure I understand. From what I understand, the settings you suggest will change the way I see diffs locally. You mention a patch I sent with less informative heading. I am not aware of a patch I sent. Is this something you get when trying to merge on sourcehut maybe?

protesilaos:
Ah sorry, I meant your PR.  Internally it is a patch (append `.patch` to a PR's URL to see it).  Yes, the change is for your local setup which, however, gets reflected in the patch/PR you produce.

jeanphilippegg:
But I create the pull request through github web interface. I was not aware of any other way.

jeanphilippegg:
I changed my configs. We'll see if it has any effect in my future pull requests. Thanks!

protesilaos:
Okay, let's see if it works.  Again, it does not matter for me.  Just thought I would mention it.

PR #38, opened 2022-07-08
More refactors to remove the big regexp

jeanphilippegg:
This pull request contains:

- Removal of `denote--keywords-in-files`. It was not doing much and it was only used by `denote--inferred-keywords` (which I have refactored a little).
- Refactor of `denote--extract` into `denote--extract-keywords-from-path`. Since we want to move away from a single big regexp, the function `denote--extract` would not be useful anymore.
  - `denote--extract-keywords-from-path` is implemented using `string-match` while `denote--extract` was creating a temp buffer for the string and using `re-search-forward`. Any reason why it was implemented that way?
- Refactor of `denote--only-note-p` to remove the use of denote--file-only-note-regexp. Instead we can just check if the file name starts with an ID and ends with a Denote extension. Everything inbetween is unchecked.

I did not understand the font locking variables in `dired-faces.el`, so I was not able to change it. That is the only thing remaining before we can delete `denote--file-title-regexp` and `denote--file-regexp`.

### DIFF
--- a/denote-faces.el
+++ b/denote-faces.el
@@ -83,14 +83,17 @@ and seconds."
   :group 'denote-faces)
 
 (defconst denote-faces-file-name-keywords
-  `((,denote--file-regexp
+  `((,denote--id-regexp
      (1 'denote-faces-date)
-     (2 'denote-faces-time)
-     (3 'denote-faces-delimiter)
-     (4 'denote-faces-title)
-     (5 'denote-faces-delimiter)
-     (6 'denote-faces-keywords)
-     (7 'denote-faces-extension))
+     (2 'denote-faces-time))
+    (,denote--title-regexp
+     (1 'denote-faces-title))
+    (,denote--keywords-regexp
+     (1 'denote-faces-keywords))
+    (,denote--extension-regexp
+     (0 'denote-faces-extension))
+    ("--"
+     (0 'denote-faces-delimiter t))
     ("_"
      (0 'denote-faces-delimiter t)))
   "Keywords for fontification of file names.")
--- a/denote.el
+++ b/denote.el
@@ -239,6 +239,9 @@ are described in the doc string of `format-time-string'."
 (defconst denote--id-regexp "\\([0-9]\\{8\\}\\)\\(T[0-9]\\{6\\}\\)"
   "Regular expression to match `denote--id-format'.")
 
+(defconst denote--keywords-regexp "__\\([0-9A-Za-z_-]*\\)"
+  "Regular expression to match keywords.")
+
 (defconst denote--file-title-regexp
   (concat denote--id-regexp "\\(--\\)\\(.*\\)\\(__\\)")
   "Regular expression to match file names from `denote'.")
@@ -278,16 +281,6 @@ We consider those characters illigal for our purposes.")
     (unless (file-directory-p path)
       (make-directory path t))
     (file-name-as-directory path)))
-
-(defun denote--extract (regexp str &optional group)
-  "Extract REGEXP from STR, with optional regexp GROUP."
-  (when group
-    (unless (and (integerp group) (> group 0))
-      (error "`%s' is not a positive integer" group)))
-  (with-temp-buffer
-    (insert str)
-    (when (re-search-forward regexp nil t -1)
-      (match-string (or group 1)))))
 
 (defun denote--slug-no-punct (str)
   "Convert STR to a file name slug."
@@ -395,24 +388,25 @@ part of the list."
         f))
     (denote--directory-files))))
 
-(defun denote--keywords-in-files ()
-  "Produce list of keywords in `denote--directory-files'."
-  (delq nil (mapcar
-             (lambda (x)
-               (denote--extract denote--file-regexp x 6))
-             ;; REVIEW 2022-07-03: I tested this with ~3000 files.  It
-             ;; has about 2 seconds of delay on my end.  After I placed
-             ;; the list of those files in a variable instead of calling
-             ;; `denote--directory-files', there was no noticeable
-             ;; performance penalty.
-             (denote--directory-files))))
+(defun denote--extract-keywords-from-path (path)
+  "Extract keywords from PATH."
+  (let* ((file-name (file-name-nondirectory path))
+         (kws (when (string-match denote--keywords-regexp file-name)
+                (match-string-no-properties 1 file-name))))
+    (when kws
+      (split-string kws "_"))))
 
 (defun denote--inferred-keywords ()
   "Extract keywords from `denote--directory-files'."
-  (let ((sequence (denote--keywords-in-files)))
-    (mapcan (lambda (s)
-              (split-string s "_" t))
-            sequence)))
+  (delete-dups
+   (mapcan (lambda (p)
+             (denote--extract-keywords-from-path p))
+           ;; REVIEW 2022-07-03: I tested this with ~3000 files.  It
+           ;; has about 2 seconds of delay on my end.  After I placed
+           ;; the list of those files in a variable instead of calling
+           ;; `denote--directory-files', there was no noticeable
+           ;; performance penalty.
+           (denote--directory-files))))
 
 (defun denote-keywords ()
   "Return appropriate list of keyword candidates.
--- a/denote.el
+++ b/denote.el
@@ -239,19 +239,14 @@ are described in the doc string of `format-time-string'."
 (defconst denote--id-regexp "\\([0-9]\\{8\\}\\)\\(T[0-9]\\{6\\}\\)"
   "Regular expression to match `denote--id-format'.")
 
+(defconst denote--title-regexp "--\\([0-9A-Za-z-]*\\)"
+  "Regular expression to match keywords.")
+
 (defconst denote--keywords-regexp "__\\([0-9A-Za-z_-]*\\)"
   "Regular expression to match keywords.")
 
 (defconst denote--extension-regexp "\\.\\(org\\|md\\|txt\\)"
   "Regular expression to match supported Denote extensions.")
-
-(defconst denote--file-title-regexp
-  (concat denote--id-regexp "\\(--\\)\\(.*\\)\\(__\\)")
-  "Regular expression to match file names from `denote'.")
-
-(defconst denote--file-regexp
-  (concat denote--file-title-regexp "\\([0-9A-Za-z_-]*\\)\\(\\.?.*\\)")
-  "Regular expression to match the entire file name'.")
 
 (defconst denote--punctuation-regexp "[][{}!@#$%^&*()_=+'\"?,.\|;:~`‘’“”/]*"
   "Regular expression of punctionation that should be removed.
--- a/denote.el
+++ b/denote.el
@@ -242,6 +242,9 @@ are described in the doc string of `format-time-string'."
 (defconst denote--keywords-regexp "__\\([0-9A-Za-z_-]*\\)"
   "Regular expression to match keywords.")
 
+(defconst denote--extension-regexp "\\.\\(org\\|md\\|txt\\)"
+  "Regular expression to match supported Denote extensions.")
+
 (defconst denote--file-title-regexp
   (concat denote--id-regexp "\\(--\\)\\(.*\\)\\(__\\)")
   "Regular expression to match file names from `denote'.")
@@ -249,10 +252,6 @@ are described in the doc string of `format-time-string'."
 (defconst denote--file-regexp
   (concat denote--file-title-regexp "\\([0-9A-Za-z_-]*\\)\\(\\.?.*\\)")
   "Regular expression to match the entire file name'.")
-
-(defconst denote--file-only-note-regexp
-  (concat denote--file-regexp "\\.\\(org\\|md\\|txt\\)")
-  "Regular expression to match the entire file name of a note file.")
 
 (defconst denote--punctuation-regexp "[][{}!@#$%^&*()_=+'\"?,.\|;:~`‘’“”/]*"
   "Regular expression of punctionation that should be removed.
@@ -321,10 +320,13 @@ trailing hyphen."
 (defun denote--only-note-p (file)
   "Make sure FILE is an actual Denote note.
 FILE is relative to the variable `denote-directory'."
-  (and (not (file-directory-p file))
-       (file-regular-p file)
-       (string-match-p denote--file-only-note-regexp file)
-       (not (string-match-p "[#~]\\'" file))))
+  (let ((file-name (file-name-nondirectory file)))
+    (and (not (file-directory-p file))
+         (file-regular-p file)
+         (string-match-p (concat "\\`" denote--id-regexp
+                                 ".*" denote--extension-regexp "\\'")
+                         file-name)
+         (not (string-match-p "[#~]\\'" file)))))
 
 (defun denote--file-name-relative-to-denote-directory (file)
   "Return file name of FILE relative to the variable `denote-directory'.


jeanphilippegg:
I have been able to fix the fontification in `denote-dired.el` by guessing how it works. The last items have a `t` at the end. I don't know if it is okay.

protesilaos:
Good day!

I notice your branch is `jeanphilippegg:file-name-permutations` while the merge target is the `main` branch.  Is this correct or did you mean to add to the `file-name-permutations` branch instead?

* * *

> - Refactor of `denote--extract` into `denote--extract-keywords-from-path`. Since we want to move away from a single big regexp, the function `denote--extract` would not be useful anymore.

Fine.  The rationale for that traces back to the pre-Denote days of this code.

> - `denote--extract-keywords-from-path` is implemented using `string-match` while `denote--extract` was creating a temp buffer for the string and using `re-search-forward`. Any reason why it was implemented that way?

I don't recall why it was implemented that way.  As it is old code, the answer usually is "because I didn't know any better".  Or maybe I was doing something else with it, but it got removed in the process.

> - Refactor of `denote--only-note-p` to remove the use of denote--file-only-note-regexp. Instead we can just check if the file name starts with an ID and ends with a Denote extension. Everything inbetween is unchecked.

Good!

jeanphilippegg:
> I notice your branch is `jeanphilippegg:file-name-permutations` while the merge target is the `main` branch. Is this correct or did you mean to add to the `file-name-permutations` branch instead?

I meant the latter. Fixed!

Check my other comment if you have not noticed! I was able to fix fontification as well.

Have a good day!

protesilaos:
> I meant the latter. Fixed!

Thanks!  Can I merge this or are there any final tweaks you need to make?

> Check my other comment if you have not noticed! I was able to fix fontification as well.

I noticed.  Thank you!

[ Not related to this branch: what do you think about how the identifier is fontified?  Do you prefer to add one colour to the whole date+time or do you see any value in using separate colours? ]

jeanphilippegg:
> Thanks! Can I merge this or are there any final tweaks you need to make?

This can be merged. Even in main after we test it thoroughly. I don't think there is anything left to do.

> I noticed. Thank you!

Ok. I had a comment about a `t` at the end of some items, which I don't know what it is for.

> [ Not related to this branch: what do you think about how the identifier is fontified? Do you prefer to add one colour to the whole date+time or do you see any value in using separate colours? ]

I personally like the way I see them now: as one single color (blue-green using dark modus-themes).

protesilaos:
> This can be merged. Even in main after we test it thoroughly. I don't think there is anything left to do.

Okay, doing it now.  I will just double-check the branch and then merge to `main`.

> Ok. I had a comment about a `t` at the end of some items, which I don't know what it is for.

Check the doc string of `font-lock-keywords`.  The `t` is for the OVERRIDE argument.  Maybe we do not need it with those new regular expressions.  Before it was necessary, otherwise it was ignored.

> I personally like the way I see them now

I also thought this was the better default.

jeanphilippegg:
> Check the doc string of `font-lock-keywords`. The `t` is for the OVERRIDE argument. Maybe we do not need it with those new regular expressions. Before it was necessary, otherwise it was ignored.

I looked at it, but couldn't understand it. Now that I read it again, I think it does not matter if they stay there.

> I also thought this was the better default.

Yes, my reasoning is that a paler color for the time would make the date stand out, but I don't care enough about the date of the note that I actually prefer a single color to avoid too many colors.

protesilaos:
> I looked at it, but couldn't understand it. Now that I read it again, I think it does not matter if they stay there.

I removed them for different reasons in commits `29a5a2d` and `5abcd2f`.

> I actually prefer a single color to avoid too many colors.

Same here.

protesilaos:
The fontifications regexps will need to be reworked, as they otherwise apply to non-Denote files.  This is relevant is `denote-dired-mode` is set globally.  I am working on it.

protesilaos:
This removes false positives, but we lose the delimiter fontification:

```elisp
(defvar denote-faces--file-name
  (format "\\(?1:%s\\)\\(?2:%s\\)\\(?3:%s\\)\\(?4:%s\\)"
          denote--id-regexp
          denote--title-regexp
          denote--keywords-regexp
          "\\..*"))

(defconst denote-faces-file-name-keywords
  `((,denote-faces--file-name
     (1 'denote-faces-date)
     (2 'denote-faces-title)
     (3 'denote-faces-keywords)
     (4 'denote-faces-extension)))
  "Keywords for fontification of file names.")
```

I will think about it a bit more before committing.

jeanphilippegg:
> This removes false positives, but we lose the delimiter fontification:
> 
> ```emacs-lisp
> (defvar denote-faces--file-name-regexp
>   (format "\\(?1:%s\\)\\(?2:%s\\)\\(?3:%s\\)\\(?4:%s\\)"
>           denote--id-regexp
>           denote--title-regexp
>           denote--keywords-regexp
>           "\\..*"))
> 
> (defconst denote-faces-file-name-keywords
>   `((,denote-faces--file-name-regexp
>      (1 'denote-faces-date)
>      (2 'denote-faces-title)
>      (3 'denote-faces-keywords)
>      (4 'denote-faces-extension)))
>   "Keywords for fontification of file names.")
> ```
> 
> I will think about it a bit more before committing.
> 
> EDIT: appended `-regexp` to the variable.

Is this not enforcing keywords and titles to be present again?

protesilaos:
> Is this not forcing keywords and titles to be present again?

It does.  Maybe this will do but I am still testing:

```elisp
(defvar denote-faces--file-name-regexp
  (format "\\(?1:%s\\)?\\(?2:%s\\)?\\(?3:%s\\)?\\(?4:%s\\)?"
          denote--id-regexp
          denote--title-regexp
          denote--keywords-regexp
          "\\..*"))
```

jeanphilippegg:
If we require a match for the whole file with all its fields for the fontification, I suppose we don't have much option.
These regexps concatenated might work:
id: `\\(?0:[0-9]\\{8\\}\\)\\(?1:T[0-9]\\{6\\}\\)`
title: `\\(\\(?2:--\\)\\(?3:[0-9A-Za-z-]*\\)\\)?`     (there is a `?` to make the whole group optional)
keywords: `\\(\\(?4:__\\)\\(?5:[0-9A-Za-z_-]*\\)\\)?`    (same here)
extension: `\\(?6:\\..*\\)`

protesilaos:
We cross-commented.  I added an `?` as well.  But yes, we may need to expand them the way you suggest.

jeanphilippegg:
This seems to work:

```
(defconst denote-faces-file-name-keywords
  `((,(concat "\\(?1:[0-9]\\{8\\}\\)\\(?2:T[0-9]\\{6\\}\\)"
              "\\(?:\\(?3:--\\)\\(?4:[0-9A-Za-z-]*\\)\\)?"
              "\\(?:\\(?5:__\\)\\(?6:[0-9A-Za-z_-]*\\)\\)?"
              "\\(?7:\\..*\\)")
     (1 'denote-faces-date)
     (2 'denote-faces-time)
     (3 'denote-faces-delimiter)
     (4 'denote-faces-title)
     (5 'denote-faces-delimiter)
     (6 'denote-faces-keywords)
     (7 'denote-faces-extension))
    ("_"
     (0 'denote-faces-delimiter)))
  "Keywords for fontification of file names.")
```

protesilaos:
Yes, this is the idea.  Though this will have to be removed:

```
("_"
     (0 'denote-faces-delimiter))
```

It gives false positives.

jeanphilippegg:
Ok, but to be sure, since this is not different from what we had before, it means we had false positives before as well?

protesilaos:
> it means we had false positives before as well?

Yes.  I had not realised it before.

protesilaos:
Your concat prefixes everything with our ID, which should give us more reliable results.

jeanphilippegg:
I don't know what I have an issue with my regexps. `re-builder` is fine and captures the files names, but a file named "20220703T030303__test.org" does not get fontified.

jeanphilippegg:
And all following files after this one are not fontified....

protesilaos:
I am having problems as well.  I think this is the cause:

```
(7 'denote-faces-extension)
```

With its correspondig regexp group.

jeanphilippegg:
I have this in `*Messages*`: `Error during redisplay: (jit-lock-function 1) signaled (error "No match 3 in highlight (3 'denote-faces-delimiter)")`

protesilaos:
Yes, I got it as well.  Needs more work.

protesilaos:
Maybe the nested subgroups are the problem?  This seems to work, though we lose some of the control we wanted to have:

```elisp
(defvar denote-faces--file-name-regexp
  (format "\\(?1:%s\\)\\(?2:%s\\)?\\(?3:%s\\)?"
          denote--id-regexp
          denote--title-regexp
          denote--keywords-regexp))

(defconst denote-faces-file-name-keywords
  `((,denote-faces--file-name-regexp
     (1 'denote-faces-date)
     (2 'denote-faces-title)
     (3 'denote-faces-keywords)))
  "Keywords for fontification of file names.")
```

jeanphilippegg:
I think the reason it does not work is because, (if I take my code), the step of applying `denote-faces-delimiter` to group 3 is done after the match. But the match does not contain a group 3.

I would expect your code to not work either if you have no title.

protesilaos:
It works with those:

```
20220502T113032--this-is-a-test__keytest
20220502T113036__keytest
20220502T113037__keytest.txt
20220502T113038--testing__keytest.txt
20220502T113048--testing.txt
20220502T113049.txt
20220502T113050
```

jeanphilippegg:
You're right. I don't understand the error then...

protesilaos:
> The nesting is the issue and your code works.

I am retrying your concat but without the nested groups:

```elisp
(defvar denote-faces--file-name-regexp
  (concat "\\([0-9]\\{8\\}\\)\\(T[0-9]\\{6\\}\\)"
          "\\(--\\)?\\([0-9A-Za-z-]*\\)?"
          "\\(__\\)?\\([0-9A-Za-z_-]*\\)?"
          "\\(\\..*\\)?"))

(defconst denote-faces-file-name-keywords
  `((,denote-faces--file-name-regexp
     (1 'denote-faces-date)
     (2 'denote-faces-time)
     (3 'denote-faces-delimiter)
     (4 'denote-faces-title)
     (5 'denote-faces-delimiter)
     (6 'denote-faces-keywords)
     (7 'denote-faces-extension)))
  "Keywords for fontification of file names.")
```

It still gives me the error.  I do not understand why...

jeanphilippegg:
This is almost what we had before all the modifications... Strange.

protesilaos:
> This is almost what we had before all the modifications... Strange.

Indeed.  Though it may be that it was always faulty but we only realised it now.  If this cannot work, we can try the simpler one:

```elisp
(defvar denote-faces--file-name-regexp
  (format "\\(?1:%s\\)\\(?2:%s\\)?\\(?3:%s\\)?"
          denote--id-regexp
          denote--title-regexp
          denote--keywords-regexp))

(defconst denote-faces-file-name-keywords
  `((,denote-faces--file-name-regexp
     (1 'denote-faces-date)
     (2 'denote-faces-title)
     (3 'denote-faces-keywords)))
  "Keywords for fontification of file names.")
```

protesilaos:
From `font-lock-keywords`:

> If LAXMATCH is non-nil, that means don’t signal an error if there is
> no match for SUBEXP in MATCHER.

protesilaos:
It works!

```elisp

(defvar denote-faces--file-name-regexp
  (concat "\\([0-9]\\{8\\}\\)\\(T[0-9]\\{6\\}\\)"
          "\\(--\\)?\\([0-9A-Za-z-]*\\)?"
          "\\(__\\)?\\([0-9A-Za-z_-]*\\)?"
          "\\(\\..*\\)?"))

(defconst denote-faces-file-name-keywords
  `((,denote-faces--file-name-regexp
     (1 'denote-faces-date)
     (2 'denote-faces-time)
     (3 'denote-faces-delimiter nil t)
     (4 'denote-faces-title nil t)
     (5 'denote-faces-delimiter nil t)
     (6 'denote-faces-keywords nil t)
     (7 'denote-faces-extension nil t )))
  "Keywords for fontification of file names.")
```

jeanphilippegg:
Fantastic!

jeanphilippegg:
We don't really need to bring back the nesting of parentheses, but if we do, it does not fontify everything... I'm fine with your solution!

protesilaos:
I pushed the change and added you as co-author of the commit.  Thank you!

protesilaos:
I will merge the branch later today.  Just to be sure everything is in order.

protesilaos:
Well, actually it is fontified as the title.  Hmm...

jeanphilippegg:
Yes, I understand why now. I am trying to bring back the nesting, but it still is fontified.

protesilaos:
I think we may need the MATCH-ANCHORED approach documented in `font-lock-keywords`.  I am reading about it now.

jeanphilippegg:
I tried to bring the nesting back with:
```elisp
(defvar denote-faces--file-name-regexp
  (concat "\\(?1:[0-9]\\{8\\}\\)\\(?2:T[0-9]\\{6\\}\\)"
          "\\(?8:\\(?3:--\\)\\(?4:[0-9A-Za-z-]*\\)\\)?"
          "\\(?9:\\(?5:__\\)\\(?6:[0-9A-Za-z_-]*\\)\\)?"
          "\\(?7:\\..*\\)?")
  "Regexp of file names for fontification.")
```
`re-builder` would suggest that it would work, but it does not... It still is fontified. I don't understand why.

Anyway, maybe it is not so bad if we fontify anything that starts with an id? This is what we consider internally as a note.

jeanphilippegg:
All these get fontified using your code:
```
"20220303T03030303030303--.org"
"20220303T03030303030303--+.org"
"20220303T030303030303+03--.org"
```

protesilaos:
What about this?

```elisp
(defvar denote-faces--file-name-regexp
  (concat "\\(?1:[0-9]\\{8\\}\\)\\(?2:T[0-9]\\{6\\}\\)"
          "\\(?:\\(?3:--\\)\\(?4:[0-9A-Za-z-]*\\)\\)?"
          "\\(?:\\(?5:__\\)\\(?6:[0-9A-Za-z_-]*\\)\\)?"
          "\\(?7:\\..*\\)?")
  "Regexp of file names for fontification.")

(defconst denote-faces-file-name-keywords
  `((,denote-faces--file-name-regexp
     (1 'denote-faces-date)
     (2 'denote-faces-time)
     (3 'denote-faces-delimiter nil t)
     (4 'denote-faces-title nil t)
     (5 'denote-faces-delimiter nil t)
     (6 'denote-faces-keywords nil t)
     (7 'denote-faces-extension nil t )))
  "Keywords for fontification of file names.")
```

I made the title bold to see the effect:

Screenshot from 2022-07-08 11-35-12

protesilaos:
Same screenshot, but red title component to make it easier:

Screenshot from 2022-07-08 11-36-52

jeanphilippegg:
I am fine with the results! I just don't understand why the first 3 items are fontified.

jeanphilippegg:
You can also this with "xxxxxx20220707T070707xxxxxxx.org" and we get fontification of the date+time. So I suppose if the whole regexp matches (only the date+time is mandatory) even a part of the file name, then we fontify. That's good with me.

jeanphilippegg:
I tried this and I got the exact behavior we have in mind:

```elisp

(defvar denote-faces--file-name-regexp-with-title-and-keywords
  (concat "\\(?1:[0-9]\\{8\\}\\)\\(?2:T[0-9]\\{6\\}\\)"
          "\\(?:\\(?3:--\\)\\(?4:[0-9A-Za-z-]*\\)\\)?"
          "\\(?:\\(?5:__\\)\\(?6:[0-9A-Za-z_-]*\\)\\)?"
          "\\(?7:\\..*\\)")
  "Regexp of file names for fontification.")

(defvar denote-faces--file-name-regexp-with-title-only
  (concat "\\(?1:[0-9]\\{8\\}\\)\\(?2:T[0-9]\\{6\\}\\)"
          "\\(?:\\(?3:--\\)\\(?4:[0-9A-Za-z-]*\\)\\)?"
          "\\(?5:\\..*\\)")
  "Regexp of file names for fontification.")

(defvar denote-faces--file-name-regexp-with-keywords-only
  (concat "\\(?1:[0-9]\\{8\\}\\)\\(?2:T[0-9]\\{6\\}\\)"
          "\\(?:\\(?3:__\\)\\(?4:[0-9A-Za-z_-]*\\)\\)?"
          "\\(?5:\\..*\\)")
  "Regexp of file names for fontification.")

(defvar denote-faces--file-name-regexp-without-title-and-keywords
  (concat "\\(?1:[0-9]\\{8\\}\\)\\(?2:T[0-9]\\{6\\}\\)"
          "\\(?3:\\..*\\)")
  "Regexp of file names for fontification.")

(defconst denote-faces-file-name-keywords
  `((,denote-faces--file-name-regexp-with-title-and-keywords
     (1 'denote-faces-date)
     (2 'denote-faces-time)
     (3 'denote-faces-delimiter nil t)
     (4 'denote-faces-title nil t)
     (5 'denote-faces-delimiter nil t)
     (6 'denote-faces-keywords nil t)
     (7 'denote-faces-extension))
    (,denote-faces--file-name-regexp-with-title-only
     (1 'denote-faces-date)
     (2 'denote-faces-time)
     (3 'denote-faces-delimiter nil t)
     (4 'denote-faces-title nil t)
     (5 'denote-faces-extension))
    (,denote-faces--file-name-regexp-with-keywords-only
     (1 'denote-faces-date)
     (2 'denote-faces-time)
     (3 'denote-faces-delimiter nil t)
     (4 'denote-faces-keywords nil t)
     (5 'denote-faces-extension))
    (,denote-faces--file-name-regexp-with-title-and-keywords
     (1 'denote-faces-date)
     (2 'denote-faces-time)
     (3 'denote-faces-extension)))
  "Keywords for fontification of file names.")
```
But I am fine with the simpler approach!

protesilaos:
> I tried this and I got the exact behavior we have in mind: [...] But I am fine with the simpler approach!

I am fine either way.  What you share here is a bit more elaborate, but not too complex for us to dismiss a priori.  I actually prefer if we do the right thing in all cases.

I say we apply your code.  If it works, we stick with it.  If we discover some problem with it, we fix it.  If the complexity ends up being too much, we revert to the simpler techniques we already discovered.

What do you think?  If you agree, do you want to prepare a PR for the `[file-name-permutations` branch or shall I commit it from here with you as the author?

jeanphilippegg:
> prepare a PR for the `[file-name-permutations` branch

Done! See pull request #39 .

PR #39, opened 2022-07-08
Fontification more precise

jeanphilippegg:


protesilaos:
Thanks!

Shouldn't this have `nil t` as well?

```
(5 'denote-faces-extension)
```

jeanphilippegg:
I removed it because I thought we wanted an extension. You can add it if you want to fontify even when there is no extension.

protesilaos:
> I removed it because I thought we wanted an extension. 

We only need an extension for what we consider "notes".  Otherwise it is okay to have a file without an extension (it would normally happen with the Denote rename command).

protesilaos:
Yes/

jeanphilippegg:
This pull request does not bring anything if the extension is optional. I am closing it.

protesilaos:
> This pull request does not bring anything if the extension is optional. I am closing it.

Okay then.

PR #40, opened 2022-07-09
Adjust fontification

jeanphilippegg:
Some adjustments:
- I used `$` for the end of the line.
- I used " " for the start of the filename in Dired (different for the backlinks buffer).


protesilaos:
Merged.  Thank you!

Let's try this for a while and see how it goes.

protesilaos:
I just added this:

```elisp
(define-obsolete-variable-alias
  'denote-faces-file-name-with-subdir-keywords
  'denote-faces-file-name-keywords-for-backlinks
  "0.3.0")
```

We will remove it at a later version.

PR #42, opened 2022-07-10
Remove else part in denote--file-name-relative-to-denote-directory

jeanphilippegg:
We have had this discussion before, but the else part in the `if-let*` in `denote--file-name-relative-to-denote-directory` does not seem necessary. You mentioned that you had issues coming from `xref` without it. I find this strange, but I added `expand-file-name` just to be sure. That is the only thing I would imagine that could lead to the else part being executed. You should test your backlinks just to be sure!

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -370,11 +370,11 @@ FILE is relative to the variable `denote-directory'."
 (defun denote--file-name-relative-to-denote-directory (file)
   "Return file name of FILE relative to the variable `denote-directory'.
 FILE must be an absolute path."
-  (if-let* ((dir (denote-directory))
-            ((file-name-absolute-p file))
-            ((string-prefix-p dir file)))
-      (substring-no-properties file (length dir))
-    file))
+  (when-let* ((dir (denote-directory))
+              ((file-name-absolute-p file))
+              (file-name (expand-file-name file))
+              ((string-prefix-p dir file-name)))
+    (substring-no-properties file-name (length dir))))
 
 (defun denote--current-file-is-note-p ()
   "Return non-nil if current file likely is a Denote note."


jeanphilippegg:
Or maybe, it can be called with a file that is not in `denote-directory`?

protesilaos:
I think this may be related to [issue 43](https://github.com/protesilaos/denote/issues/43).  Working on it.

protesilaos:
For me this works (the check for the absolute is not needed):

```elisp
(defun denote--file-name-relative-to-denote-directory (file)
  "Return file name of FILE relative to the variable `denote-directory'.
FILE must be an absolute path."
  (when-let* ((dir (denote-directory))
              (file-name (expand-file-name file))
              ((string-prefix-p dir file-name)))
    (substring-no-properties file-name (length dir))))
```

jeanphilippegg:
> (the check for the absolute is not needed):

You are right. If it is passed "some-file.org", it will return `nil`, with or without the check for absolute.

jeanphilippegg:
Humm, I think it is not right.

`(denote--file-name-relative-to-denote-directory "some-file.org"` will return either `nil` if it is outside the `denote-directory` or `<path to current dir>/some-file.org` if inside some dir in `denote-directory`.

We need to check that `file` is absolute (or that `file-name` is actually a file).

protesilaos:
> if it is outside the `denote-directory`

We do not support that workflow.  The links must be within the `denote-directory`.

> We need to check that `file` is absolute (or that `file-name` is actually a file)

Okay.

jeanphilippegg:
Calling `denote--file-name-relative-to-denote-directory` with a relative path is not supposed to happen, but if it does I think we should always return nil to avoid confusion when we will debug it.

jeanphilippegg:
Here is an example of a strange behavior if we do not check for absolute path:

Say I have those files:

```
denote-dir/
denote-dir/note1.txt
denote-dir/subdir/note2.txt
```
If I stand in the `subdir` directory and call `(denote--file-name-relative-to-denote-directory "subdir/note2.txt")`, I get `/home/user/denote-dir/subdir/subdir/note2.txt`.

protesilaos:
> Say I have those files: [...]

Maybe we should not use the substring extraction?  Perhaps we can abbreviate the path to leave out the portion that is the `denote-directory?  There is `abbreviate-file-name`, though we would need to adapt it to our needs.

jeanphilippegg:
Yes we could, but I am not sure it would help when our funciton is called with a relative path.

protesilaos:
I accidentally posted my subsequent comments in the wrong thread.  Copying them here:

* * *
Try this to see what the problem is:

```elisp
(defun denote--file-name-relative-to-denote-directory (file)
  "Return file name of FILE relative to the variable `denote-directory'.
FILE must be an absolute path."
  (when-let* ((dir (denote-directory))
              (file-name (expand-file-name file))
              ((string-prefix-p dir file-name)))
    ;; (substring-no-properties file-name (length dir))
    (format "%s %s" dir file-name)
    ))
```

The `expand-file-name` without a check for an absolute path.

* * *

Indeed, `(expand-file-name "subdir/20220710T210519--a-nested-note__test.txt")` returns the wrong path.

***

Whereas this returns the expected value (only tested this scenario):

```elisp
(defun denote--file-name-relative-to-denote-directory (file)
  "Return file name of FILE relative to the variable `denote-directory'.
FILE must be an absolute path."
  (when-let* ((dir (denote-directory))
              (file-name (expand-file-name (file-name-nondirectory file)))
              ((string-prefix-p dir file-name)))
    ;; (substring-no-properties file-name (length dir))
    (format "%s %s" dir file-name)
    ))
```

jeanphilippegg:
It still get the same strange results.

Say I have those files:

```
denote-dir/
denote-dir/note1.txt
denote-dir/subdir/note2.txt
denote-dir/subdir/subsubdir/note3.txt
```

If I stand in the `subsubdir` directory and call `(denote--file-name-relative-to-denote-directory "subdir/note2.txt")`, I get `/home/user/denote-dir/subdir/subsubdir/subdir/note2.txt`, which does not exist.

protesilaos:
> It still get the same strange results.

I see.  Then the `file-name-absolute-p` has to be preserved.

I must then figure out why your PR gives me a nil value and does not produce the backlinks' buffer.

jeanphilippegg:
If `denote--file-name-relative-to-denote-directory` is not meant to be called with relative paths, I think we should check with `file-absolute-p` to return `nil` instead of other strange results. This will be easier to debug later if it happens to be called with a relative paths even it is not supposed to.

I don't know if we can support relative paths, because juste `some-dir/some-file.txt` does not give enough information to retrieve a valid path. And we don't know what kind of relative path is going to be given to `denote--file-name-relative-to-denote-directory` because it is not supposed to happen.

protesilaos:
>  is not meant to be called with relative paths

Agreed.  We cross-posted as I was commenting on the need for absolute paths.

jeanphilippegg:
> I must then figure out why your PR gives me a nil value and does not produce the backlinks' buffer.

Yes, this is strange that you get `nil`. I am interested to know the cause.

protesilaos:
> I am interested to know the cause.

I am almost falling asleep now, so this may be fuzzy.  Your revised `denote--file-name-relative-to-denote-directory` returns nil.  This is then fed to `denote-retrieve--files-in-output` which runs `denote--only-note-p` on a nil value.  Hence the error.

jeanphilippegg:
You can try this:

```elisp
(defun denote--file-name-relative-to-denote-directory (file)
  "Return file name of FILE relative to the variable `denote-directory'.
FILE must be an absolute path."
  (if-let* ((dir (denote-directory))
            ((file-name-absolute-p file))
            ((string-prefix-p dir file)))
      (substring-no-properties file (length dir))
    (print file)
    file))
```
But it can wait until tomorrow!

protesilaos:
It echoes the file names.  Though that is the else part of the `if-let*`.

```
"20220614T130812--on-being-honest__journal.txt"

"20220614T145606--let-this-glance-become-a-stare__journal.txt"

"20220616T182958--feeling-butterflies-in-your-stomach__journal.txt"

"testting/20220708T075900--this-is-testing__test.txt"
```

> But it can wait until tomorrow!

Thanks!  I think I can stay a while longer.  I am curious to see how this unfolds.

jeanphilippegg:
Yes, this is what I wanted to see. It means your `denote--file-name-relative-to-denote-directory` is called with a relative path. This should not happen. These should be absolute!

jeanphilippegg:
For example, the same code on my side does not print anything, because the else part is not reached, because file is always absolute.

protesilaos:
> These should be absolute!

Exactly!  So the problem is somewhere in denote-retrieve.el

jeanphilippegg:
My next guest would be to check the `xrefs` in `denote-retrieve--files-in-xref`.

protesilaos:
The `denote-retrieve--xrefs` returns absolute paths.  It has to be `denote-retrieve--files-in-xrefs`.

protesilaos:
```elisp
(defun denote-retrieve--files-in-xrefs (xrefs)
  "Return sorted file names sans directory from XREFS.
Parse `denote-retrieve--xrefs'."
  (sort
   (mapcar (lambda (x)
             (denote--file-name-relative-to-denote-directory (car x)))
           xrefs)
   #'string-lessp))
```

The `(car x)` is the wrong value when a subdir exists.

jeanphilippegg:
Hum, I have subdirs and `(car x)` is the right value. What kind of value do you have in `xrefs`?

protesilaos:
When `denote-retrieve--xrefs` includes `xref--analyze`

```
(("20220614T130812--on-being-honest__journal.txt" #s(xref-match-item #("identifier: 20220614T130812" 12 27 ...) #s(xref-file-location "/home/prot/Documents/notes/20220614T130812--on-being-honest__journal.txt" 4 12) 15)) ("20220614T145606--let-this-glance-become-a-stare__journal.txt" #s(xref-match-item #("into it: [[denote:20220614T130812][On being honest]].  Michelle" 18 33 ...) #s(xref-file-location "/home/prot/Documents/notes/20220614T145606--let-this-glance-become-a-stare__journal.txt" 37 18) 15) #s(xref-match-item #("As I said in [[denote:20220614T130812][On being honest]] I have never" 22 37 ...) #s(xref-file-location "/home/prot/Documents/notes/20220614T145606--let-this-glance-become-a-stare__journal.txt" 64 22) 15)) ("20220616T182958--feeling-butterflies-in-your-stomach__journal.txt" #s(xref-match-item #("indifference.  In [[denote:20220614T130812][On being honest]] I" 27 42 ...) #s(xref-file-location "/home/prot/Documents/notes/20220616T182958--feeling-butterflies-in-your-stomach__journal.txt" 64 27) 15)) ("testting/20220708T075900--this-is-testing__test.txt" #s(xref-match-item #("[[denote:20220614T130812][On being honest]]" 9 24 ...) #s(xref-file-location "/home/prot/Documents/notes/testting/20220708T075900--this-is-testing__test.txt" 7 9) 15)))
```

When `denote-retrieve--xrefs` does not include the `xref--analyze`.

```
(#s(xref-match-item #("identifier: 20220614T130812" 12 27 (face xref-match)) #s(xref-file-location "/home/prot/Documents/notes/20220614T130812--on-being-honest__journal.txt" 4 12) 15) #s(xref-match-item #("into it: [[denote:20220614T130812][On being honest]].  Michelle" 18 33 (face xref-match)) #s(xref-file-location "/home/prot/Documents/notes/20220614T145606--let-this-glance-become-a-stare__journal.txt" 37 18) 15) #s(xref-match-item #("As I said in [[denote:20220614T130812][On being honest]] I have never" 22 37 (face xref-match)) #s(xref-file-location "/home/prot/Documents/notes/20220614T145606--let-this-glance-become-a-stare__journal.txt" 64 22) 15) #s(xref-match-item #("indifference.  In [[denote:20220614T130812][On being honest]] I" 27 42 (face xref-match)) #s(xref-file-location "/home/prot/Documents/notes/20220616T182958--feeling-butterflies-in-your-stomach__journal.txt" 64 27) 15) #s(xref-match-item #("[[denote:20220614T130812][On being honest]]" 9 24 (face xref-match)) #s(xref-file-location "/home/prot/Documents/notes/testting/20220708T075900--this-is-testing__test.txt" 7 9) 15))
```

The latter looks okay.  It is `xref--analyze` that produces `"testting/20220708T075900--this-is-testing__test.txt"`

jeanphilippegg:
You are right. The group name is absolute for me.

protesilaos:
A quick look at `xref--analyze` suggests that project.el plays a role in the output.  My notes are a "project".  Yours are not?

protesilaos:
Maybe this will do?

```elisp
(defun denote-retrieve--xrefs (identifier)
  "Return xrefs of IDENTIFIER in variable `denote-directory'.
The xrefs are returned as an alist."
  (xref--alistify
   (xref-matches-in-files identifier (denote--directory-files :absolute))
   (lambda (x)
     (xref-location-group (xref-item-location x)))))
```

protesilaos:
The revised `denote-retrieve--xrefs` works as expected with the code of your PR here.  It was giving me nil before.

jeanphilippegg:
This might fix issue #43 . This user seems to have a project in a parent directory, hence his errors.

protesilaos:
Yes.

protesilaos:
Just pushed commit `3074133`.  Will now merge this as well.

protesilaos:
Oh, I thought this PR was for main... Do you think we can put it there instead, without the rest of the refactored prompts?  (will do those tomorrow).

jeanphilippegg:
Yes!

protesilaos:
I cherry-picked it into main and pushed to the remote.  If this duplication causes any problems in the `refactor-denote-prompts` branch, we can fix it.  Sorry about that---I clicked to merge here without checking the target branch.

jeanphilippegg:
I think we should delete dup as well!

protesilaos:
> I think we should delete dup as well!

Sounds good!

jeanphilippegg:
I deleted my comment, but you commented fast. I wanted to say that we should delete dups in `denote-retrieve--files-in-xrefs`.

protesilaos:
Okay.  Yes, duplicates are not needed.

jeanphilippegg:
I will!

PR #45, opened 2022-07-11
Specify existing face for backlink buttons.

jeanphilippegg:
This is an old minor bug.

I get this in `*Messages*` whenever the backlinks buffer is open:

```
Invalid face reference: unspecified [1331 times]
```

Nothing was broken though. This pull request removes the messages.

### DIFF
--- a/denote-link.el
+++ b/denote-link.el
@@ -367,10 +367,13 @@ positions, limit the process to the region in-between."
 
 ;;;; Backlinks' buffer
 
+(defface denote-link--backlink-button nil
+  "Face for backlink buttons.")
+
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged.  Thank you!

protesilaos:
We need to make a change here because the `defface` is public and we really only need this for internal purposes.

This does not give me any errors:

```elisp
(define-button-type 'denote-link-backlink-button
  'follow-link t
  'action #'denote-link--backlink-find-file
  'face nil)
```

Please confirm.

jeanphilippegg:
It also works for me!

protesilaos:
Applied.  Thank you!

Issue #43, opened 2022-07-10
denote-retrieve--files-in-xrefs does not return correct file names.


pRot0ta1p:
Hi, I was testing out denote and I found the backlink does not work, it always says "No links to the current note",
it appears to work in your demo video so I tried to debug the code, long story short I found this piece of code:
https://github.com/protesilaos/denote/blob/0683a1937c78068e79463df51acf936da338d301/denote-retrieve.el#L114-L115
seems to return wrong path of notes.

To test it, I create 2 notes in a empty directory `~/test` by invoke `M-x denote`, `note2.org` have a denote link to `note1.org`, 
then I `M-x eval-expression` the following codes in the `note1.org` buffer that should have a backlink to it:

```emacs-lisp
(delete (denote--file-name-relative-to-denote-directory (buffer-file-name))
(denote-retrieve--files-in-xrefs
 (denote-retrieve--xrefs (denote-retrieve--filename-identifier (buffer-file-name)))))
```

It returns something like `("test/20220710T200942--note1__random.org" "test/20220710T201012--note2__random.org")`, 
when function `denote-retrieve--files-in-output` receive those as arguments, it returns nil, so no backlinks...

I run the same test with emacs28.1 and emacs29 from a recent commit after `emacs -Q`, both shows the same results.

protesilaos:
Hello @pRot0ta1p,

I cannot reproduce your issue.  I placed this in the scratch buffer of
`emacs -Q` (using Emacs 29 on GNU/Linux):

```elisp
(add-to-list 'load-path "/home/prot/Git/Projects/denote/")
(progn
  (require 'denote)
  (require 'denote-retrieve)
  (require 'denote-faces)
  (require 'denote-link)

  (setq denote-directory "~/testing-denote"))
```

I then did `M-x denote` and followed the prompts for title and keywords
to create a new file.  I saved it.

Then I did `M-x denote` again and created a second file.

From the second file, I did `M-x denote-link` and in the minibuffer
prompt I selected the first file.

I then switched to the first file and did `M-x denote-link-backlinks`.
It produced a backlinks' buffer, with a link to the second file.
Screenshot:

![Screenshot from 2022-07-10 17-22-42](https://user-images.githubusercontent.com/12828033/178149292-7bfc5f66-f48c-48ae-b638-5593639a6793.png)


While in the buffer of the first file, I evaluated the code you
provided:

```elisp
(delete (denote--file-name-relative-to-denote-directory (buffer-file-name))
[...]

PR #46, opened 2022-07-11
Expand file name of denote-directory

jeanphilippegg:
I was sure the function `denote-directory` was returning the expanded file name, but as reported in issue #43, it is not the case.

This pull request fixes that.


pRot0ta1p:
Can confirm this works in my setup, thank you!

protesilaos:
Merged.  Thank you!

[ This must be a recent regression.  I am sure it expanded the file name before.  It is the reason for its existence. ]

Issue #48, opened 2022-07-14
Convert plain file to Denote


jeanphilippegg:
Based on the `denote-dired-rename-file` function, I came up with this function to convert an existing non-Denote file into a Denote one.

The idea is not to try to convert org-roam files to Denote, though it can probably be useful for that use case. Check the docstring of the function to know the details.

I am sharing it here because I don't know if it should be integrated to the code or where to integrate it.

What do you think?

```elisp
(defun denote-dired-convert-file-to-denote (file title keywords)
  "Convert a file to Denote format.

This function adds a front matter unconditionally at the start of
the file. It does not check if one is already present. If a front
matter is already present, you will end up with two. Consider
using `denote-dired-rename-file' in that case.

It prompts for a FILE, a TITLE and KEYWORDS.

The identifier is retrieved from the filename if there is one,
else the last modification time is used.

The default title is retrieved from a line starting with
\"#+title:\" if such a line exists (and depending on the file
type). Else, the file name is used."
  (interactive
   (let ((file (denote-dired--rename-file-is-regular (denote-dired--rename-dired-file-or-prompt))))
     (list
      file
      (denote--title-prompt (or (denote-retrieve--value-title file)
                                (file-name-sans-extension (file-name-nondirectory file))))
      (denote--keywords-prompt))))
  (let* ((dir (file-name-directory file))
         (old-name (file-name-nondirectory file))
         (id (denote-dired--file-name-id file))
         (extension (file-name-extension file t))
         (new-name (denote--format-file
                    dir
                    (denote-dired--file-name-id file)
                    keywords
                    (denote--sluggify title)
                    extension))
         (max-mini-window-height 0.33)) ; allow minibuffer to be resized
    (unless (string= old-name (file-name-nondirectory new-name))
      (when (y-or-n-p
             (format "Rename %s to %s?"
                     (propertize old-name 'face 'error)
                     (propertize (file-name-nondirectory new-name) 'face 'success)))
        (rename-file old-name new-name nil)
        (denote-dired--rename-buffer old-name new-name)
        (denote-dired-update-dired-buffers)))
    (when-let* ((filetype (denote-dired--filetype-heuristics file))
                (date (denote--date (date-to-time id)))
                (new-front-matter (denote--file-meta-header title date keywords id filetype)))
      (with-current-buffer (find-file-noselect new-name)
        (goto-char (point-min))
        (insert new-front-matter)))))
```

jeanphilippegg:
[...]

PR #49, opened 2022-07-15
Add denote-dired-convert-file-to-denote

jeanphilippegg:
Pull request as discussed in #48.

I added the function in `denote-dired.el`. There is some code duplication between it and `denote-dired-rename-file`.

### DIFF
--- a/denote-dired.el
+++ b/denote-dired.el
@@ -354,6 +354,55 @@ appropriate."
           ;; those.
           (delete-blank-lines))))))
 
+(defun denote-dired-convert-file-to-denote (file title keywords)
+  "Convert a file to Denote format.
+
+This function adds a front matter unconditionally at the start of
+the file. It does not check if one is already present. If a front
+matter is already present, you will end up with two. Consider
+using `denote-dired-rename-file' in this case.
+
+It prompts for a FILE, a TITLE and KEYWORDS.
+
+The identifier is retrieved from the filename if there is one,
+else the last modification time is used.
+
+The default title is retrieved from a line starting with
+\"#+title:\" if such a line exists (and depending on the file
+type). Else, the file name is used."
+  (interactive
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you!

> There is some code duplication

Do you think we should consider this as part of the PR or do it afterwards?  Whatever helps your case.

jeanphilippegg:
> Do you think we should consider this as part of the PR or do it afterwards? Whatever helps your case.

While the code looks similar, it is not exactly the same. I would not know the best way to extract it.

protesilaos:
> I would not know the best way to extract it.

No worries!  I merge this now and we can think about it later.  It is not strictly necessary, anyway.

jeanphilippegg:
> No worries! I merge this now and we can think about it later. It is not strictly necessary, anyway.

Ok, good! Thanks

protesilaos:
I just made some tweaks to the doc string:

```
  "Convert a file to Denote format.

Prompt for a FILE, a TITLE, and KEYWORDS.  The unique identifier
is retrieved from the filename if there is one, else the last
modification time is used to derive one.

In Dired, FILE is the one at point.  It has to be a regular
file (not a directory, a named pipe, etc.).

The default title is retrieved from a line starting with a title
field, depending on the given file type.  Else, the file name is
used.

This command adds a front matter unconditionally at the start of
the file.  It does not check if one is already present.  If a
front matter is already present, you will end up with two.
Consider using `denote-dired-rename-file' in this case.

For per-file-type front matter, refer to the variables:

- `denote-org-front-matter'
- `denote-text-front-matter'
- `denote-toml-front-matter'
- `denote-yaml-front-matter'"
```

Do you think it is okay?

[ I will update the manual later. ]

jeanphilippegg:
Yes. I like it!

PR #51, opened 2022-07-17
Renaming functions

jeanphilippegg:
> DEV NOTE 2022-07-16: proof of concept---help flesh it out.

Ok!

What would you say about these proposed functions (for more precision)?

```elisp
(defun denote-dired-rename-file-and-rewrite-front-matter (file title keywords) ...)
    ; Same as the old `denote-dired-rename-file`. I added it as an alias. We could also deprecate it.
(defun denote-dired-rename-file-and-add-front-matter (file title keywords) ...)
    ; Same as the old `denote-dired-convert-file-to-denote`
(defun denote-dired-rename-marked-files () ...)
    ; Same as your new `denote-dired-batch-rename`.
```

- In all of them, marked files are used.
- In my pull request, they are all implemented similarly. Since we specify that `denote-dired-rename-file-and-rewrite-front-matter` rewrites the front matter, it does not make sense to have it be part of a user option `denote-dired-post-rename-functions`. I added an hyphen in the name of `denote-dired--rewrite-front-matter` to make it internal.
- The other function `denote-dired-update-dired-buffers` in `denote-dired-post-rename-functions` is called directly as part the new internal function `denote-dired--rename-file`. This leaves nothing in the user option `denote-dired-post-rename-functions`. This pull request removes it.
- In `denote-dired-rename-marked-files`, I have added the code to retrieve the title from a line starting with `#+title:`. I know it must open the files and it can be slow, but since it is a command meant for interactive use, I am not sure the user will select that many files.
- I moved the functions around and I renamed the two sections of the code: `File helper functions` and `Renaming commands`.

This is just a proof-of-concept to show you what this could look like and get your feedback. You don't have to merge it.

### DIFF
--- a/denote-dired.el
+++ b/denote-dired.el
@@ -422,11 +422,11 @@ Rename marked files in Dired using the following pattern:
 Batch renaming ignores files that comply with Denote's
 file-naming scheme."
   (interactive nil dired-mode)
-  (if-let ((marks (dired-get-marked-files)))
+  (if-let ((marks (dired-get-marked-files))
+           (keywords (denote--keywords-prompt)))
       (progn
         (dolist (file marks)
-          (let* ((keywords (denote--keywords-prompt))
-                 (dir (file-name-directory file))
+          (let* ((dir (file-name-directory file))
                  (id (denote-dired--file-name-id file))
                  (title (or (denote-retrieve--value-title file)
                             (file-name-sans-extension
@@ -434,8 +434,8 @@ file-naming scheme."
                  (extension (file-name-extension file t))
                  (new-name (denote--format-file
                             dir id keywords (denote--sluggify title) extension)))
-            (unless (denote--only-note-p file)
-              (rename-file (file-name-nondirectory file) new-name))))
+            (when (denote--only-note-p file)
+              (rename-file file new-name))))
         (revert-buffer))
     (user-error "No marked files; aborting")))
 
--- a/denote-dired.el
+++ b/denote-dired.el
@@ -434,8 +434,7 @@ file-naming scheme."
                  (extension (file-name-extension file t))
                  (new-name (denote--format-file
                             dir id keywords (denote--sluggify title) extension)))
-            (when (denote--only-note-p file)
-              (rename-file file new-name))))
+            (rename-file file new-name)))
         (revert-buffer))
     (user-error "No marked files; aborting")))
 
--- a/denote-dired.el
+++ b/denote-dired.el
@@ -357,7 +357,7 @@ attachments that the user adds to their notes."
 
 ;;;###autoload
 (defun denote-dired-rename-file-and-add-front-matter (file title keywords)
-  "Convert a file to Denote format.
+  "Rename FILE and add a front matter.
 
 Prompt for a FILE, a TITLE, and KEYWORDS.  The unique identifier
 is retrieved from the filename if there is one, else the last
--- a/denote-dired.el
+++ b/denote-dired.el
@@ -352,6 +352,9 @@ attachments that the user adds to their notes."
     (when (denote-dired--rename-file file new-name)
       (denote-dired--rewrite-front-matter new-name title keywords))))
 
+(defalias 'denote-dired-rename-file
+  (symbol-function 'denote-dired-rename-file-and-rewrite-front-matter))
+
 ;;;###autoload
 (defun denote-dired-rename-file-and-add-front-matter (file title keywords)
   "Convert a file to Denote format.
--- a/denote-dired.el
+++ b/denote-dired.el
@@ -143,6 +143,29 @@ old name followed by the new one."
   :type 'boolean
   :group 'denote-dired)
 
+(defcustom denote-dired-post-rename-functions
+  (list #'denote-dired-update-dired-buffers
+        #'denote-dired--rewrite-front-matter)
+  "List of functions called after `denote-dired-rename-file'.
+Each function must accept three arguments: FILE, TITLE, and
+KEYWORDS.  The first is the full path to the file provided as a
+string, the second is the human-readable file name (not what
+Denote sluggifies) also as a string, and the third are the
+keywords.  If there is only one keyword, it is a string, else a
+list of strings.
+
+DEVELOPMENT NOTE: the `denote-dired-rewrite-front-matter' needs
+to be tested thoroughly.  It rewrites file contents so we have to
+be sure it does the right thing.  To avoid any trouble, it always
+asks for confirmation before performing the replacement.  This
+confirmation ignores `denote-dired-rename-expert' for the time
+being, though we might want to lift that restriction once
+everything works as intended."
+  :type 'hook
+  :group 'denote-dired)
+
+(make-obsolete 'denote-dired-post-rename-functions nil "0.4.0")
+
 ;;;; File helper functions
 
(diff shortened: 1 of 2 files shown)


protesilaos:
Thank you!  I will review it now.

Note though that I just pushed commit `320c687` on your behalf, because it is an error we need to fix right away.

protesilaos:
We need a `make-obsolete` for the deprecated defcustom.  We have a sample in the code:

```elisp
(make-obsolete 'denote-front-matter-date-format 'denote-date-format "0.2.0")
```

The idea with those is to warn those who are using them.  We can then remove them a couple of releases later.

About renaming existing commands, we need to use `define-obsolete-function-alias` so that those who have the old keybindings will get the warning.  Alternatively, we can keep the old names to avoid breakage and write `defalias` ones.  I am not including the "batch rename" in this: we can do whatever we want with it, as it is in development.

With those granted, I think this looks good to go.  (Will double-check if I missed some detail.)

denote-dired.el
```diff
@@ -440,21 +422,20 @@ Rename marked files in Dired using the following pattern:
...
                  (new-name (denote--format-file
-                            dir
-                            (denote-dired--file-name-id file)
-                            keywords
-                            (denote--sluggify title)
-                            extension)))
-            (unless (denote--only-note-p file)
-              (rename-file (file-name-nondirectory file) new-name))))
+                            dir id keywords (denote--sluggify title) extension)))
+            (when (denote--only-note-p file)
```
protesilaos, line 437: The `when` instead of `unless` changes the effect here, no?  The original plan was to ignore files that already conform with the file-naming scheme

jeanphilippegg: Yes, I just noticed as well and I pushed a commit to remove the condition. I don't think we need it. It will handle Denote notes correctly, I think. Only the keywords will be changed in this case.

protesilaos: Okay.  Yesterday when I was trying it, I had to add the `unless` otherwise it kept renaming the file.

jeanphilippegg: It is working fine now?

jeanphilippegg:
> We need a `make-obsolete` for the deprecated defcustom.

`make-obsolete` asks for a `current-name`, but we are just removing the function, not changing its name.

protesilaos:
Sorry, forgot to mention it: the current-name can be nil as far as I know.  Otherwise we add a string with an explanation.

jeanphilippegg:
Thanks! I pushed the changes.

protesilaos:
Thank you!  We need to consider the implications of this in `denote-dired-rename-marked-files`:

```elisp
(title (or (denote-retrieve--value-title file)
           (file-name-sans-extension
            (file-name-nondirectory file))))
```

We would need to add a check , maybe in `denote-retrieve--value-title` or its subroutine, so that it does not try to read images, videos, etc.

jeanphilippegg:
Maybe `denote-retrieve--search` should return nil for "binary" files. Do you know how we can do that?

jeanphilippegg:
I think we can just check that the file ends with one of Denote's file extension. We can't look for a `#+title:` line for other file types anyway.

protesilaos:
> I think we can just check that the file ends with one of Denote's file extension.

Yes!

jeanphilippegg:
> EDIT: I don't know about binary files.

A "binary" file is just a file that is not a text file. For example, PDFs, mp4, png, etc. are said to contain binary data, while TXT, EL, etc. contain just text.

I will push something soon to check the extension.

protesilaos:
> A "binary" file is just a file that is not a text file.

Sorry, for not explaining: I meant I had no answer to your question, as in not knowing of such a function.

> I will push something soon to check the extension.

Okay!

protesilaos:
Thanks!  Note that I get this message on my end:

Screenshot from 2022-07-17 08-58-51

jeanphilippegg:
> Sorry, for not explaining: I meant I had no answer to your question, as in not knowing of such a function.

Oh, sorry for not understanding! I hope I have not offended you, haha.

I pushed the change. The merge conflict is cause be the commit you took from my branch to fix the `denote-dired-batch-rename`. You can resolve the conflict by taking everything from my branch.

protesilaos:
> I hope I have not offended you, haha.

Not at all, haha!

> You can resolve the conflict by taking everything from my branch.

What do you mean by "taking"?  Which git command?

jeanphilippegg:
I am not confortable with git merge conflicts myself. What I meant is that when merging the branch, there will be a conflict and in denote-dired.el, you can choose to keep the code on my branch.

But another way this can be solved that is probably simpler is to reset `main` to `ee53a10`. This will delete your last commit, which is already included in my branch if you are okay dropping a commit from the history.

jeanphilippegg:
I think I could do it myself (resolving the conflicting merge), if you prefer!

protesilaos:
> I think I could do it myself (resolving the conflicting merge), if you prefer!

I can do it, no worries!  Unless you already started...

jeanphilippegg:
Yes. I add resolved the conflict, but I did not want to create a mess in case you add started. I pushed it!

protesilaos:
Merged.  Thank you!

protesilaos:
Do you think there is something else we can do for `denote-dired-rename-marked-files` or, more generally, for bulk renaming?

jeanphilippegg:
You're welcome!

protesilaos:
We cross-commented, so you may miss this:

Do you think there is something else we can do for `denote-dired-rename-marked-files` or, more generally, for bulk renaming?

jeanphilippegg:
> Do you think there is something else we can do for `denote-dired-rename-marked-files` or, more generally, for bulk renaming?

Maybe when I try to use it for my own files, I will discover something, but I do not know for now.

protesilaos:
Okay!  I leave the "DEV NOTE" for now and will not document the command yet.

jeanphilippegg:
Ok, I just tested with a copy of my files. I will probably use the function `denote-rename-file-and-add-front-matter` because I will need a front matter. I realise I would probably prefer a function `denote-rename-marked-files-and-add-front-matter`.

protesilaos:
> I realise I would probably prefer a function `denote-rename-marked-files-and-add-front-matter`.

Yes, I had thought about that as well.  But I decided that it is better to go with this for a start.  The other will rewrite files, so we need to consider the implications (e.g. someone migrates from org-roam---how does that look?).

At any rate, such a command would be useful.  Though the rewriting of the front matter will not be asked each time.  We will have to have a single "Yes or No" with a basic question "Are you sure you want to prepend front matter to FILES?".

jeanphilippegg:
That's good with me!

protesilaos:
Okay!  I will give it a try later today.  Unless you want to do it.

protesilaos:
Hello again @jeanphilippegg!

In `7af5123` I removed the obsoletion of the symbol `denote-dired-rename-file`.  I was trying to update the documentation and found the longer name to be confusing.  The problem was that it was difficult to differentiate it from `denote-dired-rename-file-and-add-front-matter`.

I updated all documentation to clarify what each does.  Are you okay with this?  I thought it was easier to write it and have you review it than start a discussion here.  But do not consider it a fait accompli.

jeanphilippegg:
> I updated all documentation to clarify what each does. Are you okay with this?

Yes, I think I like it. It is coherent with the `denote-dired-rename-marked-files*` functions.

PR #53, opened 2022-07-17
Renaming commands, minor adjustments

jeanphilippegg:
I moved the check for `denote--only-note-p` in `denote-dired--add-front-matter`.

I added an `s` to `denote-dired-rename-marked-files-and-add-front-matters`. Does it make sense?


protesilaos:
Thank you!  I think the `s` at the end is not needed, though I am indifferent.  Merging it.

PR #55, opened 2022-07-18
Refactor denote--prepare-notes and other modifications

jeanphilippegg:
`denote--prepare-note` is only used in 2 places: in `denote` and for org-capture. In `denote`, we let-bind `denote-file-type` and `denote-directory`. In this pull request, I remove the let-binding and pass the file-type and the directory directly to `denote--prepare-note`.

For org-capture, I noticed that the only thing `denote--prepare-note` was doing is set `denote-last-front-matter`.

---

`denote--id-exists-p` should also check the active buffers because a new note might not be saved yet.

---

I added 2 internals functions: `denote--get-all-used-ids` and `denote--find-first-available-id`. They are not used anywhere yet, but maybe they can be used to solve the collision issues.

---

I do not understand why we use `:no-check-current` in `denote--barf-duplicate-id`. Why is it needed?

### DIFF
--- a/denote-org-capture.el
+++ b/denote-org-capture.el
@@ -91,7 +91,9 @@ Consult the manual for template samples."
         (keywords (denote--keywords-prompt))
         (denote-file-type nil)) ; we enforce the .org extension for `org-capture'
     (denote--path title keywords)
-    (denote--prepare-note denote-last-title denote-last-keywords denote-last-path)
+    (setq denote-last-front-matter (denote--file-meta-header
+                                    title (denote--date nil) keywords
+                                    (format-time-string denote--id-format nil)))
     (denote--keywords-add-to-history denote-last-keywords)
     (concat denote-last-front-matter denote-org-capture-specifiers)))
 
--- a/denote.el
+++ b/denote.el
@@ -827,11 +827,13 @@ is set to \\'(file-type title keywords)."
   (mapcar
    (lambda (name)
      (file-name-nondirectory name))
-   (delq nil
-         (mapcar
-          (lambda (buf)
-            (buffer-file-name buf))
-          (buffer-list)))))
+   (seq-filter
+    (lambda (name) (denote--only-note-p name))
+    (delq nil
+          (mapcar
+           (lambda (buf)
+             (buffer-file-name buf))
+           (buffer-list))))))
 
 (declare-function cl-some "cl-extra" (cl-pred cl-seq &rest cl-rest))
 
--- a/denote.el
+++ b/denote.el
@@ -687,23 +687,20 @@ With optional DATE, use it else use the current one."
      (t
       (denote--date-org-timestamp date)))))
 
-(defun denote--prepare-note (title keywords &optional path date id)
-  "Use TITLE and KEYWORDS to prepare new note file.
-Use optional PATH, else create it with `denote--path'.  When PATH
-is provided, refrain from writing to a buffer (useful for org
-capture).
+(defun denote--prepare-note (title keywords date id directory file-type)
+  "Prepare a new note file.
 
-Optional DATE is passed to `denote--date', while optional ID is
-used to construct the path's identifier."
-  (let* ((default-directory (denote-directory))
-         (p (or path (denote--path title keywords default-directory id)))
-         (buffer (unless path (find-file p)))
+Arguments TITLE, KEYWORDS, DATE, ID, DIRECTORY, and FILE-TYPE
+should be valid for note creation."
+  (let* ((default-directory directory)
+         (path (denote--path title keywords default-directory id))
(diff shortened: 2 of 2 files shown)


protesilaos:
> `denote--prepare-note` is only used in 2 places: in `denote` and for org-capture. In `denote`, we let-bind `denote-file-type` and `denote-directory`. In this pull request, I remove the let-binding and pass the file-type and the directory directly to `denote--prepare-note`.
> 
> For org-capture, I noticed that the only thing `denote--prepare-note` was doing is set `denote-last-front-matter`.

Some background here: `denote--prepare-note` was just a helper function for `denote`.  But because org-capture does not work well for our purposes, I did some ad-hoc modifications.  The plan was to review those, but we moved on to more pressing issues so it stayed that way.

> `denote--id-exists-p` should also check the active buffers because a new note might not be saved yet.

Agreed.

> I added 2 internals functions: `denote--get-all-used-ids` and `denote--find-first-available-id`.  They are not used anywhere yet, but maybe they can be used to solve the collision issues.

Maybe keep those out of this changeset for now?  Just so we have some more information of what the actual problem is.  Conceptually, we do have a conflict if two notes are created within a second, but do we really need to optimise for that scenario before we have seen any real-world case where this happens?

> I do not understand why we use `:no-check-current` in `denote--barf-duplicate-id`. Why is it needed?

I should have documented the reasoning behind it.  I think it was needed somewhere, but I forgot.  Sorry!

jeanphilippegg:
> Maybe keep those out of this changeset for now? Just so we have some more information of what the actual problem is. Conceptually, we do have a conflict if two notes are created within a second, but do we really need to optimise for that scenario before we have seen any real-world case where this happens?

I agree! I removed the commits.

> I should have documented the reasoning behind it. I think it was needed somewhere, but I forgot. Sorry!

Ok, that's because I did not apply the same reasoning while considering active buffer. The current buffer-file-name will be checked in the active buffers!

protesilaos:
> I agree! I removed the commits.

Thanks!

> Ok, that's because I did not apply the same reasoning while considering active buffers. The current buffer-file-name will be checked in the active buffers!

So it would return a false positive in that case?  Basically not checking for the current one is always necessary, right?

jeanphilippegg:
> So it would return a false positive in that case? Basically not checking for the current one is always necessary, right?

I don't know what is the correct behavior. I just changed `denote--id-exists-p` to look at the active buffers for the requested ID. I don't know why I would want to exclude the current buffer when looking at the active buffer list, so I did not do it, but then this is what is done when looking at the files.

protesilaos:
> I don't know why I would want to exclude the current buffer name when looking at the active buffer list

I am not sure any more.  I think we would want to exclude it to avoid a false positive error in `denote--barf-duplicate-id`?  Not sure though.  We need to do this consistently for files and buffers.

jeanphilippegg:
Ok, I think I will remove the `no-check-current` argument and handle all this directly in `denote--id-exists-p`.

protesilaos:
Yes, that makes sense.

jeanphilippegg:
Done!

protesilaos:
Thank you!  Maybe the `denote--buffer-file-names` should check for a name that matches our file-naming scheme or will that only slow things down?

jeanphilippegg:
What I had in mind for `denote--buffer-file-names` is the general function that returns all buffer-file-names. In principle, this could be used outside of Denote. But we can also make it Denote-specific if you want. I don't know if it is going to be slower.

protesilaos:
Would we use it for non-denote buffers?  Maybe for the directory that holds the notes?  I can see the potential.  I just can't think of an application right now.

To be clear, I am not arguing.  What you have is fine as-is.  I am just thinking out loud.

protesilaos:
Apart from that point, is everything else okay for me to merge or are you still working on something?

jeanphilippegg:
> Would we use it for non-denote buffers? Maybe for the directory that holds the notes? I can see the potential. I just can't think of an application right now.

I don't have a use case either. I am fine with either. I will add it.



> Apart from that point, is everything else okay for me to merge or are you still working on something?

I just noticed a small bug. I will push one last commit in a few minutes.

jeanphilippegg:
All done!

protesilaos:
Merged.  Thank you!!  I have not tested everything yet, but will do it over the course of the day.

jeanphilippegg:
> Merged. Thank you!! I have not tested everything yet, but will do it over the course of the day.

You're welcome!

protesilaos:
Hello @jeanphilippegg!  I just noticed there is a regression likely caused by `a80ffc1` which affects the `denote-type` command.  Basically, it does not create the new note with the desired file type extension.  Will work on it now.

PR #64, opened 2022-07-25
Regex to allow for non-ASCII characters in title and tags

pprevos:
Updated regex to allow fontification for non-ASCII characters


protesilaos:
Merged.  Thank you!

PR #65, opened 2022-07-29
Cleanup of denote.el

jeanphilippegg:
Just a small cleanup of denote.el. There is no new code. You can look at the individual commits.

In summary:
- Put the convenience functions in their own sections.
- Rename the "denote command" to "denote command and its prompts" and move the helper functions out of it.


protesilaos:
Merged.  Thank you!

PR #66, opened 2022-07-29
Move functions from denote-dired.el and create some utility functions

jeanphilippegg:
I have started some work following the discussion on the mailing list.

---

I moved some functions from `denote-dired.el` to `denote.el` for these reasons:

- Some are not related enough to Dired or specific to Dired.
- As a work-in-progress, it is not clear what we will end up with. We can split into more appropriate files later.

I have created a section for "Note modifications" in `denote.el` for these moved functions.

I have created 2 utility functions that will probably be useful soon:

- `denote--extract-keywords-from-front-matter`: The reverse of `denote--file-meta-keywords`.
- `denote--rename-file` that just renames the file without any prompts. Later, we can use it, for example, in functions like `denote-rename-file-using-front-matter`, `denote-rename-file-using-prompts` or `denote-rename-marked-files`.

So far, I have not made any user-visible changes. Let me know what you think of this direction.

---

I learned a few things about elisp when doing the previous modifications. I have noticed that I needed to call functions/variables in `denote-retrieve.el` from `denote.el`. However, in elisp, we cannot have cyclic dependencies. Since `denote-retrieve.el` needs to call `denote.el` and it would be natural to want to call `denote-retrieve.el` from `denote.el` as is the case in this refactor, I think it would be preferable to have it all directly in `denote.el` in its own section. It is also a short file of internal functions only. For now, I have just moved what I needed in `denote.el`, but it needs further refactor. What do you think of this?

### DIFF
--- a/denote-dired.el
+++ b/denote-dired.el
@@ -200,26 +200,6 @@ old name followed by the new one.  This applies to the command
 
 ;;;; File helper functions
 
-(defun denote-dired--file-attributes-time (file)
-  "Return `file-attribute-modification-time' of FILE as identifier."
-  (format-time-string
-   denote--id-format
-   (file-attribute-modification-time (file-attributes file))))
-
-(defun denote-dired--file-name-id (file)
-  "Return FILE identifier, else generate one."
-  (cond
-   ((string-match denote--id-regexp file)
-    (substring file (match-beginning 0) (match-end 0)))
-   ((denote-dired--file-attributes-time file))
-   (t (format-time-string denote--id-format))))
-
-(defun denote-dired--rename-buffer (old-name new-name)
-  "Rename OLD-NAME buffer to NEW-NAME, when appropriate."
-  (when-let ((buffer (find-buffer-visiting old-name)))
-    (with-current-buffer buffer
-      (set-visited-file-name new-name nil t))))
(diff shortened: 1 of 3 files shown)


protesilaos:
Merged.  Thank you!

> As a work-in-progress, it is not clear what we will end up with. We can split into more appropriate files later.

Looking at this now, I feel we should merge the entirety of denote-retrieve.el into denote.el.  Putting it under separate outline headings is easier for us, as we don't have to think about which helper file has what.

Unless you have any objections, I will prepare it now.

> So far, I have not made any user-visible changes. Let me know what you think of this direction.

I am fine with this and thank you for doing it.  If we find something that needs to be tweaked once we introduce the new user-facing code, we will adapt things accordingly.

> [...] However, in elisp, we cannot have cyclic dependencies. [...] What do you think of this?

As I wrote above, I think denote.el can hold everything that is now in denote-retrieve.el.

protesilaos:
I pushed the change as it felt innocuous.  Also renamed all `denote-retrieve--` to `denote--retrieve-`.  Please let me know if this is okay.

jeanphilippegg:
Yes. This is what I had in mind. Thank you!

PR #67, opened 2022-07-30
Rename denote-dired--rename-file to denote-dired--rename-file-prompt

jeanphilippegg:
The function `denote-dired-rename-file-prompts` now only prompts to rename the file. The renaming is done by the function denote--rename-file (which does not prompt).


protesilaos:
Hello @jeanphilippegg!  I notice you force-pushed some more commits.  Please let me know when you feel this is ready.

jeanphilippegg:
Yes. I deleted the last 2 commits, but I pushed them again if you are interested. I moved the renaming functions from `denote-dired.el` to `denote.el`, in their own section. The remainder of `denote-dired.el` is about `denote-dired-mode` and fontification.

I have not modified the Commentary sections. You can choose to merge the last two commits, if you like it.

protesilaos:
I am merging everything.  Thank you!  We should check whether we need the `(require 'dired)` now.  Maybe we can get away without it, mutatis mutandis.

About the change to the return value of `denote--rename-file`, is there a reason for `t` specifically?

protesilaos:
Since we are merging code into denote.el, I will do the same with the remainder of denote-dired.el and then with denote-faces.el.  Makes things easier this way.

protesilaos:
And for denote-link.el too: it already has glue code with Dired, so we might as well not break things into different files.  denote-org-capture.el can remain separate for now.

jeanphilippegg:
> About the change to the return value of `denote--rename-file`, is there a reason for `t` specifically?

No, I just wanted a non-nil value to be returned if the function is executed. Previously, nil was returned if the file was not visited. But we can remove this because we don't use the return value. It comes from previous code.

protesilaos:
I don't mind it  I was just curious.  If you want to change it, I am okay with it.  Though please wait a few minutes as I am now finalising some unrelated changes.

jeanphilippegg:
> I don't mind it I was just curious. If you want to change it, I am okay with it. Though please wait a few minutes as I am now finalising some unrelated changes.

Ok!

I was okay with multiple files, but everything in `denote.el` is also good with me!

Also, maybe we can remove the option `denote-dired-rename-export`? It is not used anywhere. Maybe we can reintroduce it when we are ready to use it.

protesilaos:
> Also, maybe we can remove the option denote-dired-rename-export? It is not used anywhere. Maybe we can reintroduce it when we are ready to use it.

Yes!  Though we will be refactoring the relevant parts anyway.  I suggest that part of our change introduces obselete aliases for `denote-dired-rename-file` and related.  The "dired" part is not needed: before it was necessary because it was in the denote-dired.el.

protesilaos:
Okay, pushed the changed.  I removed denote-org-capture.el as well.  There was nothing specific to org-capture there, so no new dependency is added.

PR #69, opened 2022-08-01
Remove denote-dired-rename-file-and-add-front-matter

jeanphilippegg:
First step toward a more streamlined rename function.

In `denote--edit-front-matter-p`, I check that there is a front matter to edit. The rule is just that the note must have a "title" AND a "tags" line. This is the precondition for `denote--rewrite-front-matter` to be successful. If one of them is missing, we just add a front matter unconditionally and let the user do the manual adjustments. We can further refine this later.

Now that `denote-rename-file` modifies OR ADD a front matter, I made both `denote-dired-rename-file` and `denote-dired-rename-file-and-add-front-matter` obsolete.

I have to test this, but I wanted to share it.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1343,53 +1343,20 @@ The operation does the following:
 - a prompt is asked once for the KEYWORDS field and the input is
   applied to all file names;
 
-- if the file is recognized as a Denote note, rewrite its front
-  matter to include the new keywords.  A confirmation to carry
-  out this step is performed once at the outset.  Note that the
-  affected buffers are not saved.  The user can thus check them
-  to confirm that the new front matter does not cause any
+- if the file is recognized as a Denote note, add a front matter
+  or rewrite it to include the new keywords. A confirmation to
+  carry out this step is performed once at the outset. Note that
+  the affected buffers are not saved. The user can thus check
+  them to confirm that the new front matter does not cause any
   problems (e.g. with the command `diff-buffer-with-file').
   Multiple buffers can be saved with `save-some-buffers' (read
-  its doc string)."
-  (interactive nil dired-mode)
-  (if-let ((marks (dired-get-marked-files))
-           (keywords (denote--keywords-prompt)))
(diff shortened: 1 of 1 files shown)


protesilaos:
> I have to test this, but I wanted to share it.

Thank you!  I will only be able to test this later in the day.

> The rule is just that the note must have a "title" AND a "tags" line. 

This makes sense.  If the need arises, we can add more conditions later.

> If one of them is missing, we just add a front matter unconditionally and let the user do the manual adjustments. We can further refine this later.

All good.

> Now that `denote-rename-file` modifies OR ADD a front matter, I made both `denote-dired-rename-file` and `denote-dired-rename-file-and-add-front-matter` obsolete.

Yes, I think it is better to consolidate what we offer.  Options should be there when necessary, otherwise we should just "do the right thing".

jeanphilippegg:
I have tested `denote-rename-file`.

I also pushed another commit to also delete `denote-dired-rename-marked-files-and-add-front-matter` and only keep `denote-dired-rename-marked-files`. Should we remove `dired` from the name of this command?

Now that we only have 2 renaming commands left, is there something else that we want to do?

protesilaos:
> Should we remove `dired` from the name of this command? Maybe rename it to `denote-rename-files`?

I don't mind either way.  Maybe keeping it will be easier for users who want to bind the commands specifically to the `dired-mode-map`?  Though removing it makes things more consistent on our end...  Whatever you prefer.

jeanphilippegg:
Ok. We can keep it for now!

protesilaos:
Merged.  Thank you!

jeanphilippegg:
You're welcome! Let me know if there is something else we need to do about the renaming functions.

protesilaos:
Hello @jeanphilippegg!  Just to note that I am now reviewing all the renaming-related code we have.  Pushed commit `6b2067d` which removes `denote-dired-rename-expert`.  I will check all the functionality and update the manual accordingly.  Will keep you posted, if needed.

I wanted to do this earlier, but got very busy with other projects.

PR #72, opened 2022-08-02
Replace underscore with hyphen in denote--slug-hyphenate

jeanphilippegg:
When typing a title "my_title", I think it would be more natural to get "my-title" after sluggification than "mytitle".

This pull request makes this change. It also works for the keywords as well.

I fixed the docstring of `denote-allow-multi-word-keywords`. Previously, it stated that "word1+word2" would get converted to "word1-word2", but it is instead converted to "word1word2".


protesilaos:
Good catch!  Merged.  Thank you!

PR #73, opened 2022-08-03
Cleanup

jeanphilippegg:
Just some cleanup in the code:
- Use `seq-filter` when appropriate.
- Use `file-name-base`.
- Simplify `denote--format-file`.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -454,15 +454,12 @@ names that are relative to the variable `denote-directory'."
 
 (defun denote--directory-files-matching-regexp (regexp)
   "Return list of files matching REGEXP."
-  (delq
-   nil
-   (mapcar
-    (lambda (f)
-      (when (and (denote--only-note-p f)
-                 (string-match-p regexp f)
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you!  About the return value of `denote--barf-duplicate-id`, I think there was a reason to use `if` there.  Though I may not remember correctly.  Does the `when` work on your end?

jeanphilippegg:
`denote--barf-duplicated-id` is only used in the `denote` command, but we don't check the return value.

protesilaos:
Okay.  Merged.  Thank you!

protesilaos:
I will start testing everything later today, update the manual, etc.

jeanphilippegg:
Just for your information, `denote--barf-duplicated-id` was used in a `when-let` before we refactored the `denote` command, so the return value was used. I don't think that is necessary anymore.

protesilaos:
Ah yes, that makes sense.

PR #75, opened 2022-08-05
Add denote-rename-file-using-front-matter and denote-rename-marked-fi…

jeanphilippegg:
I have not checked your command. I will check it now.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1169,11 +1169,6 @@ operation on multiple files."
 (make-obsolete 'denote-dired-rename-expert nil "0.5.0")
 (make-obsolete 'denote-dired-post-rename-functions nil "0.4.0")
 
-(defun denote--front-matter-keywords-to-list (keywords)
-  "Return string of KEYWORDS as a list of strings.
-This is the inverse of `denote--format-front-matter-keywords'."
-  (split-string keywords "[:,\s]+" t "[][ \"']+"))
-
 ;;;;; The renaming commands and their prompts
 
 (defun denote--rename-dired-file-or-prompt ()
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
It is exactly the same code! Yours is better, because it uses your latest renaming of functions. Don't merge this pull request.

One comment: If I had not read the mailing lists, it would not be clear to me what `reverse` means in `denote-reverse-rename-file`. I think `denote-rename-file-using-front-matter` is clearer.

protesilaos:
> It is exactly the same code! Yours is better, because it uses your latest renaming of functions. Don't merge this pull request.

Okay.  There is still scope for `denote-rename-marked-files-using-front-matter`, but I guess you want to tweak it?

> I think `denote-rename-file-using-front-matter` is clearer.

Agreed!  I used this intentional for the work-in-progress.  Will you rename it as part of this PR or shall I do it now?

jeanphilippegg:
Also, to retrieve the keywords, maybe use `denote--extract-keywords-from-front-matter` instead of `denote--front-matter-keywords`. The former seems more robust, because it does the right job for each file types.

I will force-push these modifications so you can see the final result.

protesilaos:
Yes, I just noticed the `denote--extract-keywords-from-front-matter`.  I had forgotten about it.  Sorry!

jeanphilippegg:
> Yes, I just noticed the `denote--extract-keywords-from-front-matter`. I had forgotten about it. Sorry!

No worry!

I just pushed the changes. Note that the identifier in the front matter is irrelevant and not checked, so I remove it from the user error.

Also, `denote-rename-marked-files-using-front-matter` does not check if the buffers are modified.

protesilaos:
Merged.  Thank you!

jeanphilippegg:
I would also remove this from the docstring: "This basically is the inverse of `denote-rename-file'". The link between it and `denote-rename-file-using-front-matter` is not so clear.

However, I would add a comment: "The identifier of the file (if any) is never modified."

protesilaos:
Okay, please do it.

jeanphilippegg:
We posted at the same time. I added some suggestions.

PR #77, opened 2022-08-07
Insert template in denote--prepare-note

jeanphilippegg:


protesilaos:
Yes, this is better!  Merged.  Thank you!

By the way, what do you think about this feature?  It has been requested before.  I feel it will be useful.

jeanphilippegg:
I think I like it. It fits among the other options. I probably will not use it for myself though.

protesilaos:
I might use it for some journaling, though I also tend to have free-form notes.  But I will be testing it periodically for maintenance.

PR #78, opened 2022-08-07
Remove most denote-last-* variables

jeanphilippegg:
Unless the removed variables are useful outside of `denote.el`, I think they can be removed. Now, only `denote--prepate-note` and `denote-org-capture` set `denote-last-path`. I am not sure if `denote--prepare-note` really needs to set it though. I think it can be removed as well. What do you think?

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -371,12 +371,6 @@ This variable is for advanced users who need to extend the
 understanding of what we should be omitting, we will update
 things accordingly.")
 
-(defvar denote-last-path nil "Store last path.")
-(defvar denote-last-title nil "Store last title.")
-(defvar denote-last-keywords nil "Store last keywords.")
-(defvar denote-last-buffer nil "Store last buffer.")
-(defvar denote-last-front-matter nil "Store last front-matter.")
-
 ;;;; File helper functions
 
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
Since they are not internal variables, they were probably intended to be available for something. You can close this pull request without merging in that case!

protesilaos:
I don't really remember the rationale for all of them.  I think it was pre version `0.1.0` and I was basically trying to test that everything works correctly (by inspecting the value of the variable).

The `denote-last-path` is used for the `org-capture` template due to technicalities with how that evalutes its templates.  Maybe we can do something about it?

To the point of this PR, I am fine with the operation.  I believe we don't mention those other variables in the manual.  Though it better to have a `make-obsolete` just to be sure.

jeanphilippegg:
`denote-last-path` is used in a template for org capture in the manual. Maybe the other variables were intended to be used similarly, but I cannot think of anything.

I removed `(setq denote-last-path ...)` from `denote--prepare-note`. It is probably unncessary in that function like the `setq` of the other variables. Now only the org-capture functions use `denote-last-path`.

I made the other variables obsolete.

protesilaos:
Merged.  Thank you!

PR #80, opened 2022-08-09
Cleanup

jeanphilippegg:
Many cleanups:

- I have removed the let-bindings of `default-directory`. I had to be careful because let-binding `default-directory` can have non obvious consequences. Please double-check the code and test the affected functions: `denote`, `denote-link-backlinks`, `denote--edit-front-matter-p` (the renaming functions) and `denote-link-add-links`. My own tests did not yield any errors.

- I deleted the unused variable `denote-link--links-to-files`.

- I removed the check for the current-buffer in `denote--id-exists-p`. We already discussed this in another issue/pull request. We did not know what it was used for, but we kept it. `denote--id-exists-p` is only used by the `denote` command now. I don't know if it was used elsewhere previously, but I cannot imagine why this check would be necessary now. I say we remove it and if there is any issue, we fix it and document it. By the way, `denote--id-exists-p` has been broken for weeks (I fixed it. See below.).

- The following functions were simplified:
  - `denote--directory-files`: Remove the `absolute` argument from `denote--directory-files` because we (almost) always set it anyway.
  - `denote--directory-files-recursively`: Merged it in `denote--directory-files` because it is now only a single line.
  - `denote--buffer-file-names`: Now returns absolute paths as well.
  - `denote--retrieve-files-in-xrefs` and `denote--retrieve-proces-grep`: Use absolute paths.
  
  The idea is to keep these functions simple and make them always return absolute paths. We can use `denote--file-name-relative-to-denote-directory` on the items of the list, when necessary, but it does not happen often (only once for each function).

- In `denote-subdirs`, I removed the TODO items. I made it to ignore hidden directories. This way, a lot of VC backends are handled and we provide a way for users to create directories ignored by Denote in `denote-subdirectory`. What do you think of this?

Bugs that were noticed and fixed as part of these changes:

- Calling `(denote--id-regexp "some-existing-id")` at the root of `denote-directory` triggers the debugger. This has probably gone unnoticed because it is only used to check collisions and it never happens.

- In a note at the root of `denote-directory`, calling `denote-link-add-links` with a regex matching the current file does not add the current file to the links. Doing the same operation in a file inside a subdirectory adds the current file to the links.

- A directory called `some.notes.about.git` is ignored by Denote because it contains `.git`.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1900,6 +1900,7 @@ Use optional TITLE for a prettier heading."
   (let ((inhibit-read-only t)
         (buf (format "*denote-backlinks to %s*" id)))
     (with-current-buffer (get-buffer-create buf)
+      (setq-local default-directory (denote-directory))
       (erase-buffer)
       (special-mode)
       (goto-char (point-min))
--- a/denote.el
+++ b/denote.el
@@ -617,18 +617,14 @@ The xrefs are returned as an alist."
   "Return sorted file names sans directory from XREFS.
 Parse `denote--retrieve-xrefs'."
   (sort
-   (delete-dups
-    (mapcar (lambda (x)
-              (denote--file-name-relative-to-denote-directory (car x)))
-            xrefs))
+   (delete-dups (mapcar #'car xrefs))
    #'string-lessp))
 
 (defun denote--retrieve-proces-grep (identifier)
   "Process lines matching IDENTIFIER and return list of files."
-  (let ((file (denote--file-name-relative-to-denote-directory (buffer-file-name))))
-    (denote--retrieve-files-in-output
-     (delete file (denote--retrieve-files-in-xrefs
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged.  Thank you!

protesilaos:
About the check for hidden directories: I think it should be okay this way, though it can still theoretically return some false positives.  If there ever is a problem with it, we will need to exclude them specifically.

protesilaos:
I think it is an elegant solution to let the user omit subdirectories by prepending the dot.  Though I wonder if it could cause confusion with those users who alwyas have dotfiles hidden.  A user option is more fool-proof in that regard, though I do not consider it important to add right now.  Let there be some demand for it.  That way we also understand better how people use those.

jeanphilippegg:
Yes, I agree! We can still add a user option later.

PR #81, opened 2022-08-09
Fix denote--id-exists-p

jeanphilippegg:
This was wrong. `denote` was failing.


protesilaos:
Merged.  Thank you!

protesilaos:
By the way, do you think we should do anything about the potential of duplicate IDs if created within a fraction of a second?  We discussed this before, but maybe you have changed your mind.  I acknowledge the problem, though I have yet to see a practical case where it causes errors.  So I would rather not fix it pre-emptively.

jeanphilippegg:
No, I don't think it is necessary. For a month, a collision would have triggered the debugger, but nobody reported it. I think `denote--barf-duplicate-id` is enough.

PR #82, opened 2022-08-10
Minor cleanups

jeanphilippegg:
Small cleanups. Check the individual commits.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -807,10 +807,10 @@ and TEMPLATE should be valid for note creation."
 
 (defun denote--dir-in-denote-directory-p (directory)
   "Return DIRECTORY if in variable `denote-directory', else nil."
-  (when-let* ((dir directory)
-              ((string-prefix-p (expand-file-name (denote-directory))
-                                (expand-file-name dir))))
-    dir))
+  (when (and directory
+             (string-prefix-p (denote-directory)
+                              (expand-file-name directory)))
+    directory))
 
 (defun denote--file-type-symbol (filetype)
   "Return FILETYPE as a symbol."
--- a/denote.el
+++ b/denote.el
@@ -169,10 +169,6 @@ the appropriate list of strings."
   :package-version '(denote . "0.1.0")
   :type 'boolean)
 
-(defconst denote--prompt-symbols
-  '(title keywords date file-type subdirectory)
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
Wait a few minutes before merging. There is a small issue.

protesilaos:
Thank you!

I have not tested this: could `denote--format-front-matter-keywords` and `denote--format-front-matter` still need the fallback conditions?

For the new `denote-link-button-action`, we need to make the old an obsolete alias.  Someone may be using it, so better warn them about the change.

jeanphilippegg:
> I have not tested this: could `denote--format-front-matter-keywords` and `denote--format-front-matter` still need the fallback conditions?

I checked. In the one place we do not set it, we let-bind `denote-file-type` to set it indirectly. It is better to set it directly.

> For the new `denote-link-button-action`, we need to make the old an obsolete alias. Someone may be using it, so better warn them about the change.

Yes. I already did this!

I am still working on one more cleanup, if you want to wait a few minutes.

protesilaos:
> I checked. In the one place we do not set it, we let-bind `denote-file-type` to set it indirectly. It is better to set it directly.

Very well!

> Yes. I already did this!

Oh, sorry!  I missed it.

> I am still working on one more cleanup, if you want to wait a few minutes.

Please take as much time as you need.  We are not in a hurry.

jeanphilippegg:
Done!

protesilaos:
Merged.  Thank you!

protesilaos:
Do you think you will make any other changes within the next hours or days?  Just asking because I believe this is the right time to release version `0.5.0`.

The only task that remains is to write a helper command to migrate Denote+Org users who had `#+filetags:  one  two` and need to convert to `#+filetags:  :one:two:`.  This probably needs to stay in the manual, as it is not pertinent to denote.el.

jeanphilippegg:
> Do you think you will make any other changes within the next hours or days? Just asking because I believe this is the right time to release version `0.5.0`.

No, I don't think I have anything else for the short term!

However, I just noticed that `denote--date` uses `denote-file-type`, but I have removed the let-bindings of `denote-file-type`. So I need to fix thix before 0.5.0 is done.

jeanphilippegg:
Done. See pull request #83.

PR #83, opened 2022-08-10
Refactor denote--date

jeanphilippegg:
I don't intend to make other changes before 0.5.0.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -752,34 +752,34 @@ provided by `denote'. FILETYPE is one of the values of
 
 ;; Adapted from `org-hugo--org-date-time-to-rfc3339' in the `ox-hugo'
 ;; package: <https://github.com/kaushalmodi/ox-hugo>.
-(defun denote--date-rfc3339 (&optional date)
+(defun denote--date-rfc3339 (date)
   "Format date using the RFC3339 specification.
-With optional DATE, use it else use the current one."
+If DATE is nil, use the current date."
   (replace-regexp-in-string
    "\\([0-9]\\{2\\}\\)\\([0-9]\\{2\\}\\)\\'" "\\1:\\2"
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged.  Thank you!

Do you have any preference for---or thoughts on---this?  (I believe you are using Org)

> The only task that remains is to write a helper command to migrate Denote+Org users who had `#+filetags:  one  two` and need to convert to `#+filetags:  :one:two:`.  This probably needs to stay in the manual, as it is not pertinent to denote.el.

jeanphilippegg:
> > The only task that remains is to write a helper command to migrate Denote+Org users who had `#+filetags:  one  two` and need to convert to `#+filetags:  :one:two:`.  This probably needs to stay in the manual, as it is not pertinent to denote.el.

If there is nothing broken when manipulating org files, I think we can mention the migration steps in the manual only. Only the functions `denote-rename-file-using-front-matter` and `denote-rename-marked-files-using-front-matter` will not work with old notes.

protesilaos:
> If there is nothing broken when manipulating org files, I think we can mention the migration steps in the manual only. 

Yes, I think the same.

>  Only the functions `denote-rename-file-using-front-matter` and `denote-rename-marked-files-using-front-matter` will not work with old notes.

This, however, suggests we need the command in denote.el so that we "officially" support the migration.  No worries though.  I will work on it today and see what fits better.  Even if it is in denote.el, it will be temporary and shall be deprecated afterwards.

jeanphilippegg:
> This, however, suggests we need the command in denote.el so that we "officially" support the migration. No worries though. I will work on it today and see what fits better. Even if it is in denote.el, it will be temporary and shall be deprecated afterwards.

That's good!

PR #84, opened 2022-08-10
denote--format-front-matter: Handle nil filetype gracefuly

juergenhoetzel:
As descriped in denote-file-type docstring:

By default (a nil value), the file type is that of Org mode.
Any other non-nil value is the same as the default."

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -764,8 +764,8 @@ provided by `denote'.  FILETYPE is one of the values of
       ('text (format denote-text-front-matter title date
                      (denote--format-front-matter-keywords keywords 'text)
                      id denote-text-front-matter-delimiter))
-      ('org (format denote-org-front-matter title date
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you @juergenhoetzel!  I merged it.

We were making some changes earlier and I missed this.  I am pinging @jeanphilippegg who may have another idea on how to handle this.

protesilaos:
@jeanphilippegg The issue is with this in `denote`:

```elisp
(file-type (denote--file-type-symbol (or file-type denote-file-type)))
```

It returns nil.

jeanphilippegg:
Internally, I think we should arrange to determine a valid filetype early when `denote` is called. This way, our internal functions do not need to guess or have fallbacks. We can always arrange for them to be called with valid values. I will make a pull request to show what I mean.

protesilaos:
Fine by me.  Please do it as you prefer.

Issue #74, opened 2022-08-05
Refining the renaming mechanism


protesilaos:
I am posting this here since it is where @jeanphilippegg is active with changes to the code.

cc @hpgisler

1. @jeanphilippegg consolidated the renaming commands to just `denote-rename-file` and `denote-dired-rename-marked-files`.  I just tested everything and updated the relevant docs.  This addresses the issue we had before with ergonomics, namely, that we had distinct commands for rewriting and for adding front matter in addition to the renaming operation.

2. We now are ready for the function that completes the front matter, as we discussed on the mailing list.  In short: add missing entries to the front matter block.

3. The "reverse rename" operation, which would read the front matter and rename the file accordingly.

I feel that 2 is the trickiest one because we do not define such entries one-by-one but only as a block (e.g. `denote-org-front-matter`).  I would thus leave this for last.

Unless @jeanphilippegg already has some code in the works, I will now start working on 3.

* * *

Mailing list thread for the sake of completeness (I wish they were not split like that, but here we are):

- https://lists.sr.ht/~protesilaos/denote/%3C87k081l6vw.fsf%40silverstone.mail-host-address-is-not-set%3E
- https://lists.sr.ht/~protesilaos/denote/%3C878rogw5kk.fsf%40protesilaos.com%3E
- https://lists.sr.ht/~protesilaos/denote/%3C87fsiljv1s.fsf%40hu.mail-host-address-is-not-set%3E
- https://lists.sr.ht/~protesilaos/denote/%3C87r122afe3.fsf%40hu.mail-host-address-is-not-set%3E

jeanphilippegg:
> Unless @jeanphilippegg already has some code in the works, I will now start working on 3.

Actually, I do! I had made a function `denote-rename-file-using-front-matter` as well as `denote-rename-marked-files-using-front-matter`. I had kept them for myself until things would be more stable and to see how we wanted to do it exactly.

Do you want me to share them as is? Should they be their own functions?

protesilaos:
Please share, though note I just pushed the ones I had.  We should consolidate everything.

protesilaos:
The code for the "reverse rename" is completed with [PR #75](https://github.com/protesilaos/denote/pull/75).  I will include them in the manual a bit later.

jeanphilippegg:
>     2. We now are ready for the function that completes the front matter, as we discussed on the mailing list.  In short: add missing entries to the front matter block.

[...]

PR #85, opened 2022-08-10
Cleanup

jeanphilippegg:
I have made the change discussed in #84.

It includes these changes:
- In `denote` command, make sure `denote--prepare-notes` is called with valid values.
- Rename the function `denote--file-type-symbol` to `denote--valid-file-type` to make it return a valid file-type.
- Simplify `denote--format-file`. We always supply a non-nil extension parameter, so I removed the fallback on `denote--file-extension denote-file-type`.

The general theme of the cleanups of the previous days is to have simpler internal functions that minimize the use of fallback conditions, optional parameters and parameter validations. I think it makes it simpler for us. We can always expect the parameters of internal functions to be in a valid state.

The changes above do not require a new version.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -654,7 +654,7 @@ Parse `denote--retrieve-xrefs'."
     ('markdown-toml ".md")
     ('markdown-yaml ".md")
     ('text ".txt")
-    (_ ".org")))
+    ('org ".org")))
 
 (defun denote--format-file (path id keywords title-slug extension)
   "Format file name.
--- a/denote.el
+++ b/denote.el
@@ -764,7 +764,7 @@ provided by `denote'.  FILETYPE is one of the values of
       ('text (format denote-text-front-matter title date
                      (denote--format-front-matter-keywords keywords 'text)
                      id denote-text-front-matter-delimiter))
-      (_ (format denote-org-front-matter title date
+      ('org (format denote-org-front-matter title date
                  (denote--format-front-matter-keywords keywords 'org) id)))))
 
 (defun denote--path (title keywords dir id file-type)
@@ -827,14 +827,15 @@ and TEMPLATE should be valid for note creation."
                               (expand-file-name directory)))
     directory))
 
-(defun denote--file-type-symbol (filetype)
-  "Return FILETYPE as a symbol."
-  (cond
-   ((stringp filetype)
-    (intern filetype))
-   ((symbolp filetype)
-    filetype)
-   (t (user-error "`%s' is not a symbol or string" filetype))))
+(defun denote--valid-file-type (filetype)
+  "Return a valid filetype given the argument FILETYPE."
+  (unless (or (symbolp filetype) (stringp filetype))
+    (user-error "`%s' is not a symbol or string" filetype))
+  (when (stringp filetype)
+    (setq filetype (intern filetype)))
+  (if (memq filetype '(text org markdown-toml markdown-yaml))
+      filetype
+    'org))
 
 (defun denote--date-add-current-time (date)
   "Add current time to DATE, if necessary.
@@ -934,7 +935,7 @@ When called from Lisp, all arguments are optional.
          ('template (aset args 5 (denote--template-prompt)))))
      (append args nil)))
   (let* ((title (or title ""))
-         (file-type (denote--file-type-symbol (or file-type denote-file-type)))
+         (file-type (denote--valid-file-type (or file-type denote-file-type)))
          (date (if (or (null date) (string-empty-p date))
                    (current-time)
                  (denote--valid-date date)))
--- a/denote.el
+++ b/denote.el
@@ -818,7 +818,7 @@ and TEMPLATE should be valid for note creation."
                   file-type)))
     (with-current-buffer buffer
       (insert header)
-      (when template (insert template)))))
+      (insert template))))
 
 (defun denote--dir-in-denote-directory-p (directory)
   "Return DIRECTORY if in variable `denote-directory', else nil."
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged.  Thank you!

Do you think this needs to be in a new `0.5.2` version?  Or shall we leave it for the next release?

jeanphilippegg:
I don't think there is anything user-facing. A new version would not be necessary.

protesilaos:
Okay.  Thank you!  Depending on how development goes, the next version may be `1.0.0`, though for now it is tentatively `0.6.0`.

jeanphilippegg:
Just to keep you informed of what I have in mind for a future version. Here are the points I would address:

- Determine what characters are acceptable in keywords (in the front matter), as discussed in #74. For example, do we want to allow keywords to be "ke!y1  Other_Key"? I don't have an answer. But we could fix it and document it.

- I would like to rework the `denote--retrieve-*` and related functions that retrieve data from the file content. Right now, we search the file using a regexp, sometimes multiple times. Also, the syntax of the different file types are handle by the same regexp. As discussed in #74, I think I prefer a more precise search, given a file type.

- #86 got me thinking and I think that, combined with the previous item, we can reach a more general way to handle file types. I will share this code instead of explaining:

```elisp
(setq denote-file-types
  '((org ".org" "org-front-matter-string")
    (text ".txt" "text-front-matter-string")
    (markdown-toml ".md" "mdtoml-front-matter-string")
    (markdown-yaml ".md" "mdyaml-front-matter-string")
    (markdown-qml ".qmd" "qmd-front-matter-string")
    (markdown-rml ".rmd" "rmd-front-matter-string")))
```

There are challenges, but if we are able to create general functions to handle the file types, this would even remove complexity in the current code (no more `cond` on file types). The user could add/remove any file-type.

protesilaos:
> - Determine what characters are acceptable in keywords (in the front matter), as discussed in #74. For example, do we want to allow keywords to be "ke!y1 Other_Key"? I don't have an answer. But we could fix it and document it.

Indeed.  We need to have a discussion about it.  Apparently Org does not allow hyphens in the `#+filetags` (?), so that is the sort of datum we would need to account for.

> - I would like to rework the `denote--retrieve-*` and related functions that retrieve data from the file content.  Right now, we search the file using a regexp, sometimes multiple times. Also, the syntax of the different file types are handle by the same regexp. As discussed in #74, I think I prefer a more precise search, given a file type.

Yes, I am all for it.  This should be simplified.  There is also a `FIXME` about tweaking the regular expressions or, generally, reducing the steps we need to take.

I am also in favour of the more precise search.  My intent was to sort out some likely faulty input.  I think that idea should be preserved while making the search more precise.

> - #86 got me thinking and I think that, combined with the previous item, we can reach a more general way to handle file types. I will share this code instead of explaining:
> 
> ```elisp
> (setq denote-file-types
>   '((org ".org" "org-front-matter-string")
>     (text ".txt" "text-front-matter-string")
>     (markdown-toml ".md" "mdtoml-front-matter-string")
>     (markdown-yaml ".md" "mdyaml-front-matter-string")
>     (markdown-qml ".qmd" "qmd-front-matter-string")
>     (markdown-rml ".rmd" "rmd-front-matter-string")))
> ```

Yes, we need to improve this situation.  What we have now does not scale.  There is no reason why it shouldn't.  I like the alist approach.  Internally, we can handle the existing symbols to translate them to what we mean.

> There are challenges, but if we are able to create general functions to handle the file types, this would even remove complexity in the current code (no more `cond` on file types). The user could add/remove any file-type.

Agreed!

How do you want to proceed?  Should we make a division of labour and see what we can come up with?  Or do you prefer to prepare the code yourself?  Whatever works for you.

protesilaos:
Just to note though that such changes will either have to go to a separate branch or wait a few more days.  I give it some time to determine if there any bug reports whose fixes need a point release.

jeanphilippegg:
> Just to note though that such changes will either have to go to a separate branch or wait a few more days. I give it some time to determine if there any bug reports whose fixes need a point release.

I agree! There have been a lot of changes lately. And I think I would need some time to ponder on these new ideas.

> I am also in favour of the more precise search. My intent was to sort out some likely faulty input. I think that idea should be preserved while making the search more precise.

Yes. I think a function like `denote--front-matter-keywords-to-list` should retain some tolerance to faulty inputs.

> How do you want to proceed? Should we make a division of labour and see what we can come up with? Or do you prefer to prepare the code yourself? Whatever works for you.

There are many challenges towards that end goal. We could just work on one at a time and merge them into `main` until hopefully we reach what we have discussed here.

I would take a few days to let these ideas mature. We could start listing what we will need though.

---

I will start with one reflexion. Given the `denote-file-types` variable above, do we have enough information? We will need to have a function to retrieve the `title` line from the file content, for example. We could analyze the single string representing the front matter, but I am not sure this would be very robust. Would there be a be more precise way to specify it?

I don't expect you to answer these questions. I am just sharing my thoughts. We can take enough time to think about them carefully.

protesilaos:
> I would take a few days to let these ideas mature. We could start listing what we will need though.

Very well!  We can do it on a new issue.  It makes it easier to link to it in the change log.

> I will start with one reflexion. Given the `denote-file-types` variable above, do we have enough information? [...] Would there be a be more precise way to specify it?

I think we cannot be 100% confident.  The `denote--filetype-heuristics` is the product of that.  I cannot think of a bulletproof constant we could rely on.  If we add some arbitrary entry, we adding noise that the user would not want.

> I don't expect you to answer these questions. I am just sharing my thoughts. We can take enough time to think about them carefully.

Sure, we need to consider everything carefully.  Though there is no need to worry about it.  We can make some changes as part of version `0.6.0` and once we are happy with what we have, we can then work towards `1.0.0`.  There are no deadlines or a timeline.

jeanphilippegg:
I believe you have unmerged this.

protesilaos:
Oh, I messed up.  Let me see if I can retrieve it.

jeanphilippegg:
I have the commit on my side if you cannot retrieve it.

protesilaos:
I think I got it right.  Can you please check?

jeanphilippegg:
You have cherry-picked the individual commits. That's good!

protesilaos:
Very well!

PR #88, opened 2022-08-11
Remove some unused variables and functions

jeanphilippegg:
This is a first step towards what we discussed in #85. I will try to remove the remaining key-regexp variables. I think I have an idea, but I will not be able to work on it tomorrow. This will let time to see if the current version contains bugs, but you can work on it if you want.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -557,17 +557,9 @@ output is sorted with `string-lessp'."
 
 ;;;; Front matter or content retrieval functions
 
-(defconst denote--retrieve-id-front-matter-key-regexp
-  "^.?.?\\b\\(?:identifier\\)\\s-*[:=]"
-  "Regular expression for identifier key.")
-
 (defconst denote--retrieve-title-front-matter-key-regexp
   "^\\(?:#\\+\\)?\\(?:title\\)\\s-*[:=]"
   "Regular expression for title key.")
-
-(defconst denote--retrieve-date-front-matter-key-regexp
-  "^\\(?:#\\+\\)?\\(?:date\\)\\s-*[:=]"
-  "Regular expression for date key.")
 
 (defconst denote--retrieve-keywords-front-matter-key-regexp
   "^\\(?:#\\+\\)?\\(?:tags\\|filetags\\)\\s-*[:=]"
@@ -608,14 +600,6 @@ If optional KEY is non-nil, return the key instead."
   (denote--retrieve-search
    file
    denote--retrieve-title-front-matter-key-regexp
-   key))
-
-(defun denote--retrieve-value-date (file &optional key)
-  "Return date value from FILE.
-If optional KEY is non-nil, return the key instead."
-  (denote--retrieve-search
-   file
-   denote--retrieve-date-front-matter-key-regexp
    key))
 
 (defun denote--retrieve-value-keywords (file &optional key)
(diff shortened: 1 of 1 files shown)


protesilaos:
Do you want me to merge this as-is or should I wait for the complete changeset?

protesilaos:
Okay, thanks!

jeanphilippegg:
This pull request contains the following changes:

- Remove all key-regexp variables.

- Refactor denote--filetype-heuristics to not use the retrieval functions because I want to make them to depend on the file type and that is what we want to find in denote--filetype-heuristics.

- Remove the optional key argument from denote--retrieve-search and related functions because it is not used anymore. Only denote--filetype-heuristics used it, but I refactored it in the previous item.

- Create functions `denote--title-key-regexp` and `denote--keywords-key-regexp` which depends on the file type.

- Add a file-type parameter to a few functions to use the new functions in the previous item.

This is just a cleanup to remove the key-regexps variables. Now, when looking for a `title` line in a file, we look for the right title line depending on the file type. This means that if a user has a line starting with "title:" in an Org file, we will not find it. I don't think that is the kind of tolerance we were talking about in #85. We want to tolerate faulty inputs in the keywords, not really in the key. This refactor brings user-visible robustness as well: if a user has both a line starting with "title:" and "#+title:" in an Org file, we will find the right one.

Also note that when renaming a file with an unknown extension, we now use `denote--filetype-heuristics` to guess the file type, and we will be looking for `#+title:` in the file to retrieve a potential title. We use the file name if we do not find it. I don't think this is a deal-breaker. We can refine this later.

protesilaos:
Thank you!  I didn't have time to check it today.  Will do it tomorrow morning.

jeanphilippegg:
Thanks! Take your time.

protesilaos:
> We want to tolerate faulty inputs in the keywords, not really allow mixed keys.

Yes!

> Also note that when renaming a file with an unknown extension [...] We use the file name if we do not find it. I don't think this is a deal-breaker. We can refine this later.

My concern with this is that maybe the user has some good reason to keep the file without an extension.  I cannot think of such a reason, except for a file-local variable.  Ultimately though, this change makes sense for us: if the user has a plain text file that is renamed by us, it should thenceforth be treated as a "note" per our tests.

protesilaos:
Merged.  Thank you!

jeanphilippegg:
You're welcome! I will make a pull request soon, but I have doubts about it. I will share it anyway to see what you think.

protesilaos:
Okay.

An aside here: there is a thread on the mailing list about linking to Denote-named files which are not `{txt,md,org}`.  I think we can do this.  Just need to make sure we don't try to visit the file for front matter retrieval (which I believe we already do, but I will double-check).

jeanphilippegg:
> My concern with this is that maybe the user has some good reason to keep the file without an extension. I cannot think of such a reason, except for a file-local variable. Ultimately though, this change makes sense for us: if the user has a plain text file that is renamed by us, it should thenceforth be treated as a "note" per our tests.

Just to be clear, if a user renames an extensionless file, we do not add ".org" as an extension. We still do as before and keep the original extension. If there is no front matter, we add an Org front matter (as before). However, if there is already some kind of front matter, ie a line starting with "title:", we will now ignore it. We specifically are looking for a line starting with "#+title:" (an Org title).

jeanphilippegg:
> An aside here: there is a thread on the mailing list about linking to Denote-named files which are not `{txt,md,org}`. I think we can do this. Just need to make sure we don't try to visit the file for front matter retrieval (which I believe we already do, but I will double-check).

Okay, I am not sure. It should be doable!

PR #89, opened 2022-08-13
File types section

jeanphilippegg:
This is a proof-of-concept. I have some doubts, but I am sharing it with you.

I added a new section "File types" in the code. You can check the new **internal** variable `denote--file-types` and its docstring and other related functions.

It would still need refinements to be merged. We could then refactor many functions in denote.el to use the content of `denote--file-types`.

I copy here what I originally wrote for that pull request:

---

With `denote--file-types`, I did not consider user-friendliness. I just consolidated the various information we have about file types, without losing any of our existing functionalities. I suggest we first use `denote--file-types` internally for a while just to test how we like it and to allow us to make adjustments. We could disclose its existence to atanasj (with a warning) so that he can use it for its qmd and rmd files as soon as possible.

However, I think that ultimately it would be better to convert `denote--file-types` to `denote-file-types` as-is, even if it contains more than just a front-matter. It must also contains entries like title-key-regexp and keywords-key-regexp. I will use an example to illustrate the issue here. Let's say we have an Org front-matter as fixed by `denote-org-front-matter`. Internally, we need to look for a `#+title:` line. But it is also desirable that we tolerate a `#title    :` line. We could try to guess a forgiving title-key-regexp ourselves from the front matter, but we cannot predict all potential file types that users will create. I think we should let the users control this themselves.

Also, we must not forget that this user option would be for advanced users. Having a little complexity in it might not be too bad.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -642,7 +642,7 @@ identifier: %s
   "Extract keywords list from front matter KEYWORDS-STRING."
   (split-string keywords-string "[:,\s]+" t "[][ \"']+"))
 
-(defvar denote--file-types
+(defvar denote-file-types
   ;; If denote-file-type is nil, we use the first element
   ;; of denote-file-types for new note creation, which we want
   ;; to be org by default.
@@ -709,39 +709,39 @@ TYPE-INFO is a list of 8 elements:
 
 (defun denote--file-extension (file-type)
   "Return file type extension based on FILE-TYPE."
-  (nth 0 (alist-get file-type denote--file-types)))
+  (nth 0 (alist-get file-type denote-file-types)))
 
 (defun denote--front-matter (file-type)
   "Return front matter based on FILE-TYPE."
-  (nth 1 (alist-get file-type denote--file-types)))
+  (nth 1 (alist-get file-type denote-file-types)))
 
 (defun denote--title-key-regexp (file-type)
   "Return the title key regexp associated to FILE-TYPE."
-  (nth 2 (alist-get file-type denote--file-types)))
+  (nth 2 (alist-get file-type denote-file-types)))
 
 (defun denote--title-value-function (file-type)
   "Function to convert the title string to a front matter title.
 Based on FILE-TYPE."
-  (nth 3 (alist-get file-type denote--file-types)))
+  (nth 3 (alist-get file-type denote-file-types)))
 
 (defun denote--title-value-reverse-function (file-type)
   "Function to convert a front matter title to the title string.
 Based on FILE-TYPE."
-  (nth 4 (alist-get file-type denote--file-types)))
+  (nth 4 (alist-get file-type denote-file-types)))
 
 (defun denote--keywords-key-regexp (file-type)
   "Return the keywords key regexp associated to FILE-TYPE."
-  (nth 5 (alist-get file-type denote--file-types)))
+  (nth 5 (alist-get file-type denote-file-types)))
 
 (defun denote--keywords-value-function (file-type)
   "Function to convert the keywords string to a front matter keywords.
 Based on FILE-TYPE."
-  (nth 6 (alist-get file-type denote--file-types)))
+  (nth 6 (alist-get file-type denote-file-types)))
 
 (defun denote--keywords-value-reverse-function (file-type)
   "Function to convert a front matter keywords to the keywords list.
 Based on FILE-TYPE."
-  (nth 7 (alist-get file-type denote--file-types)))
+  (nth 7 (alist-get file-type denote-file-types)))
 
 (defun denote--get-title-line-from-front-matter (file-type)
   "Retrieve title line from front matter based on FILE-TYPE.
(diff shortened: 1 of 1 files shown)


protesilaos:
> [...] It must also contains entries like title-key-regexp and keywords-key-regexp. [...] I think we should let the users control this themselves.

Yes, I agree.

> Also, we must not forget that this user option would be for advanced users. Having a little complexity in it might not be too bad.

Exactly!  We already have such an advanced option for the `display-action` of the backlinks.  There is nothing wrong with those.

* * *

Please let me know when you are ready to proceed with this.

jeanphilippegg:
> Please let me know when you are ready to proceed with this.

Ok! What do you think of it? Is this a good direction? Should we keep `denote--file-types` internal for some time?

protesilaos:
Yes, I like this direction!  I think we should keep everything internal at least until we flesh it out completely.

The `denote--file-types` could be written right now as `denote-file-types`.  Its doc string could then warn that it is only for advanced users, is part of the current development cycle, may still be refined, and the like.

About this sort of value:

```elisp
    (markdown-yaml . (".md"
                      ,denote--yaml-front-matter
                      "^title\\s-*:"
                      denote--surround-with-quotes
                      denote--trim-whitespace-then-quotes
                      "^tags\\s-*:"
                      denote--format-keywords-for-md-front-matter
                      denote--extract-keywords-from-front-matter))
```

We could make the cdr a plist.  It will be easier to read.  Like:

```elisp
    (markdown-yaml :extension".md"
                   :format ,denote--yaml-front-matter
                   ...))
```

Perhaps there is scope for adding our heuristic here.  For example, if a user specifies only the `:extension` and `:format`, we can try our best to figure out that the keys are found with, say, `"^title\\s-*:"`.

Does this make sense or will it complicate things?  I basically suggesting a way that is easy enough for regular users but still gives maximum control to those who really need it.

jeanphilippegg:
> We could make the cdr a plist.

Yes, good idea!

> Perhaps there is scope for adding our heuristic here. [...] Does this make sense or will it complicate things?

It is a good idea! Though, I suggest we keep it for another iteration to keep it simple for us for now.

I will push new commits later today.

protesilaos:
> It is a good idea! Though, I suggest we keep it for another iteration to keep it simple for us for now.

Yes, just for us to consider.

> I will push new commits later today.

Thank you!

jeanphilippegg:
I have done a round a refinements. I did some quick tests and it seems to work.

Notable changes:

- The title lines of `denote-md-front-matter` and `denote-yaml-front-matter` contain `%s` (lower case s) instead of `%S`. This needs to be consistent for all formats because we do not know them all. Also it is possible that a file type formats the title as `<My title>`. We must let the user decide.

- Setting `denote-file-type` to nil (the default) is equivalent to the first item in `denote-file-types` (which happens to be Org by default). This is because we cannot assume org to be a valid file-type anymore. This does not change anything for users using default settings.

- The way I refactored `denote--rewrite-front-matter` (and `denote--rewrite-keywords`) is that I find the title line from the front matter using the title-key-regexp. For example, an org title line would be "#+title:     %s". However, it is possible that the line retrieved is "#+title:     %2s". An error will be raise because of the "%$2". This is currently broken. We should find a solution to this. Maybe we should use a custom tag `<<title>>` instead of "%s".

protesilaos:
Thank you!

About the `%S`, the idea was that the user might insert a title containing quotes.  YAML cannot parse something like this:

```yaml
title:  This is a 'simple' test
```

It needs it to be formatted thus:

```yaml
title:  "This is a 'simple' test"
```

Though yes, we would still get wrong syntax if the user inputs `"` instead of `'`.  Just noting this for your information on what the rationale was.

I think the `denote--filetype-heuristics` needs a stylistic change:

```elisp
(let ((found-type (seq-find ...
  ...
  (when found-type ...
```

If I am reading it correctly, this could be a `when-let`.

The lambdas in `denote--filetype-heuristics` are only specific to it?  Maybe they could be turned into their own name functions?

jeanphilippegg:
> Just noting this for your information on what the rationale was.

Ok, good to know! Thanks!

> If I am reading it correctly, this could be a `when-let`.

Yes. Fixed!

> The lambdas in `denote--filetype-heuristics` are only specific to it? Maybe they could be turned into their own name functions?

I moved one of the lambdas into its own function. The other is quite specific. I am not sure how to extract a useful function out of it.

protesilaos:
> I moved one of the lambdas into its own function. The other is quite specific. I am not sure how to extract a useful function out of it.

That's okay.  I wasn't sure either.  This is just a general idea to keep the code in a manageable state.

Please let me know when you think this can be merged.

jeanphilippegg:
I think it can be merged, but this involves a lot of changes. I would not be confident that no functionalities are broken. I quickly tested note creation, renaming, backlinks and they seem to work.

Users should probably be advised in the changelog about the %S changed to %s in markdown front matters.

Also, I chose some names for the properties in `denote-file-types`. You can change them if you think of better ones.

protesilaos:
Merged.  Thank you!

> Users should probably be advised in the changelog about the %S changed to %s in markdown front matters.

Indeed.  This is a breaking change for them.

> Also, I chose some names for the properties in `denote-file-types`. You can change them if you think of better ones.

Sure.  Let's see how it goes.  Note that the functions specified therein should all be public, but this is something we can do as development progresses.

PR #91, opened 2022-08-15
Some cleanups for denote--only-note-p

jeanphilippegg:
Cleanups following discussion in #90. Check the individual commits.

PR #94, opened 2022-08-16
Cleanup

jeanphilippegg:
Cleanup to remove checks from various functions as discussed in #91 and other minor simplifications.

I also added an `s` to `denote-backlink-mode` and `denote-backlink-mode-map` and made the latter obsolete and removed the todo item. I don't know if we can make `denote-backlink-mode` obsolete. Now, only `denote-link-backlink-button` and `denote-link--backlink-find-file` have a `backlink` without an `s`. It is good with me. Maybe you would prefer to have singular `backlink` everywhere?

### DIFF
--- a/README.org
+++ b/README.org
@@ -1202,12 +1202,12 @@ Reproducing it here for the sake of convenience:
         (window-width . 0.3)))
 #+end_src
 
-#+findex: denote-backlink-mode
-#+vindex: denote-backlink-mode-map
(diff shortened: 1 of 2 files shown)


protesilaos:
Merged.  Thank you!

protesilaos:
I an obsoletion for the `denote-backlink-mode` as well.  I can't imagine anyone using it, but better be safe.

jeanphilippegg:
Ok, good!

Issue #95, opened 2022-08-18
Using denote as journal or diary


atanasj:
The online manual provides an example of how to use `denote` as a journal... Is there a way to make it function my similar to `org-journal`, whereby a new note can be created for a day, week, etc., and then each entry is appended to the bottom with the time stamp as a sub-header? I've tried to hack something together, by my `elisp` is not very good... I thought it might be combination of the stuff below, but couldn't get it working…

```elisp
;; from the manual
(defun my-denote-journal ()
  "Create an entry tagged 'journal' with the date as its title."
  (interactive)
  (denote
   (format-time-string "%A %e %B %Y") ; format like Tuesday 14 June 2022
   '("journal"))) ; multiple keywords are a list of strings: '("one" "two")

(with-eval-after-load 'org-capture
  (add-to-list 'org-capture-templates
               '("J" "New Journal (with Denote)" plain
                 (file+headline denote-last-path (format-time-string "%A %e %B %Y"))
                 #'denote-org-capture
                 :no-save t
                 :immediate-finish nil
                 :kill-buffer t
                 :jump-to-captured t
                 :prepend t)))
```

protesilaos:
Hello @atanasj!

Denote is designed to work with a one-file-per-note workflow.  All the Denote commands which pertain to new note creation expect a file path to be returned.  They then use `find-file` on that path.

Denote does not have the means to add to an existing file.  That is an Org-specific feature.  Also, Denote does not have a date picker the way Org does.  That too is Org-specific.  In both cases the code is not trivial.  Just check the implementation of `org-read-date` to see what I mean.

Perhaps there is a way to hack this together, though it would still not make for an optimal choice, based on the requirements you have described.  I then wonder why not use `org-journal` directly?

I personally use Denote for journaling.  Each entry has a `journal` keyword (among others), which makes it easy to filter them from the rest of the notes.  The individual files also make it possible to share a file when I need to (happens from time-to-time with people I can trust).

jeanphilippegg:
Hello @atanasj !

I pasted below the `my-denote-journal` command I have in my config. Maybe you can adapt it to your specific needs.

- `C-c n j` creates a new note for today if none already exists. If it already exists, it opens today's note.
- `C-c n J` prompts for the date of the note you want to edit (or create).

The code:

```elisp
(defun first-file-with-substring (dir substring)
  "Return the first file in DIR containing SUBSTRING.
Return nil if there is no files with SUBSTRING in its name."
[...]

PR #97, opened 2022-08-19
Add option denote-date-prompt-use-org-read-date

jeanphilippegg:
As discussed in #95, this adds the option `denote-date-prompt-use-org-read-date`.

Note that `org-read-date`, when given no timestamp defaults to the current time, which would be good. However, the default seconds are always `00`. Defaulting to the current time for everything but the seconds is documented in their docstring. I have done nothing special to handle this better. This implies that a user who would create two notes within the same minute would have a conflicting identifier.

Given this, I am not sure if it would be better not to provide this option or to document this limitation.

Also, I did not load `org-read-date` conditionally to org being loaded because I did not know how to do it properly.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1128,9 +1128,12 @@ here for clarity."
 (defvar denote--date-history nil
   "Minibuffer history of `denote--date-prompt'.")
 
+(declare-function org-read-date "org" &optional with-time to-time from-string prompt default-time default-input inactive)
+
 (defun denote--date-prompt ()
   "Prompt for date."
-  (if denote-date-prompt-use-org-read-date
+  (if (and denote-date-prompt-use-org-read-date
+           (require 'org nil :no-error))
       (let* ((time (org-read-date nil t))
              (org-time-seconds (format-time-string "%S" time))
              (cur-time-seconds (format-time-string "%S" (current-time))))
--- a/denote.el
+++ b/denote.el
@@ -316,6 +316,16 @@ are described in the doc string of `format-time-string'."
   :package-version '(denote . "0.2.0")
   :group 'denote)
 
+(defcustom denote-date-prompt-use-org-read-date nil
+  "Whether to use `org-read-date' in date prompts.
+
+If non-nil, use `org-read-date'.
+
+If nil, input the date as a string, as described in `denote'."
+  :group 'denote
+  :package-version '(denote . "0.6.0")
+  :type 'boolean)
+
 (defcustom denote-templates nil
   "Alist of content templates for new notes.
 A template is arbitrary text that Denote will add to a newly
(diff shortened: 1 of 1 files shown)


protesilaos:
> This implies that a user who would create two notes within the same minute would have a conflicting identifier.
>
> Given this, I am not sure if it would be better not to provide this option or to document this limitation.

I feel this is the sort of "gotcha" that even if we document, it will still lead to bug reports and/or confusion.  Perhaps in that case we can add the seconds ourselves?  It is kinda tricky.

> Also, I did not load `org-read-date` conditionally to org being loaded because I did not know how to do it properly.

I believe this would suffice:

```elisp
(if (and denote-date-prompt-use-org-read-date (require 'org nil :no-error))
```

jeanphilippegg:
> I feel this is the sort of "gotcha" that even if we document, it will still lead to bug reports and/or confusion. Perhaps in that case we can add the seconds ourselves? It is kinda tricky.

Yes, I thought it would be hard, but I think I have something now...

> I believe this would suffice:
> 
> ```emacs-lisp
> (if (and denote-date-prompt-use-org-read-date (require 'org nil :no-error))
> ```

I still get a warning if I do that.

jeanphilippegg:
The code I have pushed is not right. We really have to be able to determine if the user has input a time or not. I can't find a solution to this.

protesilaos:
> I still get a warning if I do that.

It likely needs a `declare-function` before `denote--date-prompt`.  This declaration would be satisfied by the subsequent `require`.

> The code I have pushed is not right.

Hmm, this gets tricky.  I was expecting it to be a simple toggle and then just a call to the function, as you did originally.  Now I'm not sure what we should do.  My concern with hacks is that we are assuming a maintanance burden that may not be worth the trouble.

Still, the code is small so this is not a major issue.

> One hack would be to add the current seconds if `org-read-date` returns 0 as seconds.

I am fine with that if you think it is worth doing.  By "worth doing" I mean if you expect it to not cause us trouble down the line.

jeanphilippegg:
> I am fine with that if you think it is worth doing. By "worth doing" I mean if you expect it to not cause us trouble down the line.

Since nobody has asked for this feature, we could keep it in a branch for now. The code is completed though. There is a difference between both methods. Using the current method, a user typing a time "23:23" would get an actual time "23:23:00" which could lead to conflicts if they input the same time again. This is "fixed" for the method of `org-read-date`.

But I think we could do nothing for now and wait and see if there are needs for any improvements coming from the users.

protesilaos:
> Since nobody has asked for this feature, we could keep it in a branch for now. The code is completed though. 

In this case, I would rather add it directly to `main`.  If in the future we realise that it is causing too much trouble, we can deprecate it.  Though I don't expect this to happen.

What do you think?

> But I think we could do nothing for now and wait and see if there are needs for any improvements coming from the users.

Agreed!

jeanphilippegg:
> In this case, I would rather add it directly to `main`. If in the future we realise that it is causing too much trouble, we can deprecate it. Though I don't expect this to happen.
> 
> What do you think?

Yes, I agree! It can also be merged as-is.

protesilaos:
Merged.  Thank you!  I will update the manual a bit later.

PR #99, opened 2022-08-21
Also remove leading dash when sluggifying title.

pprevos:
Slight update to `denote--slug-hyphenate` to also remove leading dashes. Use case: file names wit tripple dashes when accidentally adding spaces at start of title/


protesilaos:
Merged.  Thank you!

PR #100, opened 2022-08-21
Fix denote--filetype-heuristics for md files when no key regexps match

jeanphilippegg:
`denote--filetype-heuristics` should always return an md filetype if the
file extension is md even if no key regexps match in the file. In that
case, we use the first md type in `denote-file-types`.

markdown-yaml should appear before markdown-toml in `denote-file-types`
because it used to be the default markdown type of
denote--filetype-heuristics.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -665,15 +665,6 @@ identifier: %s
      :keywords-key-regexp "^#\\+filetags\\s-*:"
      :keywords-value-function denote--format-keywords-for-org-front-matter
      :keywords-value-reverse-function denote--extract-keywords-from-front-matter)
-    (markdown-toml
-     :extension ".md"
-     :front-matter ,denote-toml-front-matter
-     :title-key-regexp "^title\\s-*="
-     :title-value-function denote--surround-with-quotes
-     :title-value-reverse-function denote--trim-whitespace-then-quotes
-     :keywords-key-regexp "^tags\\s-*="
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you!

> markdown-yaml should appear before markdown-toml [...]

This will be problematic if we make it a `defcustom`.  We cannot guarantee how the user will write the values.

jeanphilippegg:
> This will be problematic if we make it a `defcustom`. We cannot guarantee how the user will write the values.

I meant to say that the default should be markdown-yaml before markdown-toml. This is not mandatory. If the user change the order, the consequence will be that markdown-toml will be prefered over markdown-yaml when there is no key regexp matches.

protesilaos:
Ah okay, thanks for the clarification!

protesilaos:
Shall I merge this, or do you want to tweak it further?

jeanphilippegg:
It is complete. Defaulting to the first filetype when no key regexps match is documented in the *internal* function `denote--filetype-heuristics`. Maybe we would like to also make this known to the users? (Defaulting to markdown-yaml was not documented in the previous versions.)

protesilaos:
I think documenting it is not needed.  It adds too much to the doc string.  If a problem arises, we deal with it on the spot and adapt things accordingly.

What do you think?

jeanphilippegg:
That's good with me!

protesilaos:
Merged.  Thank you!

PR #101, opened 2022-08-22
Fix denote-dired-rename-markd-files

jeanphilippegg:
- keywords should be allowed to be nil.
- The answer to yes-or-no-p should not yield the error "No marked files"
  if the answer is no.


protesilaos:
Merged.  Thank you!

PR #102, opened 2022-08-22
Return empty string when no keywords in org notes

jeanphilippegg:
I am not sure of the right syntax when there are no keywords in an org file, but `::` looks strange. An empty string seems more appropriate, but I could not find documentation on this.


protesilaos:
Makes sense.  Thank you!

PR #103, opened 2022-08-26
Added a prompt to denote-rename-file-using-front-matter.

pprevos:
Hi Prot,

I have been using `denote-rename-file-using-front-matter` quite a bite and kept forgetting to save before evoking it so I added a prompt to the function for convenience.

P:)

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1602,7 +1602,9 @@ the source of truth in this case to avoid potential breakage with
 typos and the like."
   (interactive (list (buffer-file-name)))
   (when (buffer-modified-p)
-    (user-error "Save buffer before proceeding"))
+    (if (y-or-n-p "Would you like to save the buffer?")
+        (save-buffer)
+      (user-error "Save buffer before proceeding")))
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you @pprevos!  This makes sense.  Before I merge it, please tell me what the status is with the copyright assignment to the Free Software Foundation.  Did you do it?

I just ask because you are on the limit of what you can contribute without it and it is better to be on the safe side legally.

pprevos:
Hi Prot,

I have sent the form to FSF a few days ago. No response yet.

protesilaos:
Very well!  Usually this takes about a week.  Let's wait.

protesilaos:
@pprevos I am merging this.  I checked your previous commits and we are just on the limit.  There should be no problem with this one.  Though the next will definitely need copyright assignment.

Thanks again for your contirbution!

pprevos:
Thanks Prot, no need to delete the fork just yet. I'll let you know when the copyright assignment is sorted.

protesilaos:
Thank you!

PR #108, opened 2022-09-30
Add denote-link-add-missing-links

EFLS:
Similar to `denote-link-add-missing-links', but only insert links that are not yet included in the current buffer.

### DIFF
--- a/README.org
+++ b/README.org
@@ -1204,6 +1204,17 @@ above is for interactive usage.
 Links are created only for files which qualify as a "note" for our
 purposes ([[#h:fc913d54-26c8-4c41-be86-999839e8ad31][Linking notes]]).
 
+** Insert links, but only those missing from current buffer
+
+#+findex: denote-link-add-missing-links
+As a variation on the =denote-link-add-links= function, one may wish to
+only include 'missing links', i.e. links that are not yet present in
+the current file.
+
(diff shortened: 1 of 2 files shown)


protesilaos:
Thank you @EFLS!  This looks good.  I will merge it now, but some minor points:

- In the manual, when we reference symbols we use the tilde instead of the equals sign.  The reason is that it has a different output in .texi files which is supposed to be semantically correct.  I learnt this with my modus-themes when I added their manual to emacs.git.
- After the `if-let*` there is another `let` form.  It could be merged into the `if-let*` instead.
- The `user-error` must not end with a full stop because the compiler will complain about it (pedantic, I know).
- I think the `user-error` could be a generic `message` in the sense that this is expected behaviour.

Now I am merging this.  Will also update the "Acknowledgements" to mention you under the code contributions.  Please let me know the FSF paperwork is done.

EFLS:
Agreed with all points, except

> After the if-let* there is another let form. It could be merged into the if-let* instead.

Not exactly as the values in the let form are allowed to be nil (when there are no links yet in the current buffer). If I understand the differences between let and if-let correctly, that is.

EFLS:
Ok perfect!

protesilaos:
Thanks again!  Looking forward to the next ones, as we discussed on the mailing list.

PR #109, opened 2022-10-08
Use region as link title when inserting a link

bitspook:
It is a common use-case to create a link in existing notes. In such cases, some text is already there, and user just want to create a link on that text (instead of inserting the link's title). With this change, if a region is active when user tries to insert a link, the region is replaced with the link while using the region's content as link's title.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -2241,10 +2241,13 @@ title."
 
 (defun denote-link--format-link (file pattern)
   "Prepare link to FILE using PATTERN."
-  (let* ((file-id (denote-retrieve-filename-identifier file))
+  (let* ((region-text (when (region-active-p)
+                        (buffer-substring (region-beginning) (region-end))))
+         (file-id (denote-retrieve-filename-identifier file))
          (file-type (denote-filetype-heuristics file))
          (file-title (unless (string= pattern denote-link--format-id-only)
-                       (denote--retrieve-title-or-filename file file-type))))
+                       (or region-text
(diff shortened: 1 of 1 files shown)


protesilaos:
Hello @bitspook!  I like this idea.  I think we should tweak how it is implemented so that:

+ The `region-active-p` is only checked by `denote-link`.  No need to have it twice.

+ Add a new `TITLE` argument for `denote-link--format-link` (or name it whatever you think is most descriptive).  If non-nil, it is used as the link's anchor text.  If nil we rely on the `(denote--retrieve-title-or-filename file file-type)` approach, which is the current behaviour.  The non-nil `TITLE` will come from `denote-link` if the region is active.

+ In `denote-link` the check for the region will handle the extraction and replacement of the text.  Though use `buffer-substring-no-properties`, just to be sure.

What do you think?

EFLS:
I just want to chime in that this looks like a great addition. I've included a hack in my config to do this for the time being, but it would be great to see this in Denote itself. And the comments by prot above seem to make a lot of sense.

bitspook:
Thank you for entertaining the idea and giving me a chance to fix the code myself.

I've updated the PR. Please take a look @protesilaos I've called the inserted link's display-text `description`, following the terminology used by org-mode.

protesilaos:
@bitspook Merged.  Thank you!  Note that your patch is small, but for future contributions you will need to assign copyright to the Free Software Foundation.  I can explain, if you need me to.

@EFLS Thanks for the feedback!  I think this is a nice addition.

Issue #117, opened 2022-12-13
conceptual similarities to date2name and filetags


nbehrnd:
Because of GitHub's panel «packages which might be of interest for you», I just discover your approach.

Conceptually, the provision of a consistent time stamp and assigning a keyword from a constrained vocabulary bears similarities to Karl Voit's earlier implementations in Python [date2name](https://github.com/novoid/date2name) and [filetags](https://github.com/novoid/filetags) as applications for the CLI.  So, perhaps his talk [*Grazer Linuxtage 2018 - The Advantages of File Name Conventions and Tagging*](https://www.youtube.com/watch?v=rckSVmYCH90) (link to a 45min video recording on youtube) equally is of interest for you/users of your approach for Emacs.

novoid:
Coming from Org-mode and its Zettelkasten implementations, I did not realize the similarities between denote and filetags/date2name myself. ;-)

But yes, there are similarities in the concept.

@protesilaos, how would you value the use of denote for file management without knowledge-management/Zettelkasten/...?

protesilaos:
Hello @nbehrnd and thanks for sharing this information!  I cannot watch the video right now.  Will do it over the weekend.

I follow @novoid blog and have read many articles therein, though I have not used those packages you mention.  The reason is that I learnt about them after I already had my own workflow in place.

As Karl notes, the concept is the same: give file names a predictable file name and be consistent about it.  Once those are granted, I don't think there is an inherently better way of doing things.  It comes down to what the user prioritises.  For example, Karl's dates are more readable and thus easy to parse with the eye, though they use hyphens which means that it is more difficult to anchor a search at a certain section of the file name.  By contrast, Denote's dates (the identifier) are more terse but free up the hyphen to be used as a delimiter.  To quote from the manual:

> File names have three fields and two sets of field delimiters between them:
> 
>     DATE--TITLE__KEYWORDS.EXTENSION
> 
> [...] Consider this example:
> 
>     20220621T062327--introduction-to-denote__denote_emacs.txt
> 
> You will notice that there are two matches for the word `denote`: one in the title field and another in the keywords' field.  Because of the distinct field delimiters, if we search for `-denote` we only match the first instance while `_denote` targets the second one.  When sorting through your notes, this kind of specificity is invaluable---and you get it for free from the file names alone!
>
> <https://protesilaos.com/emacs/denote#h:1a953736-86c2-420b-b566-fb22c97df197>

* * *

[...]

PR #123, opened 2023-01-02
add crosslink to Karl Voit's work

nbehrnd:
Denote started with management of thoughts/ideas in Emacs org-mode
and extends to files.  Some of the tools initiated by Karl Voit do
offer the addition of time stamps, key words from a controlled
vocabulary, etc on files (e.g., date2name) which may be processed
further (memacs).  It is worth to hint to them, as introduced by
this commit, in appreciation of an earlier invite.[1]

[1] https://github.com/protesilaos/denote/issues/117#issuecomment-1354372589

### DIFF
--- a/README.org
+++ b/README.org
@@ -3253,15 +3253,16 @@ should not be mutually exclusive.
 
 If I missed something, please let me know.
 
-** Alternative ideas with Emacs and further reading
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged.  Thank you!

PR #137, opened 2023-02-18
Add function denote-change-file-type

jeanphilippegg:
Hi Prot,

It has been a while since my last contribution. I have just been using
Denote seamlessly. However, I kept having this issue. By default, when I
create a note, I use txt as the default format. Sometimes, I realize
that I actually wanted to use org. In this case, I just started over and
typed the title and keywords again. Other times, I have an old text note
which I want to convert to org to use Org's functionalities.

I have written a function (denote-change-file-type) to convert the file
extension and front matter to another format. This function only changes
the file extension and ADD a front matter at the beginning of the file,
retrieving the information from the old one. The old front matter stays
there. There are probably many other modifications to be made when
changing the file type and this function does not attempt to modify
anything else in the file (headings and other file type specific
elements).

I made this function a while ago. I share it with you if you think it
can be a useful addition to Denote.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -2155,6 +2155,46 @@ files)."
           (denote--add-front-matter new-name title keywords id file-type))))))
 
 ;;;###autoload
+(defun denote-change-file-type (file new-file-type)
+  "Change file type of FILE and add an appropriate front matter.
+
+If in Dired, consider FILE to be the one at point, else prompt
+with minibuffer completion for one.
+
+Add a front matter in the format of the NEW-FILE-TYPE at the
+beginning of the file.
+
+The title is retrieved from a line starting with a title field in
+the file's contents, depending on the previous file type (e.g.
+#+title for Org). The same process applies for keywords and id.
+
+As a final step, ask for confirmation, showing the difference
+between old and new file names.
+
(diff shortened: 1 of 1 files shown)


protesilaos:
Hello @jeanphilippegg!  I am happy to see you again.

> By default, when I create a note, I use txt as the default format. Sometimes, I realize that I actually wanted to use org.
>
> [...]
>
> I made this function a while ago. I share it with you if you think it can be a useful addition to Denote.

I had this issue as well.  I would rename the file manually and then insert the front matter.  I thought it was just a personal problem, but now that you mention it I agree we need this.

I am merging it now but I will write the entry in the manual tomorrow.

jeanphilippegg:
> I am merging it now but I will write the entry in the manual tomorrow.

Glad I could help! I am not actively following Denote's development, but I still use it and keep an eye on what is going on!

Have a good day!

protesilaos:
> I am not actively following Denote's development, but I still use it and keep an eye on what is going on!

That's okay.  The only area I have not gotten into yet is `denote-modules`, which was added recently by @nobiot.  I will definitely do it before the new release of Denote.

> Have a good day!

You too!

PR #158, opened 2023-07-17
Replace seq-keep with version-agnostic variant

adamruzicka:
I'm on emacs 28.2 and I don't seem to have `seq-keep` available. This seems to work

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1620,13 +1620,14 @@ where the former does not read dates without a time component."
 
 (defun denote--buffer-file-names ()
   "Return file names of Denote buffers."
-  (seq-keep
-   (lambda (buffer)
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you @adamruzicka!  This changes the return value from file paths to buffer objects.  Maybe we can use this instead, which works with all Emacs versions?

```elisp
(defun denote--buffer-file-names ()
  "Return file names of Denote buffers."
  (delq nil
        (mapcar
         (lambda (buffer)
           (when-let (((buffer-live-p buffer))
                      (file (buffer-file-name buffer))
                      ((denote-file-is-note-p file)))
             file))
         (buffer-list))))
```

adamruzicka:
That works nicely, thank you for the suggestion

PR #159, opened 2023-07-24
Duplicate ids

jeanphilippegg:
Hi Prot,

I saw the changes about unique ids. I suggest this solution below. My
goal was to make ids robust everywhere, fast and without optional
parameters or options.

Affected functions:
- `denote`: Always creates a new unused id (see how below). No need to
  barf anymore.
- `denote-org-capture-with-prompts`: Always creates a new unused id. No
  need to barf anymore.
- `denote-barf-duplicate-id`: Obsolete as it is not used anymore.
- `denote-retrieve-or-create-file-identifier`: Removed. Use the
  functions 'denote-retrieve-filename-identifier' (already exists) and
  'denote-create-unique-file-identifier' (new).
- `denote-retrieve-filename-identifier`: New optional parameter
  'no-error' to return nil instead of an error.
- `denote-create-unique-file-identifier`: New.
- `denote-change-file-type`: Minor change. We do not check for
  duplicate ids in this case. We now enforce that the file must have an
  identifier in its name.
- `denote-add-front-matter`: Minor change. Do not check for duplicate
  ids. It does not rename a file anyway, by design.
- `denote-rename-file`: We now check for duplicate ids. Should be as
  fast as the current solution. In fact, if the renamed file already has
  an id, we do not check for duplicate ids, so this should be faster.
- `denote-rename-file-using-from-matter`: If, somehow, the file name
  did not already have an identifier, this function created one that
  was not necessarily unique. Now, we check and make sure it is unique.
  This is slower than before, but no more than the 'denote' command and
  only if the file lacks an identifier.
- `denote-dired-rename-marked-files`: Ensure ids are unique by checking
  against all denote notes. This may be slower than the current
  solution, but it is not slower than a single 'denote' command and this
  check only happens when there is a file without id.
- `denote-dired-rename-marked-files-using-from-matter`: The doc said
  that files without an id are ignored, but this was not enforced. A new
  identifier is created if none is present. I modified the function to
  enforce this.

The solution:
- Use a hash table for used ids. In the general case (when there is no
  duplicate ids), the `denote` function already had to go through ALL
  denote notes and compare with their ids. The modification I made
  should not be significantly slower, if at all.
- The way to find a new unique identifier has changed. In the current
  version, the current time is used. In my proposed solution, I just
  increment the identifier by 1 second until an unused id is found.

I have not tested the affected functions extensively. Please, do some
quick tests before merging.

Good day!

### DIFF
--- a/README.org
+++ b/README.org
@@ -3038,7 +3038,7 @@ invoke this command."
   (image-dired--with-marked
    (when-let* ((file (image-dired-original-file-name))
                (dir (file-name-directory file))
-               (id (denote-retrieve-or-create-file-identifier file))
+               (id (denote-retrieve-filename-identifier file))
                (file-type (denote-filetype-heuristics file))
                (title (denote--retrieve-title-or-filename file file-type))
                (extension (file-name-extension file t))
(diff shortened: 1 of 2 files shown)


protesilaos:
Hello @jeanphilippegg!  I haven't had the time to check this.  I am just commenting here to let you know that I have seen it.  I will review your changes as soon as possible and get back to you.  Probably in the coming days.

jeanphilippegg:
Take your time! Have a good day!

protesilaos:
Hello again @jeanphilippegg!  Just to inform you that I have not forgotten this.  I am not able to commit enough time right now, as I still do not have electricity at home.  I almost there, so I expect to revisit this and merge it by the end of next week

More to follow!

protesilaos:
Hello @jeanphilippegg!  It took me longer than expected.  I am back and ready for action.  What is the status of this PR?  Should I start testing it or do you intend to make further changes?

jeanphilippegg:
No worry about the delay! Yes, it can be tested. I only really use the `denote` command. It works well. The other commands should be tested.

jeanphilippegg:
Hello Prot!

I have seen part of your video about the maintainance of denote. This pull request is big and I understand that it must be hard to understand as it is. I suggest that we close it without merging it. I will create smaller pull requests over the weekend to make it simpler for you and without merge conflicts.

The goals of this pull request were:
- Make everything robust about unique ids and without optional parameters.
- Split `denote-retrieve-or-create-file-identifier` and use the resulting functions as needed.

I am willing to create smaller pull requests to make it all easier. What do you think of this approach?

protesilaos:
Hello @jeanphilippegg!  It is not a problem on my end.  Actually, I was planning to announce another live stream to tackle this and the other issue with Jean-Charles Bagneris: https://lists.sr.ht/~protesilaos/denote/%3C8171de83-d748-4d86-935f-32ac72403173%40app.fastmail.com%3E

If you do make the changes you suggest, then I will postpone this plan.  Otherwise, do not worry and I will get to it soon.  I want to merge this and am looking forward to it!

jeanphilippegg:
Ok, you can test and merge it. I rebased my branch on top of main so that it will not have conflicts with the newest changes. I suggest you check each of my commits individually instead of the whole patch to have an easier time understanding it.

protesilaos:
Hi @jeanphilippegg!  I just merged your changes.  Thank you!  I still have not tested everything, though it looks fine.  I will continue checking things as we move towards version 2.1.0.

jeanphilippegg:
You're welcome! I will also check the mailing list and here for any issues that might arise from these modifications.

PR #162, opened 2023-08-11
Get rid of negated predicates

sirikid:
Hi, Prot! I watched your stream about denote and noticed that there are negated predicates which can be avoided by using the appropriate filtering functions.


protesilaos:
Merged.  Thank you!

[ I am slow to respond because I still don't have electricity at home.  Almost there! ]

PR #173, opened 2023-10-09
Use templates for creating journal entries

kandread:
There's probably better ways to write this function. Also, naming convention for variables might be off.
Thanks!


protesilaos:
> There's probably better ways to write this function.

No worries.  This works.  We can always refine things as we go.  One obvious tweak I see right now is to use `(alist-get 'journal denote-templates)` instead of `(cdr (assoc 'journal denote-templates))`.  Will do it myself,

Merging now.  Thanks!

PR #174, opened 2023-10-15
Small cleanup

jeanphilippegg:
I checked and removed the function `denote--default-dir-has-denote-prefix' as it does not seem necessary. I also simplified `denote--dir-in-denote-directory-p'. The returned value is not used.


protesilaos:
Merged. Thank you!

PR #175, opened 2023-10-15
Add denote-filename-is-note-p and use it in denote--buffer-file-names

jeanphilippegg:
I fixed `denote--buffer-file-names`. At first, this function was meant
to return buffers (including unsaved ones). At some point, I made it
check that the file was an actual (saved) file. This was not intended.
To fix this, I created a new function `denote-filename-is-note-p` that
does not check that the file is an actual file, but only checks that the
filename is a valid name for a Denote note. `denote--buffer-file-names`
should use this function instead. `denote--buffer-file-names` is used in
the detection of duplicate ids and in the function
`denote-link-dired-marked-notes`.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -692,15 +692,20 @@ signatures and keywords."
 (defun denote-file-is-note-p (file)
   "Return non-nil if FILE is an actual Denote note.
 For our purposes, a note must not be a directory, must satisfy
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you! I will make some stylistic tweaks to it but it is easier to do it on my end that exchanging messages.

PR #177, opened 2023-10-16
Refactor denote-rename-buffer

jeanphilippegg:
Good job on `denote--file-with-temp-buffer`! I had noticed that unsaved
buffers where not renamed properly, but had not come up with a solution.

Some comments:

- `denote-rename-buffer` is added to `find-file-hook`. I think some
  checks should be done early to determine if the renamed buffer can be
  renamed (ie is a Denote note).

- Each specifier individually executes the
  `denote-rename-buffer--common-checks` to retrieve the file and the
  type.

- We can only check for `string-blank-p` in
  `denote-rename-buffer--with-unique-name`. It already checks for
  `string-empty-p`.

Some other general comments:

- The `or` condition in `denote--file-with-temp-buffer-1` does not seem
  to make sense as it is. It seems to only consider `file-exists` in the
  end.

### DIFF
--- a/denote-rename-buffer.el
+++ b/denote-rename-buffer.el
@@ -88,59 +88,45 @@ buffer will be used, if available."
   :package-version '(denote . "2.1.0")
   :group 'denote-rename-buffer)
 
-(defun denote-rename-buffer--common-check (buffer)
-  "Determine if BUFFER shall be renamed.
-Return the file path and the type of it as a cons cell."
-  (when-let ((file (buffer-file-name buffer))
-             ((denote-file-has-identifier-p file))
-             (type (denote-filetype-heuristics file)))
-    (cons file type)))
+(defun denote-rename-buffer--get-title (file type)
+  "Return Denote title of FILE, given TYPE."
+  (denote-retrieve-title-value file type))
 
-(defun denote-rename-buffer--get-title (buffer)
-  "Return Denote title of BUFFER."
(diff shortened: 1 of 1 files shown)


protesilaos:
> Date: Sun, 15 Oct 2023 19:06:27 -0700
>
> Good job on `denote--file-with-temp-buffer`! I had noticed that unsaved
> buffers where not renamed properly, but had not come up with a solution.

Thank you! I think this does not cause any regression. I tried with
linking and renaming. Everything seems fine.

> Some comments:
>
> - `denote-rename-buffer` is added to `find-file-hook`. I think some
>   checks should be done early to determine if the renamed buffer can be
>   renamed (ie is a Denote note).

Correct!

> - Each specifier individually executes the
>   `denote-rename-buffer--common-checks` to retrieve the file and the
>   type.

Oh, right! I was too happy to add the specifiers and forgot to optimise
them!

> - We can only check for `string-blank-p` in
>   `denote-rename-buffer--with-unique-name`. It already checks for
>   `string-empty-p`.

Very well! There may be one more case where this is happening. I will
check.

> Some other general comments:
>
> - The `or` condition in `denote--file-with-temp-buffer-1` does not seem
>   to make sense as it is. It seems to only consider `file-exists` in the
>   end.

I am happy to improve this.

* * *

Mergin your changes now. Thank you!

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #178, opened 2023-10-22
Refactor denote-link--map-over-notes

jeanphilippegg:


protesilaos:
Merged. Thank you!

* * *

By the way, I recently refactored the renaming commands. If you have time, please let me know what you think. I plan to make a minor change to a prompt text, though the functionality is there.

jeanphilippegg:
I think it is good!

I see two issues with the code in its current state:

- `denote--rename-file-subr` needs to have `used-ids` as a parameter and it needs to pass it as a parameter to `denote-create-unique-file-identifier`. The way `denote--rename-file-subr` is called in `denote-dired-rename-files` seems to suggest that this was intended at some point.
- In `denote-dired-rename-files` and `denote-dired-rename-marked-files-with-keywords`, it is important to call `denote--rename-file` with `used-ids`. It will still work if you don't, but it will be slow if you renamed a large set of files. For example, if you rename 100 files, you will loop through all denote notes 100 times instead of only once. In general, we need to be careful with that with we use `denote-create-unique-file-identifier` inside a loop.

Other than that, I think it is good refactor!

PR #182, opened 2023-10-28
Add used-ids parameter to denote--rename-file-subr

jeanphilippegg:
Hello! I replied in the other pull request (#178) where you asked me for my
thoughts on your refactoring of the renaming functions, but you might
have missed my comment.

At first glance, I thought the `denote--rename-file-subr` function was a
good abstraction. However, I noticed that it could only be used to
remove duplication in some functions and it introduces some duplication
of its own. It also introduced some slowness in
`denote-dired-rename-marked-files`. This pull request fixes the slowness
part.

- `denote-file-type-heuristics` is called inside and outside
  `denote--rename-file-subr` in some cases.

- Same thing with `denote-retrieve-filename-identifier` and
  `denote-create-unique-file-identifier`. The code has to be both inside
  and outside in some cases.

- The handling of the title, keywords, signature is split between
  `denote--rename-file-subr` and its callers.

- There is similar code in `denote-change-file-type-and-front-matter`
  and `denote-add-front-matter`. But it was not quite the same, so it
  cannot use `denote--rename-subr`.

Given the above and that `denote--rename-file-subr` ends up being used
in only three places, I think it introduces more complexity than it
removes. I tried some refactoring, but I was not satisfied with what I
came up with.

Still, this pull request fixes the issue I mentionned earlier.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -2310,7 +2310,7 @@ Throw error is FILE is not regular, else return FILE."
 (defvar denote-rename-max-mini-window-height 0.33
   "How much to enlarge `max-mini-window-height' for renaming operations.")
 
-(defun denote--rename-file-subr (file identifier title keywords signature &optional ask-date no-confirm)
+(defun denote--rename-file-subr (file identifier title keywords signature &optional ask-date no-confirm used-ids)
   "Subroutine for `denote-rename-file' and `denote-dired-rename-files'.
 
 - FILE is the target of the rename operation.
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you @jeanphilippegg!

> I think it introduces more complexity than it removes.

Do you think it is better to remove it altogether than try to add on top of it? I am fine to revert to what we had before.

jeanphilippegg:
Sorry about the delay. Yes, I think it would be simpler to revert it until we can find some other way to simplify the code.

jeanphilippegg:
Sorry for the delay! Yes, I think it would be simpler to revert it until we find another way to simplify this code.

protesilaos:
Thank you @jeanphilippegg! I made the changes. Please let me know if you have any further thoughts on this. If not, I think we are ready for version 2.1.0: I will prepare it this week.

PR #187, opened 2023-11-04
Small refactor

jeanphilippegg:
Nothing major. I just refactored `denote-rename-file-using-front-matter` and I made `used-ids` parameter non optional in `denote-create-unique-filename-identifier`. This will force us to use `denote--get-all-used-ids` whenever we want to use it. In general, when we would use `denote-create-unique-filename-identifier` without `used-ids`, there was a risk that we would introduce some performance issues. Now, it will be more obvious if we do so.

I think it is all good for 2.1.0!

### DIFF
--- a/README.org
+++ b/README.org
@@ -3827,7 +3827,8 @@ might change them without further notice.
 
 #+findex: denote-create-unique-file-identifier
 + Function ~denote-create-unique-file-identifier~ :: Create a new unique
-  =FILE= identifier.  The conditions are as follows:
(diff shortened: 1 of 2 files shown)


protesilaos:
Thank you @jeanphilippegg! Merged it.

> I think it is all good for 2.1.0!

I will try to publish it this week. Thanks again for all your contributions!

jeanphilippegg:
You're welcome! Thank you as well

PR #195, opened 2023-11-25
Refactor denote-directory-files and deprecate functions

jeanphilippegg:
I have refactored `denote-directory-files` to make it do the job of `denote-all-files`, `denote-directory-files-matching-regexp` and `denote-directory-text-only-files`. I deprecated these functions, but kept them for now because some people might have started to use them.

This is just a suggestion. I don't think it is bad that we have separate functions, but if you think this is a good change, it is ready to be merged.


protesilaos:
Merged. Thank you!

PR #199, opened 2023-12-09
Refactor

jeanphilippegg:
Hello Prot!

This is a big pull request. No functional changes though.

- I have refactored `denote-file-has-identifier-p` to remove code
duplication.

- I have refactored `denote-add-front-matter`.

- I have streamlined the `denote-retrieve-filename-*` functions.
Previously, depending on the function, they were returning nil, an error
or an empty string when the element was not found. Now, they all do the
same thing and return nil when not found. I have adjusted the code
wherever they are called to handle the empty string and error cases.
Also, `denote-retrieve-filename-base` had an optional parameter to
fallback on the actual filename. I have done some refactoring to do this
work elsewhere (`denote--retrieve-title-or-filename`).

- I have renamed the functions `denote-retrieve-*-value` and
`denote-retrieve-*-line` to `denote-retrieve-front-matter-*-value` and
`denote-retrieve-front-matter-*-line` respectively so they are more like
the `denote-retrieve-filename-*` functions. Let me know if you think
this is an improvement. If not, I can remove this commit.

- I have removed `denote-org-capture-with-prompts` because I made
 `denote-org-capture` obey `denote-prompts`. It now handles signature
 as well.

### DIFF
--- a/README.org
+++ b/README.org
@@ -3957,30 +3957,30 @@ might change them without further notice.
   To only return an existing identifier, refer to the function
   ~denote-retrieve-filename-identifier~.
 
-#+findex: denote-retrieve-title-value
-+ Function ~denote-retrieve-title-value~ :: Return title value from
+#+findex: denote-retrieve-front-matter-title-value
++ Function ~denote-retrieve-front-matter-title-value~ :: Return title value from
   =FILE= front matter per =FILE-TYPE=.
 
-#+findex: denote-retrieve-title-line
-+ Function ~denote-retrieve-title-line~ :: Return title line from
+#+findex: denote-retrieve-front-matter-title-line
++ Function ~denote-retrieve-front-matter-title-line~ :: Return title line from
   =FILE= front matter per =FILE-TYPE=.
 
-#+findex: denote-retrieve-keywords-value
-+ Function ~denote-retrieve-keywords-value~ :: Return keywords value
+#+findex: denote-retrieve-front-matter-keywords-value
++ Function ~denote-retrieve-front-matter-keywords-value~ :: Return keywords value
   from =FILE= front matter per =FILE-TYPE=. The return value is a list
   of strings. To get a combined string the way it would appear in a
-  Denote file name, use ~denote-retrieve-keywords-value-as-string~.
+  Denote file name, use ~denote-retrieve-front-matter-keywords-value-as-string~.
 
-#+findex: denote-retrieve-keywords-value-as-string
-+ Function ~denote-retrieve-keywords-value-as-string~ :: Return
+#+findex: denote-retrieve-front-matter-keywords-value-as-string
++ Function ~denote-retrieve-front-matter-keywords-value-as-string~ :: Return
   keywords value from =FILE= front matter per =FILE-TYPE=. The return
   value is a string, with the underscrore as a separator between
   individual keywords. To get a list of strings instead, use
-  ~denote-retrieve-keywords-value~ (the current function uses that
+  ~denote-retrieve-front-matter-keywords-value~ (the current function uses that
   internally).
 
-#+findex: denote-retrieve-keywords-line
-+ Function ~denote-retrieve-keywords-line~ :: Return keywords line
+#+findex: denote-retrieve-front-matter-keywords-line
++ Function ~denote-retrieve-front-matter-keywords-line~ :: Return keywords line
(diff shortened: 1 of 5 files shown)


jeanphilippegg:
I would like to have your opinion on this. Right now, `denote-sluggify`
has an optional `component` parameter, but it never makes sense to call
it with anything but `'title` because the function uses
`denote--slug-hyphenate` and `denote--slug-no-punct` in ways that are
specific to `'title` only. In fact, we always call it only with
`'title` in Denote.

I see 2 solutions to this:

- Rename `denote-sluggify` to `denote-sluggify-title` and remove the
  `component` parameter.

- Keep `denote-sluggify` with its `component` parameter, but remove
  `denote-sluggify-keywords` and `denote-sluggify-signature` and move
  their code in `denote-sluggify`.

The second option implies that the STR parameter would sometimes be a
string to handle title or signature and sometimes a list of keywords. I
am not sure if this is desirable.

What do you think?

jeanphilippegg:
I added some commits. They are ready to be merged!

Good job on Denote's development by the way!

jeanphilippegg:
I have rebased my commits to make them conflict-free with your changes
for version 2.2.0. They are ready to be merged.

I just wanted to expand on my refactor of the
`denote-retrieve-filename-*` functions. I choose `nil` (instead of the
empty string) when the item is not found because I think it is more
fitting for this meaning. These functions have been simplified and they
are all implemented in the same way. I see them more as low-level
functions. Their callers will decide what to do when a component is not
found (use an empty string or signal an error or fall back on the
filename).

protesilaos:
Hello @jeanphilippegg and thank you for this!

>  These functions have been simplified and they
are all implemented in the same way. I see them more as low-level
functions. Their callers will decide what to do when a component is not
found (use an empty string or signal an error or fall back on the
filename). 

This is fine. We just need to make sure we preserve the current behaviour of ignoring/removing the given file name component.

About `denote-sluggify`, I think it is this way because I was planning to have one function for all cases but then encountered some exception which I solved ad-hoc with another function. So here we are... The idea is to have a single function that does the right thing.

The renamed functions are fine, though we need to add an alias there so that the old names work. The old names are public, so we must assume somebody is using them.

protesilaos:
> I have removed `denote-org-capture-with-prompts` because I made
`denote-org-capture` obey `denote-prompts`. It now handles signature
as well.

I forgot to comment on this. I think the `denote-org-capture-with-prompts` function has its use case to produce separate templates. An analogue of what we have with `denote-type`, `denote-signature`, etc.

I agree that `denote-org-capture` should obey the `denote-prompts`.

jeanphilippegg:
> This is fine. We just need to make sure we preserve the current behaviour of ignoring/removing the given file name component.

Ok, it is ready to be merged. This refactor should not have changed the
way it worked before.

> About `denote-sluggify`, I think it is this way because I was planning to have one function for all cases but then encountered some exception which I solved ad-hoc with another function. So here we are... The idea is to have a single function that does the right thing.

Ok, I think I will defer this change to a later pull request if you are
ok with this.

> The renamed functions are fine, though we need to add an alias there so that the old names work. The old names are public, so we must assume somebody is using them.

I added the aliases!

> I forgot to comment on this. I think the `denote-org-capture-with-prompts` function has its use case to produce separate templates. An analogue of what we have with `denote-type`, `denote-signature`, etc.

I brought back `denote-org-capture-with-prompts` and it should work as
before. However, I implemented it by just calling the new
`denote-org-capture` and let-binding `denote-prompts`. I think a user
should still prefer the more flexible `denote-org-capture` because the
order of the prompts can be controlled, unlike
`denote-org-capture-with-prompts`, as seen in the following example.

```
(defun denote-org-capture-signature-keywords ()
  (let ((denote-prompts '(signature keywords)))
    (denote-org-capture)))
```

Given the simplicity of creating separate templates, do you think we
should keep `denote-org-capture-with-prompts`? I can remove it again if
you don't? :P


Everything has been rebased on top of main and is ready to be merged!

protesilaos:
Thank you @jeanphilippegg! I am merging this now.

> Given the simplicity of creating separate templates, do you think we
should keep denote-org-capture-with-prompts? I can remove it again if
you don't? :P

You are right. Let's keep it for a while longer while I prepare the relevant documentation changes.

About `denote-sluggify`, do as you see fit.

PR #205, opened 2023-12-16
Fix for denote-rename-file when signature has ='s for spaces.

wlharvey4:
When a signature is created and spaces are used, the spaces are converted into `=`s in the function `denote--slug-put-equals`.

When a denote file with a signature is renamed, the signature is presented to the user during the renaming process with all of the `=`s intact.  If the user then accepts the signaturre as-is or changes other parts of the signature but leaves the `=`s in place, the `=`s are then removed by the function `denote--slug-no-punct`, which uses the constant regular expression `denote-excluded-punctuation-regexp`, which includes the `=`.  The resulting new filename is thus missing all `=`s that were in the original, although the user intended them to stay (by not removing them).

This fix simply replaces all `=`s with ` `s (a single space) prior to presenting the signature to the user during the renaming process. The user can then accept the signature as-is, or change any part of it, and the spaces will subsequently be converted into `=`s by `denote--slug-put-equals` and will not be lost.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -2496,6 +2496,7 @@ rename non-note files (e.g. PDF) that can benefit from Denote's
 file-naming scheme."
   (interactive
    (let* ((file (denote--rename-dired-file-or-prompt))
+          (spaced-signature (string-replace "=" " " (denote-retrieve-filename-signature file)))
           (file-type (denote-filetype-heuristics file))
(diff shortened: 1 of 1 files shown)


protesilaos:
Hello @wlharvey4 and thank you for this!

Can you please rebase your changes because there are some conflicts right now.

PR #207, opened 2023-12-22
Fix to signature losing "="s when file is renamed.

wlharvey4:
Four file renaming commands may lose spaces (denoted by the "=" in existing filenames) when the user renames the file:

1. denote-rename-file
2. denote-dired-rename-files
3. denote-dired-rename-marked-files-with-keywords
4. denote-rename-file-using-front-matter

This is caused by the function `denote-sluggify-signature`, which calls (indirectly) `denote--slug-no-punct`, which removes, among many other characters, all "="s, which represent spaces.  These spaces are thus lost even if the user does not change the signature.

To remedy this situation, all "="s are turned into spaces before being sent to `denote--slug-no-punct`, as spaces are not removed by this function.  The function `denote--slug-put-equals` turns these spaces into "="s at the end of the file-renaming operation.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -2507,7 +2507,9 @@ file-naming scheme."
   (interactive
    (let* ((file (denote--rename-dired-file-or-prompt))
           (file-type (denote-filetype-heuristics file))
-          (file-in-prompt (propertize (file-relative-name file) 'face 'denote-faces-prompt-current-name)))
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged. Thank you!

protesilaos:
Oh, and a Happy New Year!

PR #208, opened 2023-12-23
Fix `denote-rename-file` when called with empty inputs

jeanphilippegg:
[Ready to be merged.]

Hello Prot!

Recently, you have made a change in `denote-rename-file` to have the
keywords prompt displays the existing keywords. This is good!

There are still some inconsistencies when we explicitely choose to input
nothing (empty string).

For example, given the note
"20231223T111111==aa--test__keyword1_keyword2.txt", if I input nothing
for all prompts, the title and signature are removed, but the keywords
stay. Also, when called from lisp, passing nil or empty string to title
(or signature) yield different results, contrary to what the
documentation says.

I think that now that we display the default value for all prompts
(including keywords), we can obey the prompts (and remove the component
from the file if the input is empty).

This pull request makes `denote-rename-file` more consistent. I tested it
and it does not seem to add any regressions. In particular, the recent
commit e2ba6c1 is still fixed.

PR #210, opened 2023-12-25
Slug functions

jeanphilippegg:
[Wait after pull request #208 is merged.]

Hello Prot!

Do not merge this pull request yet. I will rebase it on main when the
other one is merged.

Sorry about the recent (big) pull requests I have made recently. Please
take all the time you need and don't feel pressured.

--------

In this pull request, I have polished the denote-sluggify-* functions,
as well as the `denote-letter-case' function and related. I have made
them work in a more general and unified way. This allows more
flexibility to the file names.

I have replaced `denote-file-name-letter-casing` with
`denote-file-name-slug-functions`. The idea is the same, but the user
can define any function instead of just verbatim and downcase. The
default behavior does not change. Note that I have made it a `defvar`
just because I do not know how to make it a `defcustom`.

All sluggification goes through the unified `denote-sluggify` function.

I have integrated the sluggification in `denote-format-file-name`. I
think it makes sense that this function does the sluggification. It also
cleans up the code in many places. The sluggification of keywords
was already done in `denote-format-file-name`. Also, the expected type
of the `title` and `signature` parameters is a string. I checked and
none of our calls to `denote-format-file-name` seem to pass a nil value,
which is good. I think it would have been a mistake otherwise and we
would have had to modify them.

I have modified the file name regexps, so that any character is allowed
in the file name. The dired buffers should still work as before. This
makes the file name as flexible as the user needs it to be. We only need
"--", "++" and "==" to find the components. There is no other
restrictions, but I still kept the current behavior as a default.

---------

Notes:

- I have seen pull request #207. You can merge it, but it is also fixed
  differently as part of the current changes.

- The code of this pull request is ready to be merged, but there is
  still the section "Control the letter casing of file names" in the
  manual that needs to be updated.

Have a good day!

### DIFF
--- a/README.org
+++ b/README.org
@@ -1081,25 +1081,17 @@ Add =TITLE= to =FILE=. In interactive use, prompt for user input and
 retrieve the default =TITLE= value from a line starting with a title
 field in the file's contents, depending on the given file type ([[#h:13218826-56a5-482a-9b91-5b6de4f14261][Front matter]]).
 Else, use the file name as a default value at the minibuffer prompt.
-When called from Lisp, =TITLE= is a string.
-
-If =TITLE= is nil or an empty string, do not add it to a newly renamed
-file or remove it from an existing file.
+=TITLE= is a string.
 
 Add =SIGNATURE= to =FILE=. In interactive use, prompt for =SIGNATURE=,
 using an existing one as the default value at the minibuffer prompt.
-When called from Lisp, =SIGNATURE= is a string.
-
-If =SIGNATURE= is nil or an empty string, do not add it to a newly
-renamed file or remove it from an existing file.
+=SIGNATURE= is a string.
 
 Add =KEYWORDS= to =FILE=. In interactive use, prompt for =KEYWORDS=.
 More than one keyword can be inserted when separated by the
-`crm-sepator' (normally a comma). When called from Lisp, =KEYWORDS= is
-a list of strings.
-
-If =KEYWORDS= is nil or an empty string, do not add it to a newly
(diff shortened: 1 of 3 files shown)


jeanphilippegg:
I have added some commits. I had an issue with the file-name regexps now
that I intended to allow any character in it, but I have just fixed it.

In principle, the file naming scheme is very simple and it is not
necessary to impose any restriction on what is allowed in it. The only
rule is that ==,-- and __ tokens are interpreted in a certain way by
Denote and the dot is not allowed either. Any character permitted by the
file system is permitted.

The only challenge was in the fontification of dired buffers because it
requires regexps, but I just fixed it in my last commits.

I also arranged the manual sections on the file naming scheme and the
sluggification. This pull request will be ready.

protesilaos:
Hello @jeanphilippegg! I just merged [pull request 207](https://github.com/protesilaos/denote/pull/207). Please let me know how you want to proceed.

I will now take a look at your other PRs. I have been trying to minimise my keyboard acivity because my left forearm is in pain again.

Happy New Year, by the way!

jeanphilippegg:
> Hello @jeanphilippegg! I just merged [pull request 207](https://github.com/protesilaos/denote/pull/207). Please let me know how you want to proceed.
> 
> I will now take a look at your other PRs. I have been trying to minimise my keyboard acivity because my left forearm is in pain again.
> 
> Happy New Year, by the way!

I have replied in pull request #208.

Take care and Happy New Year!

protesilaos:
Merged. Thank you!

PR #211, opened 2023-12-28
Remove denote-user-enforced-denote-directory

jeanphilippegg:
[Ready to be merged.]

Ignore this pull request if I have missed some use-case.

I have looked in the manual because I wanted to understand
`denote-user-enforced-denote-directory`. It seems like users are meant
to let-bind it to enforce a specific value for `denote-directory`.
However, after some tests, let-binding `denote-directory` directly also
works.

I have removed this user-option in this pull request and adapted the
manual accordingly.


protesilaos:
Merged. Thank you!

PR #213, opened 2023-12-28
Do not use call-interactively-p and make renaming commands obey denote-sort-keywords

jeanphilippegg:
[Wait after pull request #208 is merged.]

I have moved the sorting of keywords outside `denote-keywords-prompt`
for these reasons:

- While investigating the error fixed by commit e2ba6c1, I discovered
  that if we set a transient for the `denote` command and call it
  through the transient, then we get the interactive behavior (prompts),
  but `call-interactively-p` seems to return nil anyway. This makes me
  think that we should try to avoid this function. Fortunately, it was
  not necessary to keep it in our case.

- When calling `denote-rename-file` from lisp, the keywords are not
  sorted. This is not coherent with the `denote` command, which always
  sorts keywords. This is fixed. `denote-add-front-matter` is also
  fixed.

- Double sorting was happening in the case of `denote-region`. Not an
  issue, but this fixes it.

### DIFF
--- a/README.org
+++ b/README.org
@@ -1081,25 +1081,17 @@ Add =TITLE= to =FILE=. In interactive use, prompt for user input and
 retrieve the default =TITLE= value from a line starting with a title
 field in the file's contents, depending on the given file type ([[#h:13218826-56a5-482a-9b91-5b6de4f14261][Front matter]]).
 Else, use the file name as a default value at the minibuffer prompt.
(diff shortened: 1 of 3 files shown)


protesilaos:
Just to note that I have seen this. Let's finish [pull request 208](https://github.com/protesilaos/denote/pull/208) first.

PR #217, opened 2024-01-03
Make denote-region obey denote-prompts

jeanphilippegg:
Hello Prot!

Even if there are still a few pull requests pending, I am moving forward
with the ideas I have for Denote. Don't feel pressured. The diff of this
PR includes the previous ones (so this one looks big), but it is
actually a small one. I decided to keep them sequential in terms of
commits so that they are easy for you to merge in order when you are
ready. I can always rearrange them if you prefer them in another order.
You can check the individual commits.

-------

I was able to make `denote-region` obey `denote-prompts`. The issue I
had was that both `denote-region` and `denote` tried to use the region
internally, so there was some conflict when I tried to call `denote`
from `denote-region`.

The solution was simple. I just added the variable
`denote--use-region-in-denote-command`. It allows to temporarily
disable the region handling of the `denote` command.

Have a good day and take care of your forearm!


protesilaos:
Merged. Thank you!

Issue #218, opened 2024-01-06
In silo, reference to variable "default-directory" doesn't work, jamming all denote-related commands for note creation etc.


hapst3r:
Hello @protesilaos 

after a recent update to denote (like ~2 weeks ago or so), whenever I issue a `denote`-related command in a silo -- such as `denote-keywords-add`, `denote-link-or-create` or `denote` --, the command is not executed, with this error message in the `*Messages*` buffer: denote--directory-all-files-recursively: Wrong type argument: stringp, default-directory.

Before this update -- currently I am on version 2.2.4, commit dfb517a8 --, those commands would work just fine, creating new notes in said silo, adding keywords to current note, etc.

This is the relevant part of my `denote` config:
```
(use-package denote
  :demand t
  :custom
  (denote-directory "~/notes/denote/")
  (denote-known-keywords nil)
  (denote-sort-keywords nil)
  (denote-dired-directories (cdr (ffap-all-subdirs denote-directory))))
```

The directory structure is ~/notes/denote/SILOS, each $SILO containing a `.dir-locals.el` with the content suggested in the manual: `((nil . ((denote-directory . default-directory))))`.

Executing `helpful-variable` on `denote-directory` from within a $SILO directory, the value is `default-directory`, with the original value `/home/USER/Documents/notes`. The value of `default-directory` correctly points to the $SILO in which I reside issuing the command.

If I can provide further information, please tell me which (and how to obtain them).

Have a good day :)

jeanphilippegg:
Hello hapst3r,

Sorry about this issue. This feature has been broken by my pull request
(#211). I had overlooked this way of setting `denote-directory` (in
`.dir-locals.el`).

I have reverted my changes in a new pull request.

hapst3r:
Hej @jeanphilippegg 

thanks for the quick response and fix. I hope this gets merged soon so I can use denote the way I usually do. Yet, it is good to know that it is not related to something strange that is hard to fix or a problem on my part.

jeanphilippegg:
No worry!

Until this is merged, you can use
```
((nil . ((denote-directory . "/path/to/your/silo/"))))
```
in your `.dir-locals.el` and accept the warning that the value may not be safe.

protesilaos:
Hello @hapst3r! I just merged it. Please check if everything is okay there.

hapst3r:
> Hello @hapst3r! I just merged it. Please check if everything is okay there.

[...]

PR #219, opened 2024-01-07
Revert changes about denote-user-enforced-denote-directory

jeanphilippegg:
As reported in #218, removing the user option
`denote-user-enforced-denote-directory` has created a regression. I did
not know that we could set `denote-directory` to `default-directory` in
this way. I now understand why this user option was added in the first
place.

I have done some testing and I am having issues with both the old and
the new code. So I just reverted commits 5e9b6b3 and 49a0608 for now.

I will do more experimentation and might come up with something later.

Sorry about that!


protesilaos:
That's fine. Thank you for doing this! Merged.

PR #222, opened 2024-01-14
New variable `denote-link-description-function` and some bug fixes

jeanphilippegg:
I have added the user option `denote-link-description-function` (as a
defvar for now) which defaults to display the signature (if present) and
the title. I think it is a good default, but we can change it. It is now
easy to configure the link descriptions to look like:
  - "Link"
  - "20240101T111111--My-note.org"
  - "My file title"
  - "My file title (keyword1, keyword2)"
  - "sig1z3   My other title"  (the default if signature is present)
  - "My other title (sig1z3)"
  - A prompt for the description
  - etc.

`denote-link-description-function` can be `let` bound if we want to fix
the description format in specific functions, but we should avoid doing
that and let the user decide most of the time. The user can also let
bind this variable if he wants to change the behavior for a specific
command.

Some comments:

- `denote-link-after-creating-with-command` previously displayed the
  signature in the description if it were specifically called with a
  command named `denote-signature`. This was not robust because a user
  can create any command that is not named `denote-signature` that still
  includes the signature prompt. In this case, he would not get the
  signature in the link's description. I have made it obey
  `denote-link-description-function` instead. The current default will
  display the signature if there is one, so the behavior is not changed.
  It can now be configured.
- Similarly, `denote-add-links`, `denote-org-dblock--get-file-contents`
  and `denote-link-with-signature` obey
  `denote-link-description-function`. With the default value, it is the
  same behavior as before.

Bugs fixed:

  - This bug has been reported on the mailing list for
    `denote-link-or-create`, but any command that made use of the
    `denote--link-after-creating-subr` (all commands except
    `denote-link`) used the file-type of the linked file instead of that
    of the current file. This is fixed.
  - `denote-link-or-create`, when called with an active region and the
    file does not exist yet, would use the active region as the initial
    title prompt (in addition to the default prompt being what was typed
    in the `denote-file-prompt`). This looks strange, so I have fixed it
    using `denote--command-with-features`. The region should only be
    used as the description text here and not the title prompt.

The rest is just cleanup and refactoring. It is ready to be merged.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -3139,88 +3139,117 @@ The format of such links is `denote-md-link-format'.")
   "Regexp to match an identifier-only link in its context.
 The format of such links is `denote-id-only-link-format'."  )
 
-(defun denote-link--file-type-format (file-type id-only)
-  "Return link format based on FILE-TYPE.
-With non-nil ID-ONLY, use the generic link format without a
-title.
+(defun denote-format-link (file description file-type id-only)
+  "Prepare link to FILE using DESCRIPTION.
 
-Fall back to `denote-org-link-format'."
-  (cond
-   (id-only denote-id-only-link-format)
-   ((when-let ((link (denote--link-format file-type)))
-      link))
-   ;; Plain text also uses [[denote:ID][TITLE]]
-   (t denote-org-link-format)))
-
-(defun denote-format-link (file format description)
-  "Prepare link to FILE using FORMAT and DESCRIPTION text.
-FILE is the path to a file name.  FORMAT is the symbol of a
-variable that specifies a string.  See the `:link' property of
-`denote-file-types'.
-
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged. Thank you!

brabalan:
Thanks a lot for fixing the bug I reported on the mailing list. I was about to report it here as well and I saw this fixed it.

protesilaos:
> From: Alan Schmitt ***@***.***>
> Date: Sun, 21 Jan 2024 05:35:57 -0800
>
> Thanks a lot for fixing the bug I reported on the mailing list. I was about to report it here as well and I saw this fixed it.

You are welcome!

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #225, opened 2024-01-20
Enforce other rules in the file name

jeanphilippegg:
Just a small pull request to remove dot characters automatically in file
name components, even if this is not done by
`denote-file-name-slug-functions`. Also =, - and _ should be removed at
the **end** of the file name component.

I think file names are robust now.


protesilaos:
Merged. Thank you!

PR #226, opened 2024-01-20
Fix denote-link-description-with-signature-and-title

jeanphilippegg:
In the docstring of `denote-link`, it was said that an empty region
leads to links of type `id-only`. I forgot to make this work also in the
new code. This commit fixes that. Note that this works because the
default `denote-link-description-with-signature-and-title` decides to
return an empty string when given an empty region. This can behave
differently if the option is set to another function, which is ok.


protesilaos:
Merged. Thank you!

Issue #230, opened 2024-01-26
A denote tag 'Identifier' is different from the id in the filename occasionally


seokbeomKim:
Hi, 

I am curious about the identifier in the denote file.
Sometimes, the value of #+identifier seems to be different from the actual id from the filename. 

Here's the example:

```
#+title:      Meeting to investigate SERDES
#+date:       [2024-01-24 Wed 14:54]
#+filetags:   :cs:des:ser:
#+identifier: **20240124T145433**
```

And the actual file name is:

$denote/meeting/**20240124T145432**--meeting-to-investiage-serdes__cs_des_ser.org

I think this is not the correct behavior...; does it need to use 'second' in the indentifier string?

jeanphilippegg:
This can happen because a `nil` date is passed to the `denote--prepare`
internal function and we do not use the same value for the identifier in
the filename and in the front matter. This is possible since the
addition of `denote-parse-date` which returns `nil` if the `date` passed
is `nil` (because `nil` satisfies `listp`). This was likely unintended.

This is fixed as part of pull request #231.

It should not have broken anything though, as the identifier in the
front matter is not used by Denote.

Issue #224, opened 2024-01-18
Denote link with direct jump to an Org heading (plus `denote-org-extras.el`)


protesilaos:
I have some working code in the `link-to-org-headings` branch and I am interested in further comments on whether this is useful, how it can be expanded, etc.

Excerpt from the manual on what is on offer right now:

```org
[ Part of {{{development-version}}}. ]

The user option ~denote-org-store-link-to-heading~ determines whether
~org-store-link~ links to the current Org heading.

When its value is non-nil, ~org-store-link~ stores a link to the
current Org heading inside the Denote file. If the heading does not
have a =CUSTOM_ID=, it creates it and includes it in the heading's
=PROPERTIES= drawer. If a =CUSTOM_ID= exists, ~org-store-link~ use it
as-is.

This make the resulting link a combination of the =denote:= link type,
pointing to the identifier of the current file, plus the value of the
heading's =CUSTOM_ID=, such as:

- =[[denote:20240118T060608][Some test]]=
- =[[denote:20240118T060608::#h:eed0fb8e-4cc7-478f-acb6-f0aa1a8bffcd][Some test]]=

Both lead to the same Denote file, but the latter jumps to the heading
with the given =CUSTOM_ID=.

The value of the =CUSTOM_ID= is determined by the Org user option
~org-id-method~. The sample shown above uses the default UUID
infrastructure.

If ~denote-org-store-link-to-heading~ is set to a nil value, the
command ~org-store-link~ only stores links to the Denote file (using
its identifier), but not to the given heading. This is what Denote was
doing in versions prior to =3.0.0=.

[ This feature only works in Org mode files, as other file types do
  not have a linking mechanism that handles unique identifiers for
  headings or other patterns to jump to. If ~org-store-link~ is
  invoked in one such file, it captures only the Denote identifier of
  the file, even if this user option is set to a non-nil value. ]"
```

jeanphilippegg:
I have tested this and it works great! As I was not using this feature before, I don't know how it can be expanded. I think this is useful!

protesilaos:
Thank you @jeanphilippegg! I just merged it into `main`. Check the command `denote-org-link-to-heading` for an idea (part of `denote-org.el` for which I have several ideas).

The `org-store-link` now also captures the text of the heading when generating the description of the link.

jeanphilippegg:
Ok, I will check this tomorrow!

jeanphilippegg:
[...]

PR #231, opened 2024-01-27
Various fixes

jeanphilippegg:
Fixes:

- Fix `denote-org-extras-dblock--get-file-contents`, as reported in
  #224.

- Make `denote--keyword-history` an alias of `denote-keyword-history.

- Fix issue #230.

- Update some docstrings and README.org about link descriptions.


protesilaos:
Merged. Thank you!

PR #237, opened 2024-01-28
Cleanup and bug fix

jeanphilippegg:
Remove `denote--command-with-default-title` and use `denote--command-with-features` instead.

Also, there is a bug in `denote-find-link` when called from a note in a subdirectory.


protesilaos:
Merged. Thank you!

PR #238, opened 2024-01-28
Denote relative file prompt

jeanphilippegg:
Please review the section in the manual "Use the tree-based file prompt for select commands" as this pull request breaks it and I do not know how we can support this anymore.

In this pull request, I have replaced `denote-file-prompt` with `denote-relative-file-prompt`. It should affect all functions that make use of it. I checked and it seems to always make sense.

Note that I did not alias `denote-file-history` to `denote-relative-file-history` as they would not contain the same values anyway.


jeanphilippegg:
The other pull request (#237) is ready to be merged before this one. This one can be merged once we decide what to do about the manual section above.

jeanphilippegg:
I am closing this as I think I have a simpler way to solve this.

PR #239, opened 2024-01-28
Denote file prompt

jeanphilippegg:
Do not display `denote-directory` in the candidates of `denote-file-prompt`.

This affects all functions that use `denote-file-prompt`, but it seems to make sense.

Note that the value of `denote-file-history` should now contain the relative paths. I think it is not an issue that it may still contain previous full paths. They will be forgotten at some point according to the settings of `savehist-mode`.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1093,8 +1093,11 @@ file in the returned list."
   "Prompt for file with identifier in variable `denote-directory'.
 With optional FILES-MATCHING-REGEXP, filter the candidates per
 the given regular expression."
-  (let ((files (denote-directory-files files-matching-regexp :omit-current)))
-    (completing-read "Select note: " files nil nil nil 'denote-file-history)))
+  (let* ((files (denote-directory-files files-matching-regexp :omit-current))
+         (file-names (mapcar (lambda (f)
+                               (propertize f 'display (denote-get-file-name-relative-to-denote-directory f)))
+                            files)))
+    (completing-read "Select note: " file-names nil nil nil 'denote-file-history)))
 
 ;;;; Keywords
 


jeanphilippegg:
This is ready to be merged after #237.

protesilaos:
In principle, I am fine with this. Though I wonder if there is a way to achieve this only visually. This way we do not have to modify paths back and forth. I believe it is possible to use display properties for this, though I have not researched the matter.

jeanphilippegg:
I think I have made it work with `propertize`! It would be ready to be merged.

protesilaos:
I like it! But we need to clear the faces. On my end, using the generic completions UI, I see everything with the `completions-common-part` face. With `vertico-mode` I get a similar result. Because of this completion faces are no longer shown.

jeanphilippegg:
I see that the face is different. Do you know how this can be done (clearing the faces)? I don't understand why they are applied or how to clear them.

protesilaos:
I am not sure what we are missing. I get the feeling it is something obvious. Could it be that we need to add a space before so that Emacs does not think we have matched the candidates by virtue of matching their common component?

mholson:
I have added this commit function update `denote-file-prompt` to my config and it makes adding links feasible again.  It was just too difficult to read the file name with a long path prepending the file name.  

Is this something new where the path is visible?  From what I can remember and have seen in the denote videos posted, when inserting a link **ONLY** the file name was visible.

protesilaos:
Hello @mholson! In the past we were using the prompt provided by the built-in project.el. It has the nice feature of hiding the common part of file names. Though it causes us other problems, such as not reading our minibuffer history. We will eventually get this back, because I think it looks better. It's just a matter of figuring out the technical details.

jeanphilippegg:
I have tried to combine `'display` with `'face` to force a face over `completions-common-part`, but it did not work. I don't know what we can do from here.

protesilaos:
> Date: Fri, 16 Feb 2024 17:20:22 -0800
>
> I have tried to combine `'display` with `'face` to force a face over
> `completions-common-part`, but it did not work. I don't know what we
> can do from here.

I also tried various combinations of propertize. Nothing seems to work.

In that case, we will do the change to the actual file paths.

-- 
Protesilaos Stavrou
https://protesilaos.com

jeanphilippegg:
I was thinking that it was nice to preserve the full path, since if we support multiple denote directories (discussed in another issue) and we might need it. But this is not the only thing to consider anyway to support multiple denote directories. We can check later.

protesilaos:
> Date: Fri, 16 Feb 2024 20:36:53 -0800
>
> I was thinking that it was nice to preserve the full path, since if we
> support multiple denote directories (discussed in another issue) and
> we might need it. But this is not the only thing to consider anyway to
> support multiple denote directories. We can check later.

Yes, this is my concern with relative file paths. We might run into
trouble.

Maybe we can ask on emacs-devel if the propertize we are doing here can
actually work.

If this were a buffer, we could put an invisible property (like how when
you fold an Org heading), but I think in the minibuffer things are different.

-- 
Protesilaos Stavrou
https://protesilaos.com

protesilaos:
Just to add though that we can still return a full path if we want (and store that in the history). It is a matter of how best to approach this.

jeanphilippegg:
I don't know how to do that and still display the relative path. You can ignore this pull request and make the change.

protesilaos:
> I don't know how to do that and still display the relative path.

The idea is to do the `mapcar` that returns the relative paths and pass that to `completing-read`. Then when we get the return value, we expand the file name against the `denote-directory` and do an `add-to-history`.

Does this make sense?

[ I know it is a roundabout way, of the sort I wanted to avoid with the use of display properties but here we are... The `project--read-file-cpd-relative` that we were using before is doing something similar. ]

> You can ignore this pull request and make the change.

I am happy to do this. I thought I would mention the idea here just to be sure we are on the same page (and since you made the initiative with the PR).

jeanphilippegg:
> Does this make sense?

I have not tested it, but if it works, it is good with me!

> I am happy to do this. I thought I would mention the idea here just to be sure we are on the same page (and since you made the initiative with the PR).

No worry! I will check how it works.

protesilaos:
I am closing this and will instead use a variant of what we had before with the project.el prompt that shows relative files. The work-in-progress code:

```elisp
;; NOTE 2024-02-29 08:09:42 +0200: Based on
;; `project--read-file-cpd-relative' from the built-in project.el
(defun denote-file-prompt (&optional files-matching-regexp)
  "Prompt for file with identifier in variable `denote-directory'.
With optional FILES-MATCHING-REGEXP, filter the candidates per
the given regular expression."
  (let* ((all-files (denote-directory-files files-matching-regexp :omit-current))
         (common-parent-directory
          (let ((common-prefix (try-completion "" all-files)))
            (if (> (length common-prefix) 0)
                (file-name-directory common-prefix))))
         (cpd-length (length common-parent-directory))
         (prompt (if (zerop cpd-length)
                     "Select note: "
                   (format "Select note in %s: " common-parent-directory)))
         (included-cpd (when (member common-parent-directory all-files)
                         (setq all-files
                               (delete common-parent-directory all-files))
                         t))
         (substrings (mapcar (lambda (s) (substring s cpd-length)) all-files))
         (_ (when included-cpd
              (setq substrings (cons "./" substrings))))
         (new-collection (project--file-completion-table substrings))
         (abs-cpd (expand-file-name common-parent-directory))
         (abs-cpd-length (length abs-cpd))
         (relname (cl-letf* ((non-essential t) ;Avoid new Tramp connections.
                             ((symbol-value 'denote-file-history)
                              (mapcan
                               (lambda (s)
                                 (setq s (expand-file-name s))
                                 (and (string-prefix-p abs-cpd s)
                                      (not (eq abs-cpd-length (length s)))
                                      (list (substring s abs-cpd-length))))
                               (symbol-value 'denote-file-history))))
                    (completing-read prompt new-collection nil nil nil 'denote-file-history)))
         (absname (expand-file-name relname common-parent-directory)))
    (add-to-history 'denote-file-history absname)
    absname))
```

And diff:

```diff
 denote.el | 36 ++++++++++++++++++++++++++++++++++--
 1 file changed, 34 insertions(+), 2 deletions(-)

diff --git a/denote.el b/denote.el
index fff4960..0fd55db 100644
--- a/denote.el
+++ b/denote.el
@@ -1113,12 +1113,44 @@ (defvar denote-file-history nil
 (defalias 'denote--file-history 'denote-file-history
   "Compatibility alias for `denote-file-history'.")
 
+;; NOTE 2024-02-29 08:09:42 +0200: Based on
+;; `project--read-file-cpd-relative' from the built-in project.el
 (defun denote-file-prompt (&optional files-matching-regexp)
   "Prompt for file with identifier in variable `denote-directory'.
 With optional FILES-MATCHING-REGEXP, filter the candidates per
 the given regular expression."
-  (let ((files (denote-directory-files files-matching-regexp :omit-current)))
-    (completing-read "Select note: " files nil nil nil 'denote-file-history)))
+  (let* ((all-files (denote-directory-files files-matching-regexp :omit-current))
+         (common-parent-directory
+          (let ((common-prefix (try-completion "" all-files)))
+            (if (> (length common-prefix) 0)
+                (file-name-directory common-prefix))))
+         (cpd-length (length common-parent-directory))
+         (prompt (if (zerop cpd-length)
+                     "Select note: "
+                   (format "Select note in %s: " common-parent-directory)))
+         (included-cpd (when (member common-parent-directory all-files)
+                         (setq all-files
+                               (delete common-parent-directory all-files))
+                         t))
+         (substrings (mapcar (lambda (s) (substring s cpd-length)) all-files))
+         (_ (when included-cpd
+              (setq substrings (cons "./" substrings))))
+         (new-collection (project--file-completion-table substrings))
+         (abs-cpd (expand-file-name common-parent-directory))
+         (abs-cpd-length (length abs-cpd))
+         (relname (cl-letf* ((non-essential t) ;Avoid new Tramp connections.
+                             ((symbol-value 'denote-file-history)
+                              (mapcan
+                               (lambda (s)
+                                 (setq s (expand-file-name s))
+                                 (and (string-prefix-p abs-cpd s)
+                                      (not (eq abs-cpd-length (length s)))
+                                      (list (substring s abs-cpd-length))))
+                               (symbol-value 'denote-file-history))))
+                    (completing-read prompt new-collection nil nil nil 'denote-file-history)))
+         (absname (expand-file-name relname common-parent-directory)))
+    (add-to-history 'denote-file-history absname)
+    absname))
 
 ;;;; Keywords
 
```

protesilaos:
I made some follow-up commits to simplify/refine the `denote-title-prompt`. Please check the source instead.

PR #240, opened 2024-02-01
Fix: Consider the CLOSED property of the org-heading for heading-date

vedang:
This property is automatically added to a task when we mark it as DONE. If both DATE as well as CREATED are not available, this is a valuable property to use when determining the date of the new note.


protesilaos:
Sounds good! How about other kids of timestamp? Maybe if they are in the heading? Or `DEADLINE`, `SCHEDULED`? Do these make sense in this case?

The idea is to make it easier for users to benefit from this facility. Though we don't want to overengineer it here.

protesilaos:
All good. Merging this. Thank you!

PR #241, opened 2024-02-01
Use denote-prompts to prompt user when extracting org subtree

vedang:
This requirement might be specific to my workflow and I would love maintainer feedback about it.

1. denote-subdirectory-prompt: When I am extracting data from my org-files as Denote notes, I _almost never_ want them to end up in the default directory. Instead, I want them to go to different directories based on the kind of content I am extracting.

2. denote-signature-prompt: I use signatures to track trails of thought, and as such I use them for certain kinds of notes. I'd like this data to be converted properly to Denote when I move from subtree-based notes to Denote notes.

### DIFF
--- a/denote-org-extras.el
+++ b/denote-org-extras.el
@@ -161,10 +161,19 @@ Make the new note an Org file regardless of the value of
   (if-let ((text (org-get-entry))
            (heading (denote-link-ol-get-heading)))
       (let ((tags (org-get-tags))
-            (date (denote-org-extras--get-heading-date)))
+            (date (denote-org-extras--get-heading-date))
+            subdirectory
+            signature)
+        (dolist (prompt denote-prompts)
+          (pcase prompt
+            ('keywords (when (not tags)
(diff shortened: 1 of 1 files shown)


protesilaos:
Hello @vedang! What you describe makes sense. Perhaps we should just respect the value of the `denote-prompts` user option?

vedang:
Respecting the value of `denote-prompts` would be perfect!

I couldn't figure out how to do that and prefill the correct data in the slots (heading, tags, date from the org-entry; prompting for other stuff). Is there an example available that I can refer to?

protesilaos:
I can work on this, though it is not easy right now because my typing time is limited. Maybe you can take a look at the `interactive` spec of the `denote` command?

vedang:
Thank you, I'll look at the `interactive` spec and get back with questions,
if any.

vedang:
I've made changes to the code to use `denote-prompts`. Please review and let me know what you think! This is working well for me.

protesilaos:
Merged. Thank you!

Issue #244, opened 2024-02-06
Setting denote-directory in .dir-locals.el doesn't override denote-directory, but is a silo flag


d-qoi:
Hello!

I've found that denote uses denote-directory defined in .dir-locals differently than how it uses it globally. 

It uses any non-nil binding of denote-directory as a flag that the directory the .dir-locals is defined in should be the root of a silo, rather than the value that denote-directory is bound to.

I've got what may not be the most common use case, in which I want to keep notes for each project in a dedicated folder in the project root. I use a .dir-locals.el in the project root for some project specific customization, style settings and such, and as it's a project root. I'd rather not have my project roots cluttered further with my notes, all things considered.

I noticed in the code that when using denote in a directory with a .dir-locals in the root, denote will always use that parent folder as the root directory.
It all seems to be this due to this specific check.
```
(defun denote--default-directory-is-silo-p ()
  "Return path to silo if `default-directory' is a silo."
  (when-let ((dir-locals (dir-locals-find-file default-directory))
             ((alist-get 'denote-directory dir-local-variables-alist)))
    (cond
     ((listp dir-locals)
      (car dir-locals))
     ((stringp dir-locals)
      dir-locals))))
```
The specific value of the locally set denote-directory is grabbed, but never used beyond just a check to see if it's bound at all.
I ended up modifying this function to get the behavior I wanted: the ability to set the denote-directory value explicitly with .dir-locals, and retain the behavior of checking to see if it's bound so denote knows to use the root at a silo. 

```
(defun denote--default-directory-is-silo-p ()
  "Return path to silo if `default-directory' is a silo."
  (when-let ((dir-locals (dir-locals-find-file default-directory))
             (local-val (alist-get 'denote-directory dir-local-variables-alist)))
    (cond
     ((and (stringp local-val)
           (file-directory-p local-val))
      local-val)
      ((listp dir-locals)
       (car dir-locals))
      ((stringp dir-locals)
       dir-locals))))
 ```
 
 I know I could have gotten around this by setting user-enforced-denote-directory in my dir-locals, but that doesn't seem to be what the variable is for, looking at the docs.
 
 The docs also specify that a path should be specified for the denote-directory in some circumstances, subdirs in silos, which may cause confusion for others if they specify a path that is not strictly to the parent directory of the .dir-locals. I not attempted to do that yet.

MirkoHernandez:
[...]

PR #246, opened 2024-02-10
Remove denote-user-enforced-denote-directory

jeanphilippegg:
This is my second attempt at removing this user option. Let binding
`denote-directory` directly should work.

The first time, I overlooked the way of setting `denote-directory` to
`default-directory` or `local` in `.dir-locals.el` and it broke some
user setup. I was doing some tests, but I was experiencing issues when I
tried to put `.dir-locals.el` in subdirectories. Apparently, they do not
stack. Only the closest `.dir-locals.el` is used. This is an issue I had
even with the actual code. It was just a misunderstanding on my part.

Let-binding `denote-directory` should be enough and we don't need the
user option `denote-user-enforced-denote-directory`. However, I think we
really should always set `denote-directory` to a path string, even in
`.dir-locals.el`. I have updated the manual accordingly. However, I
still support these special values to not break current user's config.
We might want to mention in the release notes that it could be removed
eventually.

This old behavior is retained: If `denote-directory` is the symbol
`default-directory` (or `local`), we can assume it was set in a
`.dir-locals.el` as previously specified in the manual.

We can also remove it right away if you want. We don't really need a
shortcut for a setting in `.dir-locals.el` since this is a one-time setup.
The manual already does not mention this possibility anymore. Let me
know if you would like this instead.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -123,6 +123,11 @@
 
 ;; About the autoload: (info "(elisp) File Local Variables")
 
+(define-obsolete-variable-alias
+ 'denote-user-enforced-denote-directory
+ 'denote-directory
+ "3.0.0")
+
 ;;;###autoload (put 'denote-directory 'safe-local-variable (lambda (val) (or (stringp val) (eq val 'local) (eq val 'default-directory))))
 (defcustom denote-directory (expand-file-name "~/Documents/notes/")
   "Directory for storing personal notes.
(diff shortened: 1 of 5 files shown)


denote.el
```diff
@@ -658,25 +644,22 @@ Like `denote--completion-table' but also disable sorting."
   (when (not (file-directory-p denote-directory))
     (make-directory denote-directory :parents)))
 
-(defvar denote-user-enforced-denote-directory nil
```
d-qoi, line 661: Would it be worth marking denote-user-enforced-denote-directory with define-obsolete-variable-alias so people who are using it in their own code have a chance to update it before it's dropped completely?

jeanphilippegg: Done!

protesilaos:
Thank you @jeanphilippegg! @d-qoi do you think this fixes [the issue discussed in 244](https://github.com/protesilaos/denote/issues/244)?

d-qoi:
Yeah! It should work just fine!

protesilaos:
Thank you! I merged it, though I did it by appending `.patch` to this URL and then resolving the conflict to apply the patch. I don't know how to now mark this as "merged"...

PR #248, opened 2024-02-11
Update denote-create-* commands to respect `denote-prompts`

vedang:
This commit modifies the following functions:

- `denote-type`
- `denote-date`
- `denote-subdirectory`
- `denote-template`
- `denote-signature`

Instead of overwriting the value of `denote-prompts`, this change adds the specific symbol to the existing value of `denote-prompts`.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -2249,6 +2249,10 @@ With optional PROMPT-TEXT use it instead of a generic prompt."
 (defalias 'denote-create-note 'denote
   "Alias for `denote' command.")
 
+(defun denote--add-prompts (additional-prompts)
+  "Add all the elements in the ADDITIONAL-PROMPTS list to `denote-prompts'."
+  (seq-union denote-prompts additional-prompts))
+
 ;;;###autoload
 (defun denote-type ()
   "Create note while prompting for a file type.
@@ -2257,7 +2261,7 @@ This is the equivalent to calling `denote' when `denote-prompts'
 is set to \\='(file-type title keywords)."
   (declare (interactive-only t))
   (interactive)
-  (let ((denote-prompts '(file-type title keywords)))
+  (let ((denote-prompts (denote--add-prompts '(file-type))))
     (call-interactively #'denote)))
 
 (defalias 'denote-create-note-using-type 'denote-type
@@ -2276,7 +2280,7 @@ This is the equivalent to calling `denote' when `denote-prompts'
 is set to \\='(date title keywords)."
   (declare (interactive-only t))
   (interactive)
-  (let ((denote-prompts '(date title keywords)))
+  (let ((denote-prompts (denote--add-prompts '(date))))
     (call-interactively #'denote)))
 
 (defalias 'denote-create-note-using-date 'denote-date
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you @vedang! This is better!

About the new function, does it do something more than this?

```elisp
(defun denote--add-prompts (additional-prompts)
  "Add all the elements in the ADDITIONAL-PROMPTS list to `denote-prompts'."
  (append additional-prompts denote-prompts))
```

If not, I suggest we avoid the `cl-*` functions because we do not require the `cl-lib` right now. I have no problem adding it, but if we can avoid it then let's do it.

maxbrieiev:
> ```emacs-lisp
> (defun denote--add-prompts (additional-prompts)
>   "Add all the elements in the ADDITIONAL-PROMPTS list to `denote-prompts'."
>   (append additional-prompts denote-prompts))
> ```

I think with this code there is a chance to end up with duplicated prompts:

```emacs-lisp
(let ((denote-prompts '(date title keywords)))
  (append '(title) denote-prompts))
;; => (title date title keywords)
```

I believe this is why @vedang used `cl-adjoin`.

But I think what we need is just a union on two sets:

```emacs-lisp
(let ((denote-prompts '(date title keywords)))
  (seq-union '(title) denote-prompts))
;; => (title date keywords)
```

protesilaos:
Thank you @maxbrieiev! Okay, this clear.

vedang:
yes, @maxbrieiev is exactly right. I did't know about `seq-union` :) I'll change the code to use `seq-union` instead and push the update here.

vedang:
One potential problem with using `seq-union` is that the order in which the user is prompted changes. I don't think that it's an issue, but I wanted to document it none-the-less:

```
ELISP> (seq-union '(title) '(date keyword))
(title date keyword)

ELISP> (seq-union '(title) '(date title keyword))
(title date keyword)

ELISP> (cl-reduce (lambda (prompts prompt) (cl-adjoin prompt prompts))
                  '(title)
                  :initial-value '(date title keyword))
(date title keyword)

ELISP> (cl-reduce (lambda (prompts prompt) (cl-adjoin prompt prompts))
                  '(title signature)
                  :initial-value '(date title keyword))
(signature date title keyword)
```

protesilaos:
What about this?

```elisp
(delete-dups (append '(title) '(date title keyword)))
;; => (title date keyword)
```

I think the desired behaviour is to have the new prompt first at all times. Or maybe we want something else?

vedang:
Re: new prompts are asked first all the time:

My thinking was: "your pre-configured prompts are always asked first, and then any new prompts are asked", making the experience of new notes consistent across all of Denote. 

However, the existing user experience i "new prompts are always asked first". Changing it might be jarring to end-users. 

Let me know your decision on this, and I'll update the commit accordingly.

protesilaos:
Merged, thank you!

> Let me know your decision on this, and I'll update the commit accordingly.

Let's keep the current approach. Though we can have a user option if there is demand for it. Do you want it?

protesilaos:
Oh, I just realised this was not against `main`... Nevermind, we end up with the same result.

PR #249, opened 2024-02-11
Fix typo in variable name in documentation

maxbrieiev:
- Corrected variable name from `user-enforced-denote-directory` to `denote-user-enforced-denote-directory`.

- Removed the paragraph that referred to its usage in `denote-org-extras-extract-org-subtree` function. Apparently, this function does not use `denote-user-enforced-denote-directory` variable.


protesilaos:
Merged. Thank you!

Issue #251, opened 2024-02-12
Kill "hidden" buffer generated by `denote-link-return-links`?


mattyonweb:
Hi! 

I'm developing a small extension to Denote that converts notes into a DOT graph. In order to build such graph, I need to invoke `denote-link-return-links` on every note in `denote-directory`, but I noticed that, after doing so, a buffer remains for every visited note, which is a bit annoying.

This is due to the last two lines of `denote-link-return-links`:
```
(defun denote-link-return-links (&optional file)
  ;; ...
  (with-current-buffer (find-file-noselect current-file)
      (denote-link--expand-identifiers regexp))))
```
The `with-current-buffer` does not "clean up" the opened buffers, and hence the buffer-list is polluted by potentially hundreds of useless buffers.

This problem could be avoided by e.g. exchanging the two lines above with the following. Performance-wise, on my small set of notes (~250) the time needed to run the following version of `denote-link-return-links` on all notes is, surprisingly, ~40% of the original version:

```
(defun denote-link-return-links (&optional file)
  ;; ...
    (with-temp-buffer
      (insert-file-contents current-file)
      (denote-link--expand-identifiers regexp))))
```

However, I'm not sure if this change may have any counterindication, or if, perhaps, the buffer being kept open was a deliberated choice.

protesilaos:
Hello @mattyonweb!

I believe I tried the `with-temp-buffer` before, but I do not recall why I opted it against it... Looking at it now, I must have been wrong. What you have here is fine.

We use the `(with-current-buffer (find-file-noselect ...))` in a few more places, though I think those are correct uses of it.

Do you want to send a patch or pull request?

PR #252, opened 2024-02-14
Kill "hidden" buffer generated by `denote-link-return-links`

mattyonweb:
Hi! As suggested in #251, I implemented the change to `denote-link-return-links`. 

With this change, `denote-link-return-links` will avoid keeping visited buffers open.


protesilaos:
Merged. Thank you!

PR #256, opened 2024-02-16
add function denote-org-capture-with-defaults

kimothy:
Hi @protesilaos. First of all. Thanks for the excellent work on the denote package. I use it every day and I really appreciate the work you have put into it.

I would like to propose an additional function to denote.el, named denote-org-capture-with-defaults. It very similar to the denote-org-capture-with-prompts, but instead of constructing a minibuffer prompt for each non-nil argument, it produces one for each nil argument instead. In addition, the arguments that are passed, are used as passed.

This is practical with the org-mode-capture when you want to limit the the user input. For example, if you have org-capture for a journal entry, you might want to default the subdirectory, tags, title and so on, and prompt the user only for what might differ for the template.

A minimal example of use:

```
(setq denote-org-capture-specifiers "%?")

(setq denote-templates
      '((journal . "* Entry\n")))

(with-eval-after-load 'org-capture
  (add-to-list 'org-capture-templates
      '("j" "Journal Entry" plain
        (file denote-last-path)
        (function (lambda () (denote-org-capture-with-defaults "My Journal Entry" '("journal" "entry") "journal-subdir" nil (alist-get 'journal denote-templates))))
        :no-save t
        :immediate-finish nil
        :kill-buffer t
        :jump-to-captured t)))
```

This specific example captures a note with a specific title, a specific set of tags, into a specific subdirectory using todays date, and with a specific template. The user does not need to be prompted in this case, and the capture buffer will open immediately.

An other example of a org-capture template might be:
```
(setq denote-templates
      '((journal . "* TODO \n")))

(with-eval-after-load 'org-capture
  (add-to-list 'org-capture-templates
      '("t" "Todo" plain
        (file denote-last-path)
        (function (lambda () (denote-org-capture-with-defaults nil '("task") "todo-subdir" nil (alist-get 'todo denote-templates))))
        :no-save t
        :immediate-finish nil
        :kill-buffer t
        :jump-to-captured t)))
```
This would prompt the user for a title, but default the tags, subdir, date, template template is defaulted, which makes it very efficient.

Let me know if you have any comments. Thanks a lot.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -4442,6 +4442,41 @@ option `denote-templates'."
     (when title (push 'title denote-prompts))
     (denote-org-capture)))
 
+;;;###autoload
+(defun denote-org-capture-with-defaults (&optional title keywords subdirectory date template)
+  "Like `denote-org-capture' but with optional default parameters.
+
+For each nil value passed, produce a minibuffer that corresponds to
+the TITLE, KEYWORDS, SUBDIRECORY, DATE and TEMPLATE arguments.
+
+Each argument is used passed to the stanard `denote' command and all
+of its utility commands.
+
+When returning the contents that fill in the Org capture
+template, the sequence is as follows: front matter, TEMPLATE, and
+then the value of the user option `denote-org-capture-specifiers'.
+
+Important note: in the case of SUBDIRECTORY actual subdirectories
+must exist---Denote does not create them.  Same principle for
+TEMPLATE as templates must exist and are specified in the user
+option `denote-templates'."
+  (let* ((title (if title title (denote-title-prompt)))
+         (kws (if keywords keywords (denote-keywords-prompt)))
+         (directory (file-name-as-directory (if subdirectory (concat (denote-directory) subdirectory) (denote-subdirectory-prompt))))
+         (date (if date (denote--valid-date date) (current-time)))
+         (id (denote--find-first-unused-id
+              (format-time-string denote-id-format date)
+              (denote--get-all-used-ids)))
+         (template (if template template (denote-template-prompt)))
+         (front-matter (denote--format-front-matter
+                        title (denote--date date 'org) kws
+                        (format-time-string denote-id-format date) 'org)))
+    (setq denote-last-path
+          (denote--path title kws directory id 'org ""))
+    (denote--keywords-add-to-history kws)
+    (message template)
+    (concat front-matter template denote-org-capture-specifiers)))
+
 (defun denote-org-capture-delete-empty-file ()
   "Delete file if capture with `denote-org-capture' is aborted."
   (when-let ((file denote-last-path)


jeanphilippegg:
There was a similar case recently (#241) where we wanted interactivity for *some* prompts only. In `denote-org-extras-extract-subtree`, we wanted to have some prompts and set the values of the title and keywords components.

Your use case is valid. I suggest a more general approach to allow this to work with any denote command.

My suggestion is as follow:

Introduce the variables `forced-title`, `forced-keywords`, `forced-file-type`, `forced-subdirectory`, `forced-date`, `forced-signature`. These variables, along with `denote-prompts`, would allow any denote command to be configured to meet one's need in term of interactivity.

Here is how you would configure `denote-org-capture`:

```
(with-eval-after-load 'org-capture
  (add-to-list 'org-capture-templates
      '("t" "Todo" plain
        (file denote-last-path)
        (function (lambda () (let ((denote-prompts '(title))
                                                (forced-keywords '("journal"))
                                                (forced-template 'journal))
                                            (denote-org-capture))))
        :no-save t
        :immediate-finish nil
        :kill-buffer t
        :jump-to-captured t)))
```

I think that reading the above it is obvious what is going to happen.

Maybe this is not necessary for now, but it is worth keeping in minding for when/if someone asks for a `denote-with-defaults` command that prompts for a title, but always set a specific signature.

kimothy:
@jeanphilippegg, thanks for the comment. The function that I wrote is just a slightly modified version of the existing _denote-org-capture-with-defaults_, as such is fits in with the way denote is written today. I do think that your proposal to make it more generic makes sense to.

protesilaos:
Hello @kimothy!

> I do think that your proposal to make it more generic makes sense to.

I also agree with the spirit of @jeanphilippegg point, [as also expressed here](https://github.com/protesilaos/denote/issues/257). As such, I wonder if you want to modify your code to be in line with this approach of generalising how we do things. It will be better for us long-term.

jeanphilippegg:
I have added the `denote-forced-*` variables in my fork of the code, which I will make public this weekend. Pull requests will be made in May.

You may consider closing this pull request, as it will be done (code is ready).

protesilaos:
> Date: Mon, 25 Mar 2024 19:07:05 -0700
>
> I have added the `denote-forced-*` variables in my fork of the code,
> which I will make public this weekend. Pull requests will be made in
> May.

Sounds good!

> You may consider closing this pull request, as it will be done (code is ready).

Let's keep this open until then to have a place for discussion. Then we
proceed with the new PR.

-- 
Protesilaos Stavrou
https://protesilaos.com

jeanphilippegg:
That is right. We should have `denote-use-*` (or with another name) soon.

protesilaos:
Thanks!

PR #258, opened 2024-02-17
Fix condition in denote-parse-date 

jarofromel:
Hi,
when using `denote-org-extras-extract-org-subtree` I found that `id` and `date` of the extracted note are always with current date (but `DATE` or `CREATED` properties were present in the original heading).
The rootcause is probably typo in the condition at the end of `denote-parse-date`.


protesilaos:
Merged. Thank you!

PR #260, opened 2024-02-21
Fix edge-cases when inserting missing links via org-dblock

vedang:
To reproduce the problem: open a new note and while it is empty, try to insert and populate an org dblock of type "denote-missing-links".

This leads to the following stack-trace:

```
Debugger entered--Lisp error: (wrong-type-argument stringp 65)
  file-name-nondirectory(65)
  file-name-extension(65 :period)
  denote-get-file-extension(65)
  denote-get-file-extension-sans-encryption(65)
  denote-filetype-heuristics(65)
  denote-link-description-with-signature-and-title(65 nil)
  denote--link-get-description(65)
  #f(compiled-function (file) #<bytecode -0x3aa12408d80ef11>)(65)
  mapc(#f(compiled-function (file) #<bytecode -0x3aa12408d80ef11>) "All links matching ‘reference/2024012’ are present")
  denote-link--prepare-links("All links matching ‘reference/2024012’ are present" org nil :no-other-sorting)
  denote-link--insert-links("All links matching ‘reference/2024012’ are present" org nil :no-other-sorting)
  org-dblock-write:denote-missing-links((:name "denote-missing-links" :regexp "reference/2024012" :indentation-column 0 :content #("\n" 0 1 (org-fold-core-fontified t fontified t))))
```

There are two issues in the underlying code:

1. When `denote-org-extras-dblock--get-missing-links` finds 0 missing-links, it returns with a string (return value of `message`) instead of with an empty list.

2. When there are 0 pre-existing denote links in the current buffer, `(denote-link--expand-identifiers
denote-org-link-in-context-regexp)` returns nil. This sets `linked-files` value to `nil` and short-circuits the `if-let` when it should not.

This commit fixes these two issues.

### DIFF
--- a/denote-org-extras.el
+++ b/denote-org-extras.el
@@ -232,11 +232,12 @@ Also see `denote-org-extras-dblock--files-missing-only'."
   "Return list of missing links to all notes matching REGEXP.
 Missing links are those for which REGEXP does not have a match in
 the current buffer."
-  (if-let ((found-files (denote-directory-files regexp :omit-current))
-           (linked-files (denote-link--expand-identifiers denote-org-link-in-context-regexp))
-           (final-files (seq-difference found-files linked-files)))
-      final-files
-    (message "All links matching `%s' are present" regexp)))
+  (let ((found-files (denote-directory-files regexp :omit-current))
+        (linked-files (denote-link--expand-identifiers denote-org-link-in-context-regexp)))
+    (if-let ((final-files (seq-difference found-files linked-files)))
+        final-files
+      (message "All links matching `%s' are present" regexp)
+      '())))
 
 (defun denote-org-extras-dblock--files-missing-only (files-matching-regexp &optional sort-by-component reverse)
   "Return list of missing links to FILES-MATCHING-REGEXP.


protesilaos:
Thank you! I wonder if we can achieve this by using something like the
following:

```diff
 denote-org-extras.el | 9 +++++----
 1 file changed, 5 insertions(+), 4 deletions(-)

diff --git a/denote-org-extras.el b/denote-org-extras.el
index f1d2e35..c6b2a07 100644
--- a/denote-org-extras.el
+++ b/denote-org-extras.el
@@ -245,10 +245,11 @@ (defun denote-org-extras-dblock--files-missing-only (files-matching-regexp &opti
 sorting.
 
 Also see `denote-org-extras-dblock--files'."
-  (denote-sort-files
-   (denote-org-extras-dblock--get-missing-links files-matching-regexp)
-   sort-by-component
-   reverse))
+  (when (listp files-matching-regexp)
+    (denote-sort-files
+     (denote-org-extras-dblock--get-missing-links files-matching-regexp)
+     sort-by-component
+     reverse)))
 
 ;;;;; Dynamic block to insert links
 

```

The reason is that the `message` we have there provides useful
information. It is nice to keep that sort of feedback around, if we
can.

protesilaos:
Sorry, updated diff (added the `and` because `(listp nil)` returns `t`):

```diff
 denote-org-extras.el | 9 +++++----
 1 file changed, 5 insertions(+), 4 deletions(-)

diff --git a/denote-org-extras.el b/denote-org-extras.el
index f1d2e35..50fe9b7 100644
--- a/denote-org-extras.el
+++ b/denote-org-extras.el
@@ -245,10 +245,11 @@ (defun denote-org-extras-dblock--files-missing-only (files-matching-regexp &opti
 sorting.
 
 Also see `denote-org-extras-dblock--files'."
-  (denote-sort-files
-   (denote-org-extras-dblock--get-missing-links files-matching-regexp)
-   sort-by-component
-   reverse))
+  (when (and files-matching-regexp (listp files-matching-regexp))
+    (denote-sort-files
+     (denote-org-extras-dblock--get-missing-links files-matching-regexp)
+     sort-by-component
+     reverse)))
 
 ;;;;; Dynamic block to insert links
 
```

vedang:
Hi @protesilaos,

Re `(and files-matching-regexp (listp files-matching-regexp))`: `files-matching-regexp` here is a regular expression, which we pass to `(denote-org-extras-dblock--get-missing-links files-matching-regexp)` and not a list of files.

The bug in the existing code is not that we are passing an empty regular expression, but that there are no existing denote links in the current buffer -- the call to `denote-link--expand-identifiers`.

Returning an empty list (or nil) from `denote-org-extras-dblock--get-missing-links` is actually honouring the contract of the function -- give me a list of missing links.

Before my commit, the function was returning a string (the return value of the `message`) while also echo'ing the string in the minibuffer. Now, the function still publishes the information for the end-user, but returns an empty list as expected by the calling function. 

I just want to clarify that I haven't removed the call to `message`, so the information is still being printed in the minibuffer.

protesilaos:
Thanks for the explanation! Merging now.

PR #263, opened 2024-02-24
Make denote-link-description-function require a function with only one argument

jeanphilippegg:
I have checked the documentation you made about
`denote-link-description-function`. There is a difference between what
is documented and the implementation. The current code requires that
`denote-link-description-function` be a function that takes two
parameters instead of only one. However, what you have documented is
better than what is implemented.

In this pull request, I removed this second parameter. It really is
unnecessary and may be confusing.

Since the functionality has not been released in an official version, it
does not seem to late to fix it.


protesilaos:
Merged. Thank you! This was the intent of my latest change, but I was sloppy.

Issue #202, opened 2023-12-13
Allow users to custom `denote-org-dblock--get-file-contents`?


WeissP:
Regarding `denote-org-dblock-insert-files`, users can currently only control inserted file contents via option `no-front-matter`, could we change this option to something like `extractor`, which receives a function for extracting file contents?

More detaily, we could modify [this part](https://github.com/protesilaos/denote/blob/8acd82b12afb7ca07e044a7eb2d6b4da1066ea12/denote-org-dblock.el#L158)

```elisp
(when no-front-matter
          (delete-region
           (if (natnump no-front-matter)
               (progn (forward-line no-front-matter) (line-beginning-position))
             (1+ (re-search-forward "^$" nil :no-error 1)))
           beginning-of-contents))
```

to something like 
```elisp
(when extractor (funcall extractor beginning-of-contents))
```

*I havn't yet checked whether it is possible to pass functions via org dynamic block, it is just an idea.*

My motivation is to extract only the contents before the first header, in my note system, this serves as a summary of the whole note. Here is what I have written for my extractor:

```elisp
(defun weiss-denote-extract-summary (beginning-of-contents)
  (delete-region
   (1+ (re-search-forward "^$" nil :no-error 1))
   beginning-of-contents)
  (when (re-search-forward "^\\*+ " nil :no-error 1)      
    (delete-region (pos-bol) (point-max))
    )
  (delete-blank-lines)
  )
```

Now I have just used `advice-add` to override `denote-org-dblock--get-file-contents` to make my custom extractor work, which is of course not a good idea.

protesilaos:
> From: Weiss ***@***.***>
> Date: Tue, 12 Dec 2023 21:09:51 -0800

> [... 19 lines elided]

> *I havn't yet checked whether it is possible to pass functions via org
> dynamic block, it is just an idea.*

I think this should work by passing the unquoted symbol of the function
as the value of the parameter. Like:

    :no-front-matter my-function

> My motivation is to extract only the contents before the first header,
> in my note system, this serves as a summary of the whole note. Here is
> what I have written for my extractor:

> [... 14 lines elided]

I think this is a good idea. Let's give it another week before working
on it. I just need this extra time to be sure we catch any obvious bugs
with the current stable version.

-- 
Protesilaos Stavrou
https://protesilaos.com

protesilaos:
[...]

PR #266, opened 2024-03-02
extend no-front-matter to filter

WeissP:
This is a demo to solve https://github.com/protesilaos/denote/issues/202. 

The idea is to change the keyword "no-front-matter" in dynamic block `denote-files` to "filter". It acts the same as before if a natural number is passed. To simulate `:no-front-matter t` in this implementation, one can pass `:filter 'no-front-matter`. Additionally, one can define a custom function which receives `beginning-of-contents` to do custom filtering, for example:
```
(defun denote-only-summary (beginning-of-contents)
    "Only keep contents before the first heading"
    (delete-region
     (1+ (re-search-forward "^$" nil :no-error 1))
     beginning-of-contents)
    (when (re-search-forward "^\\*+ " nil :no-error 1)      
      (delete-region (pos-bol) (point-max))
      )
    (delete-blank-lines)
    )
```

```
#+BEGIN: denote-files :regexp "my-note" :filter denote-only-summary 

#+END
```

Note that currently the [documentation](https://github.com/WeissP/denote/blob/f18014d753dde0b040dde16cea9692f7d00b3c3c/denote-org-extras.el#L421) for the parameter FILTER is missing as it is just a demo and the feature might change later.

### DIFF
--- a/denote-org-extras.el
+++ b/denote-org-extras.el
@@ -252,7 +252,7 @@ point to a file with a Denote file name."
               (let ((new-text (if desc
                                   (format "[[denote:%s%s]%s]" id search desc)
                                 (format "[[denote:%s%s]]" id search))))
-              (replace-match new-text :fixed-case :literal)))))
+                (replace-match new-text :fixed-case :literal)))))
         ;; TODO 2024-02-28: notify how many changed.
         (message "Converted as `file:' links to `denote:' links"))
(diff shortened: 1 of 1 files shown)


protesilaos:
I have not had the chance to review this. Sorry! We will do it after the release of Denote version `3.0.0`.

Issue #261, opened 2024-02-22
denote-rename-buffer should not be used for files that are not notes


brabalan:
I have set-up `denote-rename-buffer' as follows:
```emacs-lisp
(setq denote-rename-buffer-format "[D] %t (%k)")
(denote-rename-buffer-mode 1)
```

My problem is that it is used for every denote file, including my pdfs stored in denote and viewed in emacs. For them, the buffer name is `[D] ()`, as there is no title nor keyword to retrieve from the front matter. I see there is some code that prevents the renaming when the string is empty, but as I configured `denote-rename-buffer-format`, this is never the case for me.

Could it be possible to (have an option to) only rename buffers that are notes?

protesilaos:
Thank you @brabalan for reporting this!

You are right. I think we can work around this by tweaking `denote-rename-buffer--format`. The idea is to check the front matter only if the file is writable, otherwise retrieve data from the file name.

> Could it be possible to (have an option to) only rename buffers that are notes?

Yes, we can have this. Though I think a more do-the-right-thing sort of approach is better here to give meaningful, albeit not always perfect, names at all times.

protesilaos:
Perhaps this change is enough?

```elisp
(defun denote-rename-buffer--format (buffer)
  "Parse the BUFFER through the `denote-rename-buffer-format'."
  (when-let ((file (buffer-file-name buffer))
             (type (denote-filetype-heuristics file)))
    (string-trim
     (format-spec denote-rename-buffer-format
                  (list (cons ?t (cond
                                  ((not (denote-file-is-writable-and-supported-p file))
                                   (denote-retrieve-filename-title file))
                                  ((denote-retrieve-front-matter-title-value file type))
                                  (t  "")))
                        (cons ?i (or (denote-retrieve-filename-identifier file) ""))
                        (cons ?d (or (denote-retrieve-filename-identifier file) ""))
                        (cons ?s (or (denote-retrieve-filename-signature file) ""))
                        (cons ?k (cond
                                  ((not (denote-file-is-writable-and-supported-p file))
                                   (denote-retrieve-filename-keywords file))
                                  ((denote-retrieve-front-matter-keywords-value-as-string file type))
                                  (t  "")))
                        (cons ?% "%"))
                  'delete))))
```

The diff:

```diff
 denote-rename-buffer.el | 12 ++++++++++--
 1 file changed, 10 insertions(+), 2 deletions(-)

[...]

PR #268, opened 2024-03-04
rename-buffer: really retrieve keywords for non-text files

brabalan:
The fix for #261 did not work for keywords, as `denote-retrieve-front-matter-keywords-value-as-string` returns an empty string when no keyword is found, which is considered as satisfying the `cond`. I propose instead to get the list of keywords using `denote-retrieve-front-matter-keywords-value`, combine them if non-empty, otherwise get the keywords from the file name.


protesilaos:
Merged. Thank you!

Issue #271, opened 2024-03-06
`--` in titles


mentalisttraceur:
I feel like I've seen people mention this somewhere but I can't find an issue for it. So I'm opening this to

1. note that a user like me exists, this is what I want, this is what I'm *already doing*, and this is what I'll keep doing, hopefully with Denote's upstream help;

2. help others with similar wants but less relevant programming skills to do the same (caveat: my current way of doing this might break with future releases of Denote); and

3. add one more voice to request that this be officially supported.

---

I use a (superset of) the Denote file naming convention for many different files.

Like several others, I found that sometimes I want two levels of word separation in my titles, and the most natural thing to use is "--".

<details><summary>I won't use MixedCase for normal separation and single hyphens for Greater separation. As far as I'm concerned, that's worse. [expand if you want a detour on why]</summary>

1. MixedCase has measurably worse readability than dash-case. You can test this in your life right now by asking the nearest normal non-programmer to read some strings in MixedCase and dash-case and report impressions. (You'll also notice this if you ever spend long enough not reading any strings in MixedCase when you start again.)

2. Reasonably good programmatic transformation of arbitrary strings to dash-case and back is inherently simpler than to MixedCase and back, and involves less ambiguities. Not very relevant, but it makes certain UI possibilities less tractable out the gate, and your brain has to do much of the same work just to parse those strings into works.

3. Several file systems and other mechanisms are still case insensitive - in fact some people advocate for case insensitive UX as the default. For example, I use Emacs on Android and Windows, through Termux and WSL respectively, with my denote-directory on Android's emulates-limitations-of-an-SD-card shared storage and Windows' native file system respectively - and both of those are case-insensitive, which leads to fun annoyances like having to first rename a file to a dummy name and then the actual intended name to fix a capitalization issue.

4. MixedCase plays worse with lots of search GUIs - for example, as of a few months ago, VLC on Android failed to find `RollingInTheDeep__Adele.mp4` with search "Deep", but found `rolling-in-the-deep__adele.mp4` with search "deep" just fine.

</details>

I particularly find myself wanting this for files that have more than one title - either two alternative names where neither is more canonical than the other (I have several songs in my music folder like this), or a generic/vague title and a distinct/clarifying/searchable sub-title (I often want this when saving books as PDFs or some ebook format). (Incidentally, I have a separate set of hacks which allow me to reuse Denote's functions on filenames that don't have an ID.)

---

[...]

PR #281, opened 2024-03-09
Allow consecutive hyphens in titles

jeanphilippegg:
As discussed in #271, I have removed a few restrictions to the file
naming scheme.

- The dot character is allowed in file names. This can be useful if a
  signature looks like `2.3.0`.

- "--" are permitted in a title.

- "==" are permitted in a signature.

The last two points can lead to a file being badly fontified in Dired if
it makes use of this new relaxation, but I don't think this is
problematic. The fontification will have to be reworked soon to allow
the reordering of the file name components anyway. We can make it right
then.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -685,13 +685,13 @@ The note's ID is derived from the date and time of its creation.")
 (defconst denote-id-regexp "\\([0-9]\\{8\\}\\)\\(T[0-9]\\{6\\}\\)"
   "Regular expression to match `denote-id-format'.")
 
-(defconst denote-signature-regexp "==\\([^.]*?\\)\\(--.*\\|__.*\\|\\..*\\)*$"
+(defconst denote-signature-regexp "==\\([^.]*?\\)\\(==.*\\|--.*\\|__.*\\|\\..*\\)*$"
   "Regular expression to match the SIGNATURE field in a file name.")
 
-(defconst denote-title-regexp "--\\([^.]*?\\)\\(--.*\\|__.*\\|\\..*\\)*$"
+(defconst denote-title-regexp "--\\([^.]*?\\)\\(==.*\\|__.*\\|\\..*\\)*$"
   "Regular expression to match the TITLE field in a file name.")
 
-(defconst denote-keywords-regexp "__\\([^.]*?\\)\\(--.*\\|__.*\\|\\..*\\)*$"
+(defconst denote-keywords-regexp "__\\([^.]*?\\)\\(==.*\\|--.*\\|__.*\\|\\..*\\)*$"
   "Regular expression to match the KEYWORDS field in a file name.")
 
 (defconst denote-excluded-punctuation-regexp "[][{}!@#$%^&*()=+'\"?,.\|;:~`‘’“”/]*"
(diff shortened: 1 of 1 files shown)


README.org
```diff
@@ -1652,10 +1652,9 @@ Files names can contain any character that the file system
 permits. Denote imposes a few additional restrictions:
 
 + The tokens "==", =__= and =--= are interpreted by Denote and should
-  appear only once. =_= is the separator for keywords.
+  appear only once.
```
mentalisttraceur, line 1655: Seems like "should appear only once" is outdated wording?

denote.el
```diff
@@ -781,43 +781,53 @@ EXTRA-CHARACTERS is an optional string.  See
...
-   (replace-regexp-in-string
-    "_\\{2,\\}" "_"
-    (replace-regexp-in-string
-     "=\\{2,\\}" "=" str))))
+Hyphens, underscores and equal signs are replaced with a single
+one in str, if necessary according to COMPONENT."
+  (cond (;; -- are allowed in titles
+         (eq component 'title)
+         (replace-regexp-in-string
+          "_\\{2,\\}" "_"
```
mentalisttraceur, line 804: How do y'all feel about `__+` instead of `_\\{2,\\}`? To me it's easier to read readable.

jeanphilippegg: Yes. This is more readable.

README.org
jeanphilippegg: I did not want to overwhelm users with users in the manual even though we are technically more permissive.

denote.el
```diff
@@ -681,13 +681,13 @@ The note's ID is derived from the date and time of its creation.")
...
-(defconst denote-signature-regexp "==\\([^.]*?\\)\\(--.*\\|__.*\\|\\..*\\)*$"
+(defconst denote-signature-regexp "==\\(.*?\\)\\(--.*\\|__.*\\)*$"
   "Regular expression to match the SIGNATURE field in a file name.")
 
-(defconst denote-title-regexp "--\\([^.]*?\\)\\(--.*\\|__.*\\|\\..*\\)*$"
+(defconst denote-title-regexp "--\\(.*?\\)\\(__.*\\|==.*\\)*$"
   "Regular expression to match the TITLE field in a file name.")
 
-(defconst denote-keywords-regexp "__\\([^.]*?\\)\\(--.*\\|__.*\\|\\..*\\)*$"
+(defconst denote-keywords-regexp "__\\(.*?\\)\\(--.*\\|__.*\\|==.*\\)*$"
```
mentalisttraceur, line 690: Bug? This regex will also capture the extension:

```lisp
(let ((file "01020304T123456--foo__qux.md"))
    (string-match "__\\(.*?\\)\\(--.*\\|__.*\\|==.*\\)*$" file)
    (match-string 1 file))
```

mentalisttraceur: I relate to the desire to enable as many characters as possible\*, but I think `.` might just be too prone to pitfalls. Especially if you want to let people change the order of fields later.

Remember also that outside of notes, people might have files like `20230000T000000--foo__bar_qux.tar.gz` (is it `qux.tar` or tag `qux` and inner extension `.tar`? what if we replace `.tar` with `.xyz`?)

\* Although over the years I have come to appreciate limitations. A good set of tight limitations empowers ecosystem contributions by simplifying the problem space of interoperability, and I think the right balance of limitations and flexibility is just creating something just barely so useful that you decide to tolerate the limitations.

jeanphilippegg: Yes, this regex, as-is, captures the extension. But I also modified the places where it is used to first strip the extension and THEN apply the regexp. The regex has already been removed on the string the we apply the regexp on.

I also understand that some extensions have multiple `.` in them, like `.tar.gz`. This is not perfect and we cannot really do better then `file-name-extension`. `file-name-extension` also fails to recognize the "right" extension here. Denote would be no better.

jeanphilippegg: By the way, it is not has crucial as one may think to parse the file name with perfect accuracy. Generally, the information in the front matter has priority (when available) over what is in the file name. Same with fontification. Even if there are some edge cases that are not perfectly handled, this does not break Denote internally. Moreover, this information that we retrieve in the file name is mostly used to provide suggestions in the creation or renaming commands.

mentalisttraceur: > it is not has crucial as one may think to parse the file name with perfect accuracy

So the issue that I see is:

1. Having a file like `foo__bar.tar.gz`
2. I call my keybind to
    * remove tags: it presents candidate `bar.tar` instead of `bar`;
    * add tags: I type/select `qux`, it renames my file to `foo__bar.tar_qux.gz`.

jeanphilippegg: You're right! I have removed that part from the pull request.

jeanphilippegg:
I have removed the part about the dot characters. We cannot allow it in file name components, because, as explained by @mentalisttraceur, it is possible that part of the extension would be removed when renaming `foo__bar.tar.gz`. Thanks for the explanations, @mentalisttraceur .

So this pull request fixes issue #271 only.

jeanphilippegg:
I have removed the possibility of consecutive = signs in signatures because I intend to reuse "==" for identifiers that do not appear as the first component of the file name. Though this would not really be an issue, it makes fontification harder and I think it is okay to have a bit more restrictions on keywords and signatures than titles.

This means that this pull request only allows consecutive hyphens in titles.

pprevos:
> * The dot character is allowed in file names. This can be useful if a
>       signature looks like `2.3.0`.

A delimiter inside signatures would be quite useful because we could then visualise the structure of signatures with Denote-Explore like this: https://zettelkasten.de/introduction/2020-08-13_folgezettel-sequence.png

jeanphilippegg:
Can you give an example file name of what you mean? I am not sure I understand. What is meant by "delimiter" are "--", "__" and "==". Also the dot character is not allowed in file name components.

pprevos:
@jeanphilippegg You wrote about allowing dots in a signature, which I think is a sensible in principle.

When using the Zettelkasten method, signatures are divided in blocks. See the [image I linked above](https://zettelkasten.de/introduction/2020-08-13_folgezettel-sequence.png). 

Example filename: `19970101T183001==1.1a.1a2--kebab-case-title__snake_case_keywords.org`

However, users can currently use `=` (`==1=1a=1a2`), so perhaps this point is moot.

jeanphilippegg:
Ok, I understand. However, there are issues with allowing dot characters in file names. A file `20240303T111111--foo.tar.gz` could be broken after renaming the title. I think it is safer to keep this restriction instead of guessing the extension.

pprevos:
> there are issues with allowing dot characters in file names

As I mentioned, users can use `=` as a delimiter in signatures, so problem is solved.

jeanphilippegg:
I will make the code of my fork of the repo public this weekend. I am closing this as the relevant pull request will be made in May.

PR #282, opened 2024-03-09
Disable sluggification of keywords and signature, by default

jeanphilippegg:
I suggest disabling sluggification of the keywords and signature
components of file names, by default, for these reasons:

- Sluggification of titles makes sense because it is a string of free
  text and it is desirable to make it "clean", by default, in a file's
  name. On the other hand, keywords and signature can be input in a
  correct manner from the start by the user.

- CamelCase or other schemes will be supported automatically.

- It is less necessary now that we always enforce the few rules that
  remains to the file naming scheme. Previously, the sluggification was
  taking care of this aspect, but not anymore.

- Ultimately, I would like to propose a change of the keywords separator
  from "_" to "__" and this step will help to make this less impactful
  to users.

### DIFF
--- a/README.org
+++ b/README.org
@@ -1727,8 +1727,8 @@ default value:
 
 #+begin_example emacs-lisp
 '((title . denote-sluggify-title)
-  (signature . denote-sluggify-signature)
-  (keyword . denote-sluggify-keyword))
+  (signature . identity)
+  (keyword . identity))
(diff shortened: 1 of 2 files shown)


jeanphilippegg:
I will make the code of my fork of the repo public this weekend. I am closing this as the relevant pull request will be made in May.

PR #291, opened 2024-03-16
Make denote--retrieve-title-or-filename public

pprevos:
`denote--retrieve-title-or-filename` is used in the `denote-explore` package.


protesilaos:
Merged. Thank you! I will make some more tweaks on top of this.

Issue #292, opened 2024-03-16
Commit 232bd90 does not only apply to signatures


jeanphilippegg:
This is a comment on commit 232bd90 where signatures are treated specially in renaming commands.

The issue this commit is trying to fix does not only apply to the signature component of the file name. For example, if `denote-prompts` does not contain `title`, then `denote-rename-file` will suppress the title. This is also undesirable.

I suggest instead that the renaming commands keep obeying `denote-prompts` and that `denote-rename-file` only modify the file name components for which we are being prompted and keep the others untouched. This would make it work with all components and the experience would be consistent and intuitive.

jeanphilippegg:
I think it also defeats the intention of making them obey `denote-prompts`. Someone that has configured multiple renaming commands with different values of `denote-prompts` tailored to its need in the moment will now always be prompted for a signature (assuming all his notes have signatures). I think this is undesirable.

jeanphilippegg:
By the way, I think `denote-keywords-add`, `denote-keywords-remove`, `denote-signature-add` and `denote-signature-remove` can all be implemented by let binding `denote-prompts` appropriately. So it may be better to leave users to decide which commands they want in their configuration and explain how Denote can be configured in such a way in the manual. We can also provide them as convenience functions.

protesilaos:
> From: Jean-Philippe Gagné Guay ***@***.***>
> Date: Sat, 16 Mar 2024 06:46:59 -0700
>
> I think it also defeats the intention of making them obey
> `denote-prompts`. Someone that has configured multiple renaming
> commands with different values of `denote-prompts` tailored to its
> need in the moment will now always be prompted for a signature
> (assuming all his notes have signatures). I think this is undesirable.

You are right! I did not think it through. Your suggestion of ignoring
them instead of deleting them is better. Do you want to send a patch or
shall I do it?

-- 
Protesilaos Stavrou
https://protesilaos.com

jeanphilippegg:
I can do it. I'll send the patch in 1-2 hours.

protesilaos:
> From: Jean-Philippe Gagné Guay ***@***.***>
> Date: Sat, 16 Mar 2024 07:00:07 -0700
>
> By the way, I think `denote-keywords-add`, `denote-keywords-remove`,
> `denote-signature-add` and `denote-signature-remove` can all be
> implemented by let binding `denote-prompts` appropriately. So it may
> be better to leave users to decide which commands they want in their
[...]

PR #293, opened 2024-03-17
Renaming commands

jeanphilippegg:
In this pull request, I have done the following:

- Revert commit 232bd90 because of what is explained in #292.
- Make `denote-dired-rename-files` also obey the *order* of elements in
  `denote-prompts`.

I have also simplified a `:safe` declaration, since `nil` is a list (the empty list).

### DIFF
--- a/README.org
+++ b/README.org
@@ -1224,12 +1224,6 @@ The command does the same for the =SIGNATURE= prompt, subject to
 ~denote-prompts~, by prefilling the minibuffer with the current
 signature of =FILE=, if any.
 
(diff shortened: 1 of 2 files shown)


jeanphilippegg:
I have also done the fix itself : Only rename the prompted file name components and keep the others unchanged.

protesilaos:
Merged. Thank you!

protesilaos:
I had to make a change because if the `denote-prompts` had no keywords, then we were passing a string instead of a list of strings.

PR #295, opened 2024-03-23
Remove denote-desluggify

jeanphilippegg:
Hello Prot,

I suggest that we remove `denote-file-name-deslug-function` from the
upcoming version 2.3.0 because it will be new in this version and
it is actually not very useful.

First, desluggification of keywords and signature is never used in the
code base, so this part is totally useless. As for the title
desluggification, it is only used in
`denote-retrieve-title-or-filename`. This is used in renaming commands
to provide a suggested title and for the link descriptions. This only
applies when we cannot find the title from the front-matter and there is
a title in the file name. I think the raw title (not desluggified) is
not a bad suggestion/description in those cases. The desluggified
suggestion may not be useful enough to warrant a user option.

Ignore this pull request if you think this is still useful and we should
keep it.

### DIFF
--- a/README.org
+++ b/README.org
@@ -4313,11 +4313,6 @@ might change them without further notice.
 + Function ~denote-sluggify-keyword~ :: Sluggify =STR= while joining
   separate words.
 
-#+findex: denote-desluggify
-+ Function ~denote-desluggify~ :: Upcase first char in =STR= and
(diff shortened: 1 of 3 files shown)


protesilaos:
Merged. Thank you!

Issue #283, opened 2024-03-09
Path in `denote-backlinks` buffer is wrong (extra directory in the link)


brabalan:
When I run `denote-backlinks` on a note, I get a listing of the form
```
Backlinks to "Marine Ludo" (20240209T154753)
--------------------------------------------

notes/20210323T111850--mot-malin__boardgame.org
notes/20230605T171758--trio__boardgame.org
notes/20230609T161606--phantom-ink__boardgame.org
notes/20230929T171050--lets-summon-demons__boardgame.org
notes/20240209T154425--6-qui-prend__boardgame.org
```

The problem is the extra `notes/` at the beginning of each link, as `denote-directory` already includes it.

I don’t understand how xref works, so I don’t know how to debug this…

If it matters, my notes live in `~/Document/Org/notes`, `~/Documents/Org` is under version control, and I use projectile (so the path to the current projectile project is `~/Documents/Org` when looking at the notes).

jeanphilippegg:
@protesilaos This is caused by commit ac0c361. When removing the modules, we also removed `project.el integration` (and `xref integration`). The function that we need back is `denote-project-find`.

You can see the differences when you evaluate this : 

```
(xref--analyze (xref-matches-in-files "20240309T105635" (denote-directory-files nil t t)))
```
vs
```
(let ((project-find-functions #'denote-project-find)) (xref--analyze (xref-matches-in-files "20240309T105635" (denote-directory-files nil t t))))
```
The group names are relative to the project's root, which may not be `denote-directory`.

The solution is to bring back `denote-project-find` from this commit (or the section on `project.el integration`.

@brabalan Until it is fixed, you can evaluate `(setq xref-file-name-display 'abs)`. Your results will be in absolute form, but they will work.

brabalan:
Thank you very much for the explanation and for the workaround.

protesilaos:
Hello folks! I just pushed this commit:

```
commit 3e7273953aed79bb8fc6c10652ebd4a8d8ae4f4c
Author: Protesilaos Stavrou <info@protesilaos.com>
Date:   Wed Mar 13 09:18:34 2024 +0200

  Reinstate denote-project-find function for project/xref to work properly
  
  We had removed this in commit ac0c361, thinking it was not necessary,
  though Alan Schmitt reported a case where we actually needed this
  functionality. This was reported in issue 283:
  <https://github.com/protesilaos/denote/issues/283>.
  
[...]

PR #296, opened 2024-03-23
Actually... remove `project.el` integration again

jeanphilippegg:
When removing modules, we removed `project.el` integration at the same
time. This caused an issue with backlinks because xref can use relative
file names (relative to the current project).

However, it is better to have xref retrieve its results in absolute form
and then arrange the results to relative form ourselves because,
eventually, notes will not be all under the same directory/project.

I have thus removed `project.el` integration again and fixed the code to
still use relative paths. I am unsure why we added `denote-project-find`
to `project-find-functions` hook for backlinks buffer. I suspect it was
to support relative paths somehow. I have removed it and tested the
backlinks buffers and they work as intended.

This is ready to be merged. It can be part of 2.3.0 as you have already
removed Denote modules.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -4073,31 +4073,6 @@ To be assigned to `markdown-follow-link-functions'."
 
 ;;;;; Backlinks' buffer
 
-(cl-defmethod project-root ((project (head denote)))
-  "Denote's implementation of `project-root' method from `project'.
-Return current variable `denote-directory' as the root of the
-current denote PROJECT."
-  (cdr project))
-
-(cl-defmethod project-files ((_project (head denote)) &optional _dirs)
-  "Denote's implementation of `project-files' method from `project'.
-Return all files that have an identifier for the current denote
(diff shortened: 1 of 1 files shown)


protesilaos:
We did it because of what @brabalan reported here: https://github.com/protesilaos/denote/issues/283

I am fine without this. I prefer absolute paths because they are unambiguous and we can then do whatever we need with them.

I plan to publish version 2.3.0 at around 15:00 Europe/Athens time. But I can wait for this to be resolved first.

Thank you!

jeanphilippegg:
It is already resolved and ready to be merged!

jeanphilippegg:
To be clear: I removed the project.el integration part and I was able to fix the bug reported in #283 without it. xref is working with absolute paths, but I arrange the results to be relative myself. It just works as before!

protesilaos:
Merged. Thank you! Will prepare the new release today.

Issue #298, opened 2024-03-25
Can't add link to file from within org-capture: "The current file type is not recognized by Denote"


svnsbck:
Hi Prot!

I use the `denote-link` command to add links to files within my todo's  with `org-capture` which somehow stopped working after the recent update. 

I realized your blog-post about the options the option to link to headers and set also `denote-org-store-link-to-heading`~ to `nil` but only get "The current file type is not recognized by Denote".  Adding links from within an org-file seems to work as expected.

I am sure I am missing the obvious here somewhere, but any help is appreciated!

All the best, Sven

protesilaos:
Thank you Sven! This is a bug. I pushed two tweaks to fix it. A new version of the package will be available in a few hours.

Once you get the package, please confirm if the problem is solved on your end.

svnsbck:
Hi Prot!

Nice, the latest update works.  I realized that the link title now includes the signature as well.  Is this expected behaviour?  I don't think Denote did this earlier.

protesilaos:
I now realise I did not document this in the changelog... Thanks Sven! When a file has a signature, a link to it will include it. Before, we had a separate command to do this. The `denote-link-description-function` variable controls this behaviour. I am now thinking this should be a user option with a choice between:

1. Title link.
2. Signature plus title.
3. Arbitrary function.

Copying @jeanphilippegg who modified this part for any thoughts on the matter. I know adding a `defcustom` here is technically a new feature, though we really want to keep it small to basically announce the variable that is already there (makes it discoverable through the Custom UI).

protesilaos:
The diff for what I described above (NOT TESTED YET):

```diff
 denote.el | 50 ++++++++++++++++++++++++++++++++++++++++----------
 1 file changed, 40 insertions(+), 10 deletions(-)

diff --git a/denote.el b/denote.el
index 86e8960..1f3877a 100644
--- a/denote.el
+++ b/denote.el
@@ -676,6 +676,21 @@ (make-obsolete
  'denote-file-name-slug-functions
  "2.3.0")
 
+(defcustom denote-link-description-function #'denote-link-description-with-signature-and-title
+  "Function to use to create the description of links.
+
+The function specified should take a FILE argument and should
+return the description as a string.  By default, the title of the
+file is returned as the description."
+  :group 'denote
+  :type '(choice
+          (function :tag "Link to title and include signature, if present"
+                           denote-link-description-with-signature-and-title)
+          (function :tag "Link to title only"
[...]

PR #300, opened 2024-03-26
Fix error message for all linking commands in Org capture buffers

jeanphilippegg:
#298 affected all renaming commands : `denote-link-or-create`, `denote-link-after-creating`, `denote-link-with-signature`, `denote-link-after-creating-with-command`.

`denote-add-links` was not affected, but it benefits from this check as well. It was crashing when used on a non-supported buffer.


protesilaos:
Fine. Merged now. Thanks!

PR #305, opened 2024-03-31
Remove denote--path

jeanphilippegg:
It was not doing much more than `denote-format-file-name`.


protesilaos:
Good! This was one of the older functions and a lot has changed since then. Merged!

PR #306, opened 2024-03-31
Issue error if buffer is not recognized in denote-link-dired-marked-notes

jeanphilippegg:


protesilaos:
Merged, thank you! By the way, the docstring reads:

> FILES are Denote notes, meaning that they have our file-naming scheme, are writable/regular files, and use the appropriate file type extension (per `denote-file-type').

I think we do not want this restriction. We can link to images and PDFs, for example.

PR #307, opened 2024-04-01
Issue error in denote--valid-file-type if denote-file-types is nil

jeanphilippegg:
Maybe someone has set `denote-files-types` to nil because he only uses the file name scheme of Denote.

I issue an error in `denote--valid-type` instead of failing later.


protesilaos:
Merged, thank you!

PR #308, opened 2024-04-01
Make denote command return the path of the new note

jeanphilippegg:


protesilaos:
Good idea! Merged, thank you!

PR #310, opened 2024-04-03
Extend org file type detection

brabalan:
- rename `denote--file-type-org-capture-p` into `denote--file-type-org-extra-p`
- add case for an `*Org Note*` buffer


protesilaos:
Merged, thank you!

PR #313, opened 2024-04-06
Use denote-extract-keywords-from path in renaming commands

jeanphilippegg:
Just a small refactor.


protesilaos:
Merged, thank you!

PR #320, opened 2024-04-14
Minor cleanup in `denote-keywords--combine`

jeanphilippegg:
Also remove a small unnecessary check.


protesilaos:
Merged, thank you!

PR #322, opened 2024-04-14
Remove unused functions

jeanphilippegg:
Consider removing `denote-convert-keywords-to-crm`. We never use it because it always makes more sense to work with keywords as a list. It is better to use `denote-extract-keywords-from-path`.

We can also remove `denote-retrieve-front-matter-keywords-value-as-string` for similar reasons.

Let me know if you think it is necessary to make them obsolete for now instead or if you would like to keep one or both of the functions. I think it is better to encourage the manipulation of keywords as a list and we already have suitable functions for that.


protesilaos:
Merged, thank you! I will add the obsolete notice now.

Issue #274, opened 2024-03-06
Option to ask `y` or `n` just once from `denote-rename-file`


mentalisttraceur:
One of the things I immediately monkey-patched in Denote once I started using it was to disable *just* the frontmatter rewrite confirmation (but *not* the file rename confirmation) in my calls to `denote-rename-file`.

I didn't want to disable *all* confirmation in interactive calls. It's very useful to be asked so that I can double-check/abort if I made a mistake.

I just didn't want *two* confirmations, nor to have the number of confirmations depend on the type and location of the file (I like my UX uniform so that the same muscle memory applies to all uses of the same operation, and I can already start executing movements for the next operation without *having to* make my movements conditional on more specific information - a confirmation that's only asked for file renames when that file can have Denote frontmatter complects one more thing into my thinking+movements).

If I take the time to check that my inputs were correct and I'm operating on the right file, questions to which the sluggified file name confirmation is an almost-complete and concise answer, then as a *user* of Denote's stable releases, I trust that the front matter rewrite code is going to do the right thing.

---

The way I currently do it basically looks like this (except my actual wrapper does more than this):

```lisp
(defun hack-denote-rewrite-front-matter
        (denote-rewrite-front-matter path &rest arguments)
    (with-advice ('y-or-n-p :override 'always)
        (apply denote-rewrite-front-matter path arguments)))

(defun my-denote-rename-file-wrapper (...)
    (with-advice ('denote-rename-file :around 'hack-denote-rename-file)
        (denote-rename-file ...)))
```

A more complete version would have to also advise `denote--add-front-matter`:

```lisp
(defun suppress-denote-front-matter-y-or-n-p (function path &rest arguments)
    (with-advice ('y-or-n-p :override 'always)
        (apply function path arguments)))

(defun my-denote-rename-file-wrapper (...)
    (with-advice ('denote-rewrite-front-matter
                      :around 'suppress-denote-front-matter-y-or-n-p
                  'denote--add-front-matter
                      :around 'suppress-denote-front-matter-y-or-n-p)
        (denote-rename-file ...)))
```

---

I'd appreciate support in-package for this.

The lisp advice to do it externally has to be coupled to implementation details, and has to be more complicated than support in-package would be - note, in particular, that `with-advice` doesn't ship with Emacs, so you must either hand-roll that with `unwind-protect`, or use some other way if you want to limit the behavior change to just these calls - the only way to make it simpler from the outside is to suppress confirmation in more cases.

[...]

Issue #299, opened 2024-03-25
Problem due to overloading `no-confirm` for "save"


mentalisttraceur:
I noticed that 2.3.0 overloaded the `no-confirm` parameter of `denote-rewrite-front-matter` to both turn off confirmation and turn on unconditional saving.

I appreciate that this was probably to help users like me who want both of these features, so thank you for that.

However, to me it's important to be able to control these two behaviors independently. Just because I don't want to be prompted doesn't mean I want to risk unrelated unwanted saving/clobberring, and if my checks report it's not safe to save (as described in #273) that still doesn't mean I want to be prompted about the frontmatter diff.

(I also don't think it's just me. These are two qualitatively different kinds and severity of (un)safety. Combining them in one boolean collapses their individual trade-offs to the worst of both: you either get prompted to confirm the in-buffer change just to not overwrite/save unrelated changes, or you have to pay the risk of the latter just to not deal with the former.)

Since neither of the new combined behaviors of `no-confirm` is ever the right thing for me, I am now back to injecting advice (personally I'm fine with this, almost every package I use I end up advising something, but I think it's a good symptom / illustration that two independent booleans got ORed into one here):

```lisp
    (defun my-denote-rewrite-front-matter-wrapper (path title tags type)
        (let* ((buffers (buffer-list))
               (buffer  (find-file-noselect path))
               (was-already-open    (memq buffer buffers))
               (had-unsaved-changes (when was-already-open
                                        (refresh-modified-state buffer)
                                        (buffer-modified-p buffer))))
            (with-advice ('y-or-n-p :override 'always)
                (denote-rewrite-front-matter path title tags type))
            (unless had-unsaved-changes
                (with-current-buffer buffer
                    (save-buffer)))
            (unless was-already-open
                (kill-buffer buffer))))
```

Personally, I would recommend

1. return `no-confirm` to its uncoupled behavior of just controlling the prompt;

2. one of either

    * add the additional optional save parameter that was mentioned in #273,
    * (if you're not already using the return value for anything) just return non-nil if you did the rewrite and `nil` otherwise (maybe it already ends up doing this, make it explicitly part of the API if so) or
    * make possible to reuse the front matter rewrite prompt, same as we can use the rename prompt

    Any one of these three allows people who *do* want the prompt to have the save conditional on the prompt. Contrast in my wrapper above, where if I did want to re-enable the prompt, my code couldn't easily hold back from saving even if I hit `n`.

[...]

PR #324, opened 2024-04-20
Fix issues #274 and #299

jeanphilippegg:
This pull request fixes issues #274 and #299.

The current renaming functions are inconsistent. Some of them have
optional parameters `no-confirm` and `save-buffer`. Some have only one,
but it means both. Some have no optional parameters (but could have).

In this pull request, all renaming commands obey `denote-save-buffers`
(renamed from `denote-save-buffer-after-creation`) and
`denote-rename-confirmations` (see its docstring).

There are no user-visible changes, except that all optional parameters
have been removed in favor of obeying these two variables. Note that we
explicitly let-bind them ourselves in some commands, where it makes
sense.

The manual remains to be updated. Let me know how you want to proceed.
I can update it if you want.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -140,8 +140,10 @@ the function `denote-directory' instead."
   :link '(info-link "(denote) Maintain separate directories for notes")
   :type 'directory)
 
-(defcustom denote-save-buffer-after-creation nil
-  "Control whether commands that creeate new notes save their buffer outright.
+(define-obsolete-variable-alias 'denote-save-buffer-after-creation 'denote-save-buffers "3.0.0")
+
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
The code is ready to be merged.

There are still commands that could be made to comply with `denote-save-buffers` and `denote-rename-confirmations`, but there is no need to fix them now as I will soon make a pull request to remove all code duplication from renaming commands. They will all become consistent then.

The code in its current state should be working and a step in this direction.

protesilaos:
Merged. Thank you!

PR #329, opened 2024-04-26
Refactorings

jeanphilippegg:
Two commits. Nothing major or breaking.

- Use `denote-retrieve-front-matter-title-value` instead of an alias
- Use `denote-retrieve-filename-identifier` and compare id with `string=` instead of `string-prefix-p`. This is more general and will be ready for when file name components are in a different order.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -4184,7 +4184,7 @@ concomitant alist, such as `denote-link-backlinks-display-buffer-action'."
 Use FILE to detect a suitable title with which to name the buffer.  Else
 use the ID."
   (if-let ((type (denote-filetype-heuristics file))
-           (title (denote-retrieve-title-value file type)))
+           (title (denote-retrieve-front-matter-title-value file type)))
       (format "*Denote FILE backlinks for %S*" title)
     (format "*Denote FILE backlinks for %s*" id)))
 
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged. Thank you!

For the name of the backlinks buffer, I wonder if we should have third condition:

1. Use title from the front matter.
2. Use title from the file name.
3. Use the identifier.

jeanphilippegg:
> For the name of the backlinks buffer, I wonder if we should have third condition:
> 
>     1. Use title from the front matter.
> 
>     2. Use title from the file name.
> 
>     3. Use the identifier.

We previously used `denote-retrieve-title-or-filename` which is almost what you suggest here. Maybe we could go back to using that and delete `denote--backlinks-get-buffer-name`? The identifier should be visible in the file name if we reach 3.

protesilaos:
> Date: Fri, 26 Apr 2024 21:43:02 -0700
>
>> For the name of the backlinks buffer, I wonder if we should have third condition:
>> 
>>     1. Use title from the front matter.
>> 
>>     2. Use title from the file name.
>> 
>>     3. Use the identifier.
>
> We previously used `denote-retrieve-title-or-filename` which is almost
> what you suggest here. Maybe we could go back to using that and delete
> `denote--backlinks-get-buffer-name`? The identifier should be visible
> in the file name if we reach 3.

I guess that would be fine too because the user probably wants to read
the keywords and/or signature in the absence of a title.

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #330, opened 2024-04-26
Add denote--used-ids and simplify renaming commands

jeanphilippegg:
An internal simplification that makes the hash table of used-ids more transparent in the code.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1652,8 +1652,13 @@ current time."
    denote-id-format
    (when date (denote-valid-date-p date))))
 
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged. Thank you!

Do you think we need to persist this new variable between sessions? Maybe by writing it to a file ourselves or using the `persist` package?

jeanphilippegg:
Absolutely not! :)

We never want it to be persisted actually. This is a dynamic variable that we want to let-bind only in the few commands that handle multiple files at once (for performance reasons). Otherwise, it should stay nil. This is what I wanted to express in its docstring.

PR #331, opened 2024-04-26
Fix issue #271 (Allow -- in titles)

jeanphilippegg:
A pull request to fix issue #271. Has no impact on users that keep the default sluggification. Those that don't are allowed to have consecutive hyphens in titles. We do not remove them.

The fontification does not need to be updated. It handles this correctly already.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -711,13 +711,13 @@ The note's ID is derived from the date and time of its creation.")
 (defconst denote-id-regexp "\\([0-9]\\{8\\}\\)\\(T[0-9]\\{6\\}\\)"
   "Regular expression to match `denote-id-format'.")
 
-(defconst denote-signature-regexp "==\\([^.]*?\\)\\(--.*\\|__.*\\|\\..*\\)*$"
+(defconst denote-signature-regexp "==\\([^.]*?\\)\\(==.*\\|--.*\\|__.*\\|\\..*\\)*$"
   "Regular expression to match the SIGNATURE field in a file name.")
 
-(defconst denote-title-regexp "--\\([^.]*?\\)\\(--.*\\|__.*\\|\\..*\\)*$"
+(defconst denote-title-regexp "--\\([^.]*?\\)\\(==.*\\|--.*\\|__.*\\|\\..*\\)*$"
   "Regular expression to match the TITLE field in a file name.")
 
-(defconst denote-keywords-regexp "__\\([^.]*?\\)\\(--.*\\|__.*\\|\\..*\\)*$"
+(defconst denote-keywords-regexp "__\\([^.]*?\\)\\(==.*\\|--.*\\|__.*\\|\\..*\\)*$"
   "Regular expression to match the KEYWORDS field in a file name.")
 
 (defconst denote-excluded-punctuation-regexp "[][{}!@#$%^&*()=+'\"?,.\|;:~`‘’“”/]*"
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged. Thank you!

Though now we need to think how best to design this to make it opt-in for users who want more flexibility with the file-naming scheme. Maybe it is too early to add a user option, given the planned changes you are working on, but it is something to consider once things have settled.

jeanphilippegg:
Is it not already opt-in? Currently, with the default sluggification as it is, it is not possible to have consecutive hyphens in a file name title. If users disable the title sluggification, that's when they are allowed to type any title that they want, including the "-----" if they want.

I think that "disabling all sluggification" to mean "allow as much flexibility as possible in file name components" is intuitive.

jeanphilippegg:
Actually, I am not sure I understood your point about an opt-in option. I think we should impose as few rules as possible to the file naming scheme when sluggification is disabled. Then, with the default sluggification, we "impose" some defaults.

Here are the strict rules of the file naming scheme (without sluggification), as they currently stand:
- No dot character except for the extension.
- No consecutive delimiter characters (-,_,=) in the components, except for the title which accepts multiple hyphens.
- No delimiter character at the end of a component (to prevent conflict with a possible next component).

I think we should aim for as much flexibility as possible when sluggification is disabled. Then, through sluggification, users can impose further rules on their file name components.

protesilaos:
> Date: Fri, 26 Apr 2024 21:57:31 -0700
>
> Actually, I am not sure I understood your point about an opt-in
> option. I think we should impose as few rules as possible to the file
> naming scheme when sluggification is disabled. Then, with the default
> sluggification, we "impose" some defaults.
>
> Here are the strict rules of the file naming scheme (without sluggification), as they currently stand:
> - No dot character except for the extension.
> - No consecutive delimiter characters (-,_,=) in the components, except for the title which accepts multiple hyphens.
> - No delimiter character at the end of a component (to prevent conflict with a possible next component).
>
> I think we should aim for as much flexibility as possible when
> sluggification is disabled. Then, through sluggification, users can
> impose further rules on their file name components.

I agree with the flexibility part. What I am not understanding is the
intent: how would the user input the multiple dashes in this scenario.
Is it by configuring 'denote-file-name-slug-functions'?

I just want to make sure what the expectation is to document things
accordingly.

-- 
Protesilaos Stavrou
https://protesilaos.com

jeanphilippegg:
There are 2 cases:
- The user accepts the default sluggification. He cannot input "--" in titles because of the default sluggification.
- The user disables the default sluggification for titles. He can input (almost) anything in the title, including "--".

> Is it by configuring 'denote-file-name-slug-functions'?

Yes, a user would need to disable the default sluggification through this option to be able to input "--" in titles.

I think we only need to document the rules of the file naming scheme (in my previous comment) and the default sluggification.

protesilaos:
> Date: Fri, 26 Apr 2024 23:08:48 -0700

> [... 4 lines elided]

>> Is it by configuring 'denote-file-name-slug-functions'?
>
> Yes, a user would need to disable the default sluggification through this option to be able to input "--" in titles.
>
> I think we only need to document the rules of the file naming scheme (in my previous comment) and the default sluggification.

All good, thank you! Let's leave the documentation for the end then.

-- 
Protesilaos Stavrou
https://protesilaos.com

Issue #339, opened 2024-05-07
`denote-file-prompt` does not seem to take `denote-file-history` into account


brabalan:
When I call a command that uses `denote-file-prompt`, I’m surprised that the history is not taken into account. Even though `denote-file-history` is not empty, the first file suggested is always the same one.

Suspicion: the call to `(completing-read prompt new-collection nil nil nil 'denote-file-history)` seems to use a `new-collection` with short (relative) names, whereas the history contains absolute file names. Could it be the reason why it’s not working?

protesilaos:
You mean that `M-p` in the minibuffer does not work? It seems correct on my end, but maybe I am missing something.

brabalan:
The problem may be that I’m using vertico…

When I for instance run `denote-open-or-create` I see this

<img width="882" alt="Screenshot 2024-05-07 at 14 02 44" src="https://github.com/protesilaos/denote/assets/1021637/e821007b-9ee7-4ad4-9ca6-6385eb63c3f2">

and this is unrelated to the history of files I accessed.

If I use `M-p` I indeed have the absolute file name in the prompt

<img width="971" alt="Screenshot 2024-05-07 at 14 05 27" src="https://github.com/protesilaos/denote/assets/1021637/3b9d6ff5-fb82-469e-bd4e-619237757fe1">

but it does not select the file (as the absolute name does not correspond to anything, since it mentions the path), and then I cannot access the note and denote proposes to create it.

Disabling vertico "solves" the issue in the sense that nothing is suggested, and `M-p` does end up finding the note. So I guess this bug report is more about the interaction between vertico and `denote-prompt`.

protesilaos:
I think the issue is with our prompt and how the history is recorded. How about this?

```diff
 denote.el | 10 ++++------
 1 file changed, 4 insertions(+), 6 deletions(-)

diff --git a/denote.el b/denote.el
index a500437..4035d2b 100644
--- a/denote.el
+++ b/denote.el
@@ -1196,12 +1196,10 @@ (defun denote-file-prompt (&optional files-matching-regexp prompt-text)
            (_ (when included-cpd
                 (setq substrings (cons "./" substrings))))
            (new-collection (denote--completion-table 'file substrings))
-           (relname (completing-read prompt new-collection nil nil nil 'denote-file-history))
-           (absname (expand-file-name relname common-parent-directory)))
-      ;; NOTE 2024-02-29: This delete and add feels awkward.  I wish
[...]

PR #342, opened 2024-05-10
Rework denote-file-prompt to work with relative paths in denote-file-history

jeanphilippegg:
This fixes issue #339, as discussed there.


jeanphilippegg:
I pushed a change, but it is ready to be merged.

protesilaos:
Merged. Thank you!

Issue #340, opened 2024-05-08
denote-link creates a button for non-org files


haji-ali:
Currently, `denote-link` creates a button around the link when the major mode is not `org-mode`. This is a temporary solution as it means that if the note is closed and re-opened the button disappears.

I am writing a font-lock function to enable fontifying the link and making it clickable, but the button that denote creates will interfere with it. It would be great if this button creation is controlled with an option, though I would argue that it should not be created at all.

protesilaos:
> From: Abdul-Lateef Haji-Ali ***@***.***>
> Date: Wed,  8 May 2024 14:40:58 -0700
>
> Currently, `denote-link` creates a button around the link when the
> major mode is not `org-mode`. This is a temporary solution as it means
> that if the note is closed and re-opened the button disappears.

Indeed, it disappears aferwards. Something like this re-enables it:

    (add-hook 'find-file-hook #'denote-link-buttonize-buffer)

> I am writing a font-lock function to enable fontifying the link and
> making it clickable, but the button that denote creates will interfere
> with it. It would be great if this button creation is controlled with
> an option, though I would argue that it should not be created at all.

Is this function going to do more than make the links clickable? In
other words, is denote-link-buttonize-buffer not suitable for your
needs?

Add an option is fine, though I am trying to understand the case better.

-- 
Protesilaos Stavrou
https://protesilaos.com

haji-ali:
I think implementing this feature through font-lock is more standard (and would allow automatic formatting if the link is edited). Also, I wanted to implement folding of links similar to org-mode (using `org-fold` or overlays).

Perhaps moving the button creation and hook into a minor-mode that can be enabled or disabled is best?

protesilaos:
> From: Abdul-Lateef Haji-Ali ***@***.***>
> Date: Thu,  9 May 2024 00:40:19 -0700
>
> I think implementing this feature through font-lock is more standard
> (and would allow automatic formatting if the link is edited). Also, I
> wanted to implement folding of links similar to org-mode (using
> `org-fold` or overlays).

Sounds good! If you do work on this, consider sending a patch or pull
request.

> Perhaps moving the button creation and hook into a minor-mode that can
> be enabled or disabled is best?

Fine by me. Though the :init-value must be non-nil to preserve the
current behaviour. But then users get the opt-out functionality.

-- 
Protesilaos Stavrou
https://protesilaos.com

protesilaos:
I forgot to close this. We are done here. Thank you!

PR #344, opened 2024-05-10
Implemented fontification for displaying links

haji-ali:
A follow-on from #340 

This pull request implements a custom user option `denote-link-display-style` and a minor mode `denote-fontify-links-mode` which controls how links are displayed; as buttons (default), fontified or fontified and folded. Setting `denote-link-display-style` on its own is not sufficient. If it is set to `button`, the function `denote-link-buttonize-buffer` should be called or added to a hook. If it's set to `fontify` or `fontify-and-fold` then the newly defined minor mode `denote-fontify-links-mode` needs to be enabled (enabling this mode also calls `denote-link-buttonize-buffer` when `denote-link-display-style` is `button`). I also had to change the regex for links for org and md to capture the link captions.

I am putting this as a draft, since I am fairly new to denote and I haven't studied it extensively, so there might be something I am doing wrong or some design decisions that need to be discussed.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -3826,12 +3826,14 @@ argument.
 Also see `denote-id-only-link-in-context-regexp'.")
 
 (defvar denote-org-link-in-context-regexp
-  (concat "\\[\\[" "denote:"  "\\(?1:" denote-id-regexp "\\)" "]" "\\[.*?]]")
+  (concat "\\[\\[" "denote:"  "\\(?1:" denote-id-regexp "\\)" "]" "\\["
+          "\\(?2:" ".*?" "\\)" "]]")
   "Regexp to match an Org link in its context.
 The format of such links is `denote-org-link-format'.")
 
 (defvar denote-md-link-in-context-regexp
-  (concat "\\[.*?]" "(denote:"  "\\(?1:" denote-id-regexp "\\)" ")")
+  (concat "\\[" "\\(?2:" ".*?" "\\)" "]"
+          "(denote:"  "\\(?1:" denote-id-regexp "\\)" ")")
   "Regexp to match a Markdown link in its context.
 The format of such links is `denote-md-link-format'.")
 
@@ -3946,9 +3948,7 @@ Also see `denote-link-with-signature'."
     (user-error "The linked file does not exist"))
   (let ((beg (point)))
     (denote--delete-active-region-content)
-    (insert (denote-format-link file description file-type id-only))
-    (unless (derived-mode-p 'org-mode)
-      (make-button beg (point) 'type 'denote-link-button))))
+    (insert (denote-format-link file description file-type id-only))))
 
 (define-obsolete-function-alias
   'denote-link-insert-link
@@ -4241,6 +4241,90 @@ To be assigned to `markdown-follow-link-functions'."
 (eval-after-load 'markdown-mode
   '(add-hook 'markdown-follow-link-functions #'denote-link-markdown-follow))
 
+;;;;; Link fontification
+(defvar denote-link-keymap
+  (let ((map (make-sparse-keymap)))
+    (define-key map [mouse-2] #'denote-link-open-at-mouse)
+    (define-key map [mouse-3] #'denote-link-open-at-mouse)
+    (define-key map [follow-link] #'mouse-face)
+    map)
+  "Keymap used in fontified denote links")
+
+(defun denote-link-open-at-mouse (ev)
+  "Open denote link after mouse click."
+  (interactive "e")
+  (mouse-set-point ev)
+  (when-let ((id (get-text-property (point) 'denote-link-id)))
+    (funcall denote-link-button-action (denote-get-path-by-id id))))
+
+(defun denote-fontify-links (&optional limit)
+  "Fontify denote links.
+
+Implementation based on `org-activate-links'"
+  (catch :exit
+    (when-let (type (denote-filetype-heuristics (buffer-file-name)))
+      (while (re-search-forward
+              (denote--link-in-context-regexp type)
+              limit t)
+        (save-match-data  ;; to return the matches to font-lock
+          (let* ((start (match-beginning 0))
+	         (end (match-end 0))
+	         (visible-start (match-beginning 2))
+	         (visible-end (match-end 2))
+                 (id (match-string-no-properties 1))
+                 (link (concat "file:" (denote-get-path-by-id id))))
+	    (unless (let ((face (get-text-property
+			         (max (1- start) (point-min)) 'face)))
+		      (if (consp face)
+                          (memq 'font-lock-comment-face face)
+		        (eq 'font-lock-comment-face face)))
+              (let* ((properties `(mouse-face highlight
+			                      keymap ,denote-link-keymap
+                                              denote-link-id ,id
+                                              help-echo ,(concat "denote:" id)
+			                      htmlize-link (:uri ,link)
+			                      font-lock-multiline t))
+                     (non-sticky-props
+                      '(rear-nonsticky (mouse-face highlight keymap invisible
+                                                   intangible help-echo
+                                                   htmlize-link)))
+                     (face-property 'link)
+                     (hidden (append '(invisible t) properties)))
+                (remove-text-properties start end '(invisible nil))
+                (add-text-properties start visible-start hidden)
+                (add-face-text-property start end face-property)
+		(add-text-properties visible-start visible-end properties)
+		(add-text-properties visible-end end hidden)
+                (dolist (pos (list end visible-start visible-end))
+                  (add-text-properties (1- pos) pos non-sticky-props)))
+              (throw :exit t))))))		;signal success
+    nil))
+
+(defun denote--get-link-url-at-point ()
+  "Return the denote link under point.
+To be used as a `thing-at' provider."
+  (when-let (id (get-text-property (point) 'denote-link-id))
+    (concat "file:" (denote-get-path-by-id id))))
+
+(define-minor-mode denote-fontify-links-mode
+  "A minor mode to fontify and fold denote links."
+  :init-value nil
+  :group 'denote
+  (if denote-fontify-links-mode
+      (progn (font-lock-add-keywords nil '(denote-fontify-links))
+             (with-eval-after-load 'thingatpt
+               (make-local-variable 'thing-at-point-provider-alist)
+               (add-to-list 'thing-at-point-provider-alist
+                            '(url . denote--get-link-url-at-point))))
+    (font-lock-remove-keywords nil '(denote-fontify-links))
+    (with-eval-after-load 'thingatpt
+      (set 'thing-at-point-provider-alist
+           (cl-remove
+            '(url . denote--get-link-url-at-point)
+            (symbol-value 'thing-at-point-provider-alist)
+            :test 'equal))))
+  (font-lock-update))
+
 ;;;;; Backlinks' buffer
 
 (define-button-type 'denote-link-backlink-button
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you @haji-ali!

I will review this on Sunday.

Two quick notes:

- This change is more than ~15 lines of code. Have you assigned copyright to the Free Software Foundation? If yes, we are all good. If not, you need to do it before I can merge this. Is that okay?
- Do you think we really need the `denote-link-display-style`? I think that having a minor mode is basically an opt-in/opt-out mechanism. Adding an option on top may cause confusion.

haji-ali:
> * This change is more than ~15 lines of code. Have you assigned copyright to the Free Software Foundation? If yes, we are all good. If not, you need to do it before I can merge this. Is that okay?

Yes, I've done this before when I contributed to AUCTeX.

> * Do you think we really need the `denote-link-display-style`? I think that having a minor mode is basically an opt-in/opt-out mechanism. Adding an option on top may cause confusion.

I agree about the possibility for confusion. But I think we need an option to allow the three modes of highlighting links (buttons as currently done, font-lock and font-lock with folding).

Unless you're suggesting that we set the value of the minor mode to be one of these options. I think this probably could be done, but I've never seen it before.

denote.el
```diff
@@ -3933,6 +3937,143 @@ To be assigned to `markdown-follow-link-functions'."
...
+          (const :tag "Fontified" fontify)
+          (const :tag "Fontified and folded" fontify-and-fold))
+  :package-version '(denote . "0.6.0")
+  :group 'denote)
+
+(defvar denote-link-keymap
+  (let ((map (make-sparse-keymap)))
+    (define-key map [mouse-2] #'denote-link-open-at-mouse)
+    (define-key map [mouse-3] #'denote-link-open-at-mouse)
+    (define-key map [follow-link] #'mouse-face)
```
protesilaos, line 3967: Is the `mouse-face` a function? I cannot find it on my end (Emacs 30).

denote.el
```diff
@@ -3933,6 +3937,143 @@ To be assigned to `markdown-follow-link-functions'."
...
+                 (id (match-string-no-properties 1))
+                 (link (concat "file:" (denote-get-path-by-id id))))
+	    (unless (let ((face (get-text-property
+			         (max (1- start) (point-min)) 'face)))
+		      (if (consp face)
+                          (memq 'font-lock-comment-face face)
+		        (eq 'font-lock-comment-face face)))
+	      (when (and (featurep 'org-fold)
+                         (eq denote-link-display-style 'fontify-and-fold))
+                (org-fold-region start end nil 'org-link)
```
protesilaos, line 4001: The byte compiler is warning me that `org-fold` requires Emacs 29 to work. Is this correct? We now depend on Emacs 28 and I hope we keep it this way for the foreseeable future (let's say until the next Debian stable, which is about a year away).

denote.el
```diff
@@ -3933,6 +3937,143 @@ To be assigned to `markdown-follow-link-functions'."
...
+                ;; We are folding the whole emphasized text with 'org-link
+                ;; first.  It makes everything invisible (or whatever
+                ;; the user wants).
+                (org-fold-region start end t 'org-link)
+                ;; The visible part of the text is folded using
+                ;; 'org-link-description, which is forcing this part of
+                ;; the text to be visible.
+                (org-fold-region visible-start visible-end t
+                                 'org-link-description)
+                (org-rear-nonsticky-at visible-start)
```
protesilaos, line 4012: A lot of this code is specific to Org which is a problem for us in practice because `denote.el` does not `(require 'org)`.  Of course, we still have the blocking issue with `org-fold` that I mentioned above.

denote.el
```diff
@@ -4158,7 +4299,10 @@ inserts links with just the identifier."
              (beg (point)))
         (progn
           (denote-link--insert-links files file-type id-only)
-          (denote-link-buttonize-buffer beg (point)))
+          (when (and
+                 denote-fontify-links-mode
+                 (eq denote-link-display-style 'button))
```
protesilaos, line 4304: I think that since we are doing this change, it is an opportunity for us to remove the old functionality. The button was never meant to be there long-term. It was an "okay" solution to have links that are actionable, but using them is not optimal as you already suggest.

Given that `org-fold` is a constraint, maybe we can simplify all this and not offer any option at all? Just go with the `fontify` approach. Then the user can enable the minor mode to activate the links (or we can even make its `:init-value t` if we think that is appropriate).

protesilaos (commented):
I am taking a closer look at this now. I could not do it on Sunday because of the issues we had with the `denote-file-prompt`.

denote.el
haji-ali: You're right. It's not defined on mine either. This keymap is mirroring one that I got from `org` (you can check `org-mouse-map` on your setup).

denote.el
haji-ali: I am not sure, but note that I only enable the folding features when `org-fold` is available as a feature.

denote.el
haji-ali: I can see that denote already has several calls to `(org-*) functions, `denote-link-ol-*` and `denote-date-prompt`. In any case, the code is protected with a `featurep` check for `org-fold`.

denote.el
haji-ali: I would also remove the button if backward compatibility is not important to you in this aspect. If you agree with keeping folding, then we can introduce a simple boolean variable to control its behavior. 
Otherwise, the code for `org-fold` can be removed and fontification would be kept. I am happy to make this change, let me know if I should also remove any obsolete code for buttons as well.

haji-ali:
@protesilaos, I've removed the dependence on `org-fold`, removed the extra option (now always fontifying and folding), and removed the old function calls which added buttons. 

There is some old code related to buttons which I haven't removed yet, in case you want to keep it for backward compatibility.

denote.el
haji-ali: See [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Clickable-Text.html) for an explanation of `mouse-face`.

protesilaos:
Thank you! I will review it tomorrow morning.

protesilaos:
Thank you @haji-ali! I have just merged your patch. I made some further tweaks on top of it. I also deprecated the old code we had, as I think we do not need it going forward.

Note that I saved your PR as a patch file and installed it locally. Then I made changes on top and pushed everything. But GitHub does not seem to recognise this. I don't know if there is a way to mark this as "Merged". Your commit is here:

```
commit 4d1245223c726ef4e740a10a619e264e1ffa20bd
Author: Al Haji-Ali <a.hajiali@hw.ac.uk>
Date:   Fri May 10 14:49:06 2024 +0100

  Implemented a fontification method for displaying links and removed buttons

 denote.el | 101 +++++++++++++++++++++++++++++++++++++++++++++++++++++++-------
 1 file changed, 91 insertions(+), 10 deletions(-)
```

protesilaos:
Just to note that we have these two to still deal with:

```patch
commit ce1f59af6a06fd3c63ec8258c7083bfebbda92ad
Author: Protesilaos Stavrou <info@protesilaos.com>
Date:   Wed Jun 19 10:40:02 2024 +0300

    Add TODO and FIXME for new link fontification mode
---
 denote.el | 7 +++++++
 1 file changed, 7 insertions(+)

index 1ae00a0..f8e6c77 100644
--- a/denote.el
+++ b/denote.el
@@ -4191,6 +4191,10 @@ (eval-after-load 'markdown-mode
   '(add-hook 'markdown-follow-link-functions #'denote-link-markdown-follow))
 
 ;;;;; Link fontification
+
+;; TODO 2024-06-19: We need to bind RET and maybe even C-c C-o to a
+;; command that opens the link at point.  Then we may also rename this
+;; keymap.
 (defvar denote-link-mouse-map
   (let ((map (make-sparse-keymap)))
     (define-key map [mouse-2] #'denote-link-open-at-mouse)
@@ -4261,6 +4265,9 @@ (defun denote--get-link-file-path-at-point ()
 
 (defvar thing-at-point-provider-alist)
 
+;; FIXME 2024-06-19: We are missing a function that clean up all those
+;; properties when the mode is disabled.  Otherwise, we are left with
+;; invisible text, which looks broken.
 (define-minor-mode denote-fontify-links-mode
   "A minor mode to fontify and fold Denote links."
   :init-value nil
```

haji-ali:
> Note that I saved your PR as a patch file and installed it locally. Then I made changes on top and pushed everything. But GitHub does not seem to recognise this. I don't know if there is a way to mark this as "Merged". Your commit is here:

I am also not sure, but I can just close this PR and rebase my fork to your changes.

> ```diff
> +;; TODO 2024-06-19: We need to bind RET and maybe even C-c C-o to a
> +;; command that opens the link at point.  Then we may also rename this
> +;; keymap.
> ```

I am not sure this is needed or advised (at least by default). The reason for the `thingat` integration is that typically a more generic function/global bindings (such as `org-open-at-point-global`) opens links and this would work with this fontification. This is the approach that is followed in `org-mode` at least and that I was trying to replicated. I also believe RET should insert a new line, even if inside a link.

> ```diff
> +;; FIXME 2024-06-19: We are missing a function that clean up all those
> +;; properties when the mode is disabled.  Otherwise, we are left with
> +;; invisible text, which looks broken.
> ```
Are you referring to the text properties? When the mode is disabled the font-lock keyword is removed and `font-lock-update` is called, so thef properties won't be added.

> [Make denote-fontify-links-mode not do anything in Org mode](https://github.com/protesilaos/denote/commit/5dcbf34feac371071687c0c99179cce122cb0f68)
protesilaos

I personally would prefer documenting that the mode should simply not be enabled in `org-mode`, or issue a warning.

protesilaos:
> From: Abdul-Lateef Haji-Ali ***@***.***>
> Date: Thu, 20 Jun 2024 03:46:14 -0700
>
>> Note that I saved your PR as a patch file and installed it locally.
>> Then I made changes on top and pushed everything. But GitHub does not
>> seem to recognise this. I don't know if there is a way to mark this
>> as "Merged". Your commit is here:
>
> I am also not sure, but I can just close this PR and rebase my fork to your changes.

Sure!

>> ```diff
>> +;; TODO 2024-06-19: We need to bind RET and maybe even C-c C-o to a
>> +;; command that opens the link at point.  Then we may also rename this
>> +;; keymap.
>> ```
>
> I am not sure this is needed or advised (at least by default). The
> reason for the `thingat` integration is that typically a more generic
> function/global bindings (such as `org-open-at-point-global`) opens
> links and this would work with this fontification. This is the
> approach that is followed in `org-mode` at least and that I was trying
> to replicated. I also believe RET should insert a new line, even if
> inside a link.

We cannot assume that the user will opt for 'org-open-at-point-global'.
There are those who do not want to depend on Org in their workflow.

Before, we had the buttonisation bind RET to visit a link. This is why I
thought we would want to keep this. Personally, I do not mind having
another key, such that RET always inserts a newline.

>> ```diff
>> +;; FIXME 2024-06-19: We are missing a function that clean up all those
>> +;; properties when the mode is disabled.  Otherwise, we are left with
>> +;; invisible text, which looks broken.
>> ```
> Are you referring to the text properties? When the mode is disabled
> the font-lock keyword is removed and `font-lock-update` is called, so
> thef properties won't be added.

I have not checked it again, but in my test when I would disable the
mode the link would not be unfolded/unhidden. I tried it in a txt file.

>> [Make denote-fontify-links-mode not do anything in Org mode](https://github.com/protesilaos/denote/commit/5dcbf34feac371071687c0c99179cce122cb0f68)
> protesilaos
>
> I personally would prefer documenting that the mode should simply not be enabled in `org-mode`, or issue a warning.

Is this safe though? My concern is with the scenario where the user adds
the mode via the text-mode-hook, which applies to Org as well. Will our
fontification have any weird interaction with that of Org? If not, then
covering it only in the docs is fine.

-- 
Protesilaos Stavrou
https://protesilaos.com

haji-ali:
```
We cannot assume that the user will opt for 'org-open-at-point-global'.
There are those who do not want to depend on Org in their workflow.
```

`org-open-at-point-global` was just an example for a function that uses `thing-at-point 'url` to get a url and navigate to it. In any case, I see the point of adding a key-binding in the local keymap, but I personally would probably remove the mapping in my config since I rarely want to shadow key-bindings with text properties (I would find it too unpredictable).

Closing this issue as the other points are being discussed in #383

PR #348, opened 2024-05-14
Make denote-file-prompt return an absolute path

jeanphilippegg:
As explained in issue #339, I kept the `no-require-match`, but changed it back to using a single function.


protesilaos:
Merged. Thank you!

Issue #284, opened 2024-03-09
A `denote-filetype-heuristics` variant to help other code


mentalisttraceur:
As I was trying to generalize some of my code to share in #280 (so that it's not narrowly limited to just my use cases) I found myself wanting to check if a file path was one of the types that Denote recognizes.

I found/remembered `denote-filetype-heuristics` and thought "yes! this is what I-"

```lisp
(denote-filetype-heuristics "~/test.png")
;;  returns 'org
```

"...huh."

I then looked closer and I guess it's intended for when you already know that you are treating a file as a note, and you just need to decide on the type.

But external code might sometimes want to reuse these heuristics to ask "is this file of a type that Denote recognizes as a possible note? (and if so, which type?)", and in that case, it's more helpful if we had a function that:

1. in the absence of a match, returned `nil` rather than using the first entry of `denote-file-types`, and

2. that only ever operates on the file given, regardless of what buffer is current when it's called.

So I propose either a separate function, or a boolean parameter to `denote-filetype-heuristics`:

```diff
-(defun denote-filetype-heuristics (file)
+(defun denote-filetype-heuristics (file &optional match-only)
-  (if (denote--file-type-org-capture-p)
+  (if (and (not match-only)
+           (denote--file-type-org-capture-p))
       'org
     (let* ((extension (denote-get-file-extension-sans-encryption file))
            (types (denote--file-types-with-extension extension)))
       (cond ((not types)
-             (caar denote-file-types)))
+             (unless match-only
+                 (caar denote-file-types)))
-            ((= (length types) 1)
-             (caar types))
             (t
              (or (car (seq-find
                        (lambda (type)
                          (denote--regexp-in-file-p (plist-get (cdr type) :title-key-regexp) file))
                        types))
-                 (caar types)))))))
+                 (unless match-only
+                     (caar types)))))))
```

(I just eliminated the length= 1 case to the cond because when `match-only` is true, we'd still need to check if that sole element matches - falling through to the `(or (seq-find ...) ...)` works in all cases.)

mentalisttraceur:
Another way to look at this is as an a complement (and for some uses, a better fitting alternative) to `denote-file-is-note-p`.

Because given that this file type heuristic matching both checks the extension and looks inside the file if the extension matches, in some ways it does more to determine if a file actually looks like a Denote note.

mentalisttraceur:
Taking the separate function alternative in my own config for now (actually might be better this way! I think it turned out a lot easier to follow):

[...]

PR #350, opened 2024-05-16
Various cleanup commits and bug fixes

jeanphilippegg:
Just a few commits to make the code more robust and fix some bugs I discovered:

- `denote--directory-get-files` would consider a directory "20240505T010101/"
  as a valid note.

- Renaming a file if `denote-file-types` is nil would throw an error. In
  general, we cannot assume that the file-type is non-nil as
  `denote-file-types` may be nil.

- As discussed in #284, I have reviewed `denote-filetype-heuristics` and we
  should not return `org` when the file type is not found. We should just
  return `nil`. I checked all the places it is used and made sure it was okay
  to do so. #284 can be closed.

- Most of the time, we don't actually need to use `denote-file-is-note-p` and
  instead should use `denote-filename-is-note-p`. This will help when a user
  wants to rename a new note that has not yet been saved, for example. This is
  not yet fully supported, but this pull request is an improvement. I have
  reviewed the impact of this in the places I have made the change.


denote.el
```diff
@@ -1744,28 +1742,32 @@ Subroutine of `denote--file-with-temp-buffer'."
 (defun denote-retrieve-front-matter-title-value (file file-type)
   "Return title value from FILE front matter per FILE-TYPE."
   (denote--file-with-temp-buffer file
-    (when (re-search-forward (denote--title-key-regexp file-type) nil t 1)
+    (when (and file-type
```
protesilaos, line 1745: Is it not better to have the check for the `file-type` before the `denote--file-with-temp-buffer`?

denote.el
```diff
@@ -2508,23 +2510,19 @@ If more than one file type correspond to this file extension, use the
 first file type for which the :title-key-regexp in `denote-file-types'
 matches in the file.
 
-If no file type in `denote-file-types' has the file extension,
-the file type is assumed to be the first one in `denote-file-types'."
+Return nil if the file type is not recognized."
```
protesilaos, line 2513: I think the reason we did this was because we ultimately needed some way to, e.g., format links. Org links can work anywhere in Emacs with `org-open-at-point-global`.

protesilaos:
I made some comments for clarity, but I already merged this. Thank you!

denote.el
jeanphilippegg: You are right! I will fix this in my next pull request.

denote.el
```diff
@@ -1015,7 +1015,8 @@ are not backups."
    #'expand-file-name
    (seq-filter
     (lambda (file)
-      (and (denote-file-has-identifier-p file)
+      (and (file-regular-p file)
+           (denote-file-has-identifier-p file)
```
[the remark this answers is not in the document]
jeanphilippegg: I don't think `file-regular-p` should be part of `denote-file-has-identifier-p` because I do not think the latter should be doing any extra work or checking. And actually, `denote-file-has-identifier-p` may be called with a file name that does not (yet) exists. This scenario is possible.

denote.el
jeanphilippegg: You mean that a non-org file can create links of org type? Right now (and also before this PR), `denote-link` would not allow that anyway. I think if we want to allow that eventually, we should add this support in the linking commands directly (and keep `denote-filetype-heuristics` simply returning the type of the file given to it in its parameter).

PR #351, opened 2024-05-17
Other fixes

### LINKED ISSUE #279: Enhancing `denote-rename-file` to handle not-yet-saved visiting buffer
I use `denote-rename-file` a lot for changing information that shows up in a Denoted file name.

Sometimes, I'll create a note, do some writing, and only then I'll think of some change to the title or the tags. But at that moment I often haven't saved the note yet and don't even know if I'm going to keep it - I don't want to save it just to change the relevant information.

So I think it would be a nice addition if `denote-rename-file` would work on files that don't exist, but for which a visiting buffer exists, doing everything except writing the file out to disk.

---

Architecturally, to me it makes a whole lot of sense, especially when using/considering Denote's public functions as building blocks to build even more good things on top of Denote's goodness, that

1. there might reasonably be other possible use cases in things people build on top of Denote, for manipulating information that shows up in a Denoted file name before the first save of that file (which may or may not be a note), and
2. having a single entry point to manipulate parts of a Denoted file name is very convenient. I take the existence of `denote-rename-file`, and the upcoming/mentioned change of making that function respect the prompts variable to control what it prompts for when used interactively, as agreement. (Consider the combinatoric explosion otherwise: right now I only have to learn+understand and wrap+advise one function - `denote-rename-file` - to get all the behavior that I want.)

---

My first solution to this was to have the following code in my `denote-rename-file` wrapper to make this happen. It was crude, but it was quick, simple, didn't require me to know or couple any implementation details, and it works "well enough" - it temporarily creates the file if it doesn't exist before running `denote-rename-file` and then deletes it right after:

```lisp
            (if (file-exists-p path)
                (denote-rename-file path title keywords signature)
                (with-current-buffer (find-file-noselect path)
                    (write-file-no-visit path))
                (let ((new-path (denote-rename-file path title keywords signature)))
                    (delete-file new-path)
                    (refresh-modified-state)
                    new-path))))
```

But obviously, it would be cleaner to just operate on the visited buffer without touching the file system.

Currently I use this more invasive monkey-patching to achieve that:

```lisp
    (defvar fixed-denote-rename-file--missing nil)

    (defun hack-rename-file (rename-file &rest arguments)
        (condition-case error
            (apply rename-file arguments)
            (file-missing
                (setq fixed-denote-rename-file--missing t))))

    (defun hack-denote--file-regular-writable-p
            (denote--file-regular-writable-p file)
        (if fixed-denote-rename-file--missing
            (find-buffer-visiting file)
            (funcall denote--file-regular-writable-p file)))

    (defun fixed-denote-rename-file (denote-rename-file &rest arguments)
        (with-advice ('rename-file :around 'hack-rename-file
                      'denote--file-regular-writable-p
                          :around 'hack-denote--file-regular-writable-p)
            (let ((fixed-denote-rename-file--missing nil))
                (apply denote-rename-file arguments))))

    (advice-add 'denote-rename-file :around 'fixed-denote-rename-file)
```

---

I suspect in-package support would be cleaner, simpler code that's easier to follow and maintain. If there's agreement that it's worth adding in-package, I can draft a full patch/PR.

jeanphilippegg:
In this PR, I have addressed your comment in my previous PR and introduced a
few more cleanups.

The changes will help with resolving #279 to allow renaming new notes that are still
unsaved.

I also added checks to creation and renaming commands to avoid creating a note
if the file already exists. This is not a concern for now because the id is
unique and part of the file name, but we may want to make it optional
eventually.

Just a comment regarding your last commit: The change is not only stylistic:
`re-search-forward` behave differently whether `NOERROR` is `t` or any other
non-nil value such as `:no-error`. Check its docstring. I have discovered this
myself as well while working on the upcoming new fontification. Same thing
with `completing-read`: `t` and `:require-match` do not have the same
behaviors.

I have reverted your commit myself as it could conflict with a change I needed
to make to address one of your comment on my previous pull request.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -2548,17 +2548,19 @@ variable `denote-directory'."
 (defun denote-rename-file-and-buffer (old-name new-name)
   "Rename file named OLD-NAME to NEW-NAME, updating buffer name."
   (unless (string= (expand-file-name old-name) (expand-file-name new-name))
-    (cond
-     ((derived-mode-p 'dired-mode)
-      (dired-rename-file old-name new-name nil))
-     ;; NOTE 2024-02-25: The `vc-rename-file' requires the file to be
-     ;; saved, but our convention is to not save the buffer after
-     ;; changing front matter unless we absolutely have to (allows
-     ;; users to do `diff-buffer-with-file', for example).
-     ((and denote-save-buffers (not (buffer-modified-p)) (vc-backend old-name))
-      (vc-rename-file old-name new-name))
-     (t
-      (rename-file old-name new-name nil)))
+    (when (and (denote--file-regular-writable-p old-name)
+               (file-writable-p new-name))
+      (cond
+       ((derived-mode-p 'dired-mode)
+        (dired-rename-file old-name new-name nil))
+       ;; NOTE 2024-02-25: The `vc-rename-file' requires the file to be
+       ;; saved, but our convention is to not save the buffer after
+       ;; changing front matter unless we absolutely have to (allows
+       ;; users to do `diff-buffer-with-file', for example).
+       ((and denote-save-buffers (not (buffer-modified-p)) (vc-backend old-name))
+        (vc-rename-file old-name new-name))
+       (t
+        (rename-file old-name new-name nil))))
     (when-let ((buffer (find-buffer-visiting old-name)))
       (with-current-buffer buffer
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you!

I had not realised the subtleties of non-nil for `re-search-forward` and `completing-read`. Good to know!

PR #352, opened 2024-05-17
Make renaming commands rename the current file without prompt

jeanphilippegg:
I propose that `denote-rename-file` and related do not prompt if we are
editing a file and directly use the current file as the target of the renaming
right away for a better user experience and also to complete the solution of
#279 (allow renaming new notes that have not yet been saved). We already avoid
this prompting when we are in Dired. I think maybe it was also the original
intention: prompt only as a fallback.

What do you think?

If we want, we could provide documentation to get the old behavior back in the
manual, though I would not see why someone would want to call
`denote-rename-file` out of context in a random location that is not a file.

### DIFF
--- a/README.org
+++ b/README.org
@@ -1483,8 +1483,9 @@ prepends new front matter to the top of the file.  We do this as a
 safety precaution since the user can, in principle, add arbitrary
 extras to their front matter that we would not want to touch.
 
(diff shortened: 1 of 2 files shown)


protesilaos:
Merged, thank you!

I think this is better overall. If someone asks, then we can document it. I don't think it is a major workflow change.

PR #358, opened 2024-05-18
Fix issue on note creation caused by commit bce7778

### LINKED ISSUE #357: Problem when creating new notes
After calling `denote` to create new note, and going through all the prompt,
the following error shows up
`denote: Wrong type argument: stringp, #<buffer 20240518T010559--aien__testing.org>`
And the resulting file was lacking frontmatter and not being saved despite having `denote-save-buffers` set to `t` 

The problem stops after reverting bce7778

jeanphilippegg:
I made this pull request to fix my mistake as reported in #357. Sorry about this!

I may add another commit soon. I feel like there is something awkward with the
way we handle the title and the renaming convenience commands. For example,
`denote-rename-file-signature` called on a file that is not yet a Denote file.
This is a case I may not have considered in my refactoring.

Anyway, don't wait to merge this current pull request.

resolves #357

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1922,8 +1922,8 @@ TEMPLATE, and SIGNATURE should be valid for note creation."
                   title (denote--date date file-type) keywords
                   id
                   file-type)))
-    (when (file-regular-p buffer)
-      (user-error "A file named `%s' already exists" buffer))
+    (when (file-regular-p path)
+      (user-error "A file named `%s' already exists" path))
     (with-current-buffer buffer
       (insert header)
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
Actually, never mind my previous comment. I checked out the code from a few
weeks ago and tested `denote-rename-{add/remove}-signature` and they also
duplicate the id if we cannot find a title. Same thing with
`denote-dired-rename-marked-files-with-keywords`.

protesilaos:
Merged, thank you!

PR #359, opened 2024-05-18
Rework the fontification

jeanphilippegg:
This is a proposition for a new fontification. It works with our current file
names, but it also supports the future file names that can have permuted file
name components. It works in Dired and the backlinks buffer.

For now, titles always need to be prefixed with their delimiter (`--`) even if
they are the first component. I prefer keeping options open for now.

Also the code for the various matchers are all similar. I prefered keeping
them separate for now, even if it meant some code duplication.

You can give it a try!

Whenever we are ready and satisfied with the result, we can merge it, even if
we don't have the code for the permutation of file name components yet.

Here is an image of the results:

fontification

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -3323,37 +3323,131 @@ and seconds."
   :group 'denote-faces
   :package-version '(denote . "2.1.0"))
 
-(defvar denote-faces--file-name-regexp
-  (concat "\\(?11:[\t\s]+\\|.*/\\)?"
-          "\\(?1:[0-9]\\{4\\}\\)\\(?12:[0-9]\\{2\\}\\)\\(?13:[0-9]\\{2\\}\\)"
-          "\\(?10:T\\)"
-          "\\(?2:[0-9]\\{2\\}\\)\\(?14:[0-9]\\{2\\}\\)\\(?15:[0-9]\\{2\\}\\)"
-          "\\(?:\\(?3:==\\)\\(?4:[^.]*?\\)\\)?"
-          "\\(?:\\(?5:--\\)\\(?6:[^.]*?\\)\\)?"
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you!

> Whenever we are ready and satisfied with the result, we can merge it, even if
we don't have the code for the permutation of file name components yet.

I wonder what is your idea for the first field in this scenario. What you show in the picture above preserves the field separator, though this is not consistent with how we do it with the identifier. Maybe the plan is to have those without the separator? Of course, we do not need to discuss this here and now. Just mentioning it because I noticed it now.

jeanphilippegg:
The idea for now is to always keep the delimiter, except only for the identifier which can drop its delimiter. The reason is that in a file named "some-word@@id.org", what is "some-word"? Is it a title? A keyword? A signature perhaps?

We could arguably drop the delimiter for only one component (the title probably would be best), but we cannot for all components. That is what I meant above when I said that I wanted to keep options open for now. Let's keep the delimiter for all components and we may choose which one is allowed to drop it later (if we go this way).

protesilaos:
> Date: Sat, 18 May 2024 10:33:31 -0700
>
> The idea for now is to always keep the delimiter, except only for the
> identifier which can drop its delimiter. The reason is that in a file
> named "some-word@@id.org", what is "some-word"? Is it a title? A
> keyword? A signature perhaps?
>
> We could arguably drop the delimiter for only one component (the title
> probably would be best), but we cannot for all components. That is
> what I meant above when I said that I wanted to keep options open for
> now. Let's keep the delimiter for all components and we may choose
> which one is allowed to drop it later (if we go this way).

It would indeed be ambiguous. Keeping the delimiter is fine in this
case.

-- 
Protesilaos Stavrou
https://protesilaos.com

Issue #287, opened 2024-03-11
I use `denote-rename-file` as my main file renaming UI


mentalisttraceur:
This issue just describes a big part of how I use Denote.

I'm using `denote-rename-file`\* as my primary file renaming interface, across *all* my files (and directories).

\* actually, a similar but more flexible function, implemented using a mix of other Denote functions, a couple functions+patches I've suggested in other Denote issues, and a few of my own helpers.

`whatever.txt` or `foo.c`? With #280, that's a Denoted file with just a title, so if I want to rename it I just hit my 'change title' keybind.

I have a separate keybind for each of:

* changing title,
* changing datetime,
* changing signature,
* adding tags (keywords), and
* removing tags.

If I'm in a Dired buffer, those keybinds act on `(dired-get-filename)`; if I'm in a file-visiting buffer, they act on that file.

What I really like about this is the orthogonality. I have the same keybinds and versatile Denoted naming for everything. With #279, it's strictly superior to Emacs' out-of-the-box rename commands:

* it prompts for confirmation in a nice way (but can be told to not with a `let` form in specific cases),
* if the file exists it renames it (but it works even if there's no file and just a visiting buffer),
* if a visiting buffer exists it updates the visited path and buffer name (but it works even if the file isn't open so long as the file exists),
* it refreshes any Dired buffers of that file's directory (in my function, I use `let` around the `denote-update-dired-buffers` call to set `denote-directory` to the file's directory, to benefit from the Dired updates upon rename in all directories),
* if I am calling it in Dired, my cursor stays on the renamed file after the Dired buffers get updated (thanks to #183).

The only file name changes I can't cover this way are changing the extension and names which must be incompatible with Denoted naming (`__init__.py`, files whose name start with a `.`, and so on). *...or can't I?*

I have a final sixth keybind prompts for the full name and uses that exactly as I typed it for the actual rename. If the file is a Denote note, it still updates the front matter according to the Denoted fields in the new file name.

Edit: I also have logic to handle one leading ".", so I can use all this with Unix hidden files.

So that way I achieve a 100% consistent interface for all file rename operations. The happy common case is structured renaming of Denoted file names, and there's a general fallback to handle the uncommon cases. There's no predication or data dependency on either the file type or if the file exists or is open in a buffer or both, or if I'm in the file's buffer or on it in Dired - the path from intent to muscle memory is uniform and purely depends on how I want to change the name. It all works on any file, and if the file looks like a note it also does the right thing with the front matter.

[...]

Issue #277, opened 2024-03-07
`denote-file-is-note-p`, but `t` if file exists *or* buffer visiting file exists


mentalisttraceur:
TL;DR: I propose adding:

```lisp
(defun denote-file-or-visiting-buffer-is-note-p (file)
    "Return non-nil if FILE is an actual Denote note
(as determined by `denote-file-is-note-p'), or if
there is a buffer visiting FILE and FILE would be
an actual Denote note once that buffer is saved."
    (and (not (file-directory-p file))
         (or (file-regular-p file)
             (find-buffer-visiting file))
         (denote-filename-is-note-p file)))
```

At least in some use-cases, when checking if a file is a note, notes which are merely not-yet-saved should count as notes.

After all, that *is* a note file, conceptually. I know it's a note file, Emacs knows it's a note file, only the file system doesn't know it's a note file.

For example, I have a function which grabs the title either from the front matter or from the file name (but *without* deslugifying the title, because that makes more sense to me), and I noticed some code was wrongly using the filename title instead of the front matter, if I just created a note but hadn't saved it yet. So now I'm using the above function instead of either is-filename-p or is-file-p there, to hit that sweet spot of intuitive/expected/reasonable (to me anyway) behavior.

mentalisttraceur:
By the way, I don't think you need the `(not (file-directory-p file))` line - I left it in for consistency and because maybe you want to make it more explicit, but to my knowledge file-regular-p and file-directory-p are always mutually exclusive.

jeanphilippegg:
Though this function could be useful, I don't think that we need it in Denote's code. In many places, we now use `denote-filename-is-note-p` instead of `denote-file-is-note-p` because we don't need to check that the file actually exists. I don't think we need to test that the file is visited in a buffer.

Maybe your original intention was to handle files that are still not saved? The code should now handle this better.

Would you consider this issue done (in a way)?

mentalisttraceur:
@jeanphilippegg yeah I'm happy to close this.

The use-case I had for this is currently handled just fine by existing Denote code and the `denote-file-note-type` function suggested in #284 :

```lisp
    (defun denoted-title-get (path)
        (let ((note-type (denote-file-note-type path)))
            (if note-type
                (denote-retrieve-title-value path note-type)
                (let ((name (file-name-nondirectory path)))
                    (when (string-prefix-p "." name)
                        (setq path (substring name 1))))
                (if (denote-file-has-identifier-p path)
[...]

PR #360, opened 2024-05-18
Add `denote-file-name-components-order` to allow reordering

### LINKED ISSUE #332: What should be the delimiter of identifiers in file names?
Ahead of the change that makes it possible to have file name components in another order, I would like to have opinions on what should be the delimiter of identifiers when it does not appear as the first component of the file name.

In my fork of the code, I have reused the signature delimiter (==), but I am not sure that reusing an existing delimiter is a good idea. I am thinking of using ++ instead. It is already sluggified away by default. Any objections? Ideas?

For example, we could have `My-note==sig1__emacs_philosophy++20240303T030303.org`.

jeanphilippegg:
This is a small pull request, but the whole feature is there!

I made `denote-file-name-components-order` a defvar because I do not know how to make it a defcustom. All file name components must be present.

This closes #332. (The delimiter of identifiers is @@.)

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -833,20 +833,17 @@ The removal is done only if necessary according to COMPONENT."
 
 (defun denote--replace-consecutive-token-characters (str component)
   "Replace consecutive characters with a single one in STR.
-Hyphens, underscores and equal signs are replaced with a single
-one in str, if necessary according to COMPONENT."
-  ;; -- are allowed in titles
-  (if (eq component 'title)
+Hyphens, underscores, equal signs and at signs are replaced with
+a single one in str, if necessary according to COMPONENT."
+  (let ((str (replace-regexp-in-string
+              "_\\{2,\\}" "_"
+              (replace-regexp-in-string
+               "=\\{2,\\}" "=" str))))
+    ;; -- are allowed in titles
+    (if (eq component 'title)
+        str
       (replace-regexp-in-string
-       "_\\{2,\\}" "_"
-       (replace-regexp-in-string
-        "=\\{2,\\}" "=" str))
-    (replace-regexp-in-string
-     "-\\{2,\\}" "-"
-     (replace-regexp-in-string
-      "_\\{2,\\}" "_"
-      (replace-regexp-in-string
-       "=\\{2,\\}" "=" str)))))
+       "-\\{2,\\}" "-" str))))
 
 (defun denote-sluggify (component str)
   "Make STR an appropriate slug for file name COMPONENT.
--- a/denote.el
+++ b/denote.el
@@ -303,6 +303,32 @@ Also see `denote-history-completion-in-prompts'."
                      (const :tag "Template" template)
                      (const :tag "Signature" signature))))
 
+(defvar denote-file-name-components-order '(identifier signature title keywords)
+  "Specify the order of the file name components of new notes.
+
+The value is a list of the following symbols:
+
+- `identifier': An identifier in a file name looks like
+  \"20240101T111111\" if it appears as the first component.
+  Else, it is prepended with \"@@\" if it appears anywhere else
+  in the file name.
+
+- `signature': A signature in a file name is an arbitrary string
+  that combines \"==\" with the signature of
+  `denote-signature-prompt'.
+
+- `title': A title in a file name is an arbitrary string that
+  combines \"--\" with the title of `denote-title-prompt'.
+
+- `keywords': Keywords in a file name are represented as a string
+  that combines \"__\" with the keywords of
+  `denote-keywords-prompt' separated with an underscore.
+
+All the symbols above must appear exactly once.  Any symbol
+missing will be added automatically.
+
+See also `denote-prompts'.")
+
 (defcustom denote-sort-keywords t
   "Whether to sort keywords in new files.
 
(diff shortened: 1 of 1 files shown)


denote.el
```diff
@@ -825,35 +851,34 @@ leading and trailing hyphen."
...
-      "_\\{2,\\}" "_"
+Hyphens, underscores, equal signs and at signs are replaced with
+a single one in str, if necessary according to COMPONENT."
+  (let ((str (replace-regexp-in-string
+              "_\\{2,\\}" "_"
+              (replace-regexp-in-string
+               "=\\{2,\\}" "="
+               (replace-regexp-in-string
+                "@\\{2,\\}" "@" str)))))
+    ;; -- are allowed in titles
```
protesilaos, line 870: Are these allowed unconditionally? I just tried it and the consecutive dashes were not in the file name (our current behaviour). Anyway, the consecutive dashes should be opt-in.

protesilaos:
Merged, thank you! I just made a small comment about the double dash in titles, but it is not a blocking issue.

protesilaos:
I just noticed that the fontification does not work in one case:

Screenshot_2024-05-19_08-22-53

jeanphilippegg:
This is strange. It seems I can reproduce the issue, but it only happens when it is the first line of the Dired buffer, for example. I'll check it out.

jeanphilippegg:
> Merged, thank you! I just made a small comment about the double dash in titles, but it is not a blocking issue.

That is fine! I made pull request #362 to clarify this comment.

We already had this discussion in #331. Sorry if I left some
misunderstandings. I will try to clear it with a better explanation.

In Denote, we have the "hard rules" of file names as well as the
"sluggification".

# The "hard rules"

The hard rules of file names are as permissive as possible. There are not many
such rules, but there is no way for the user to opt-out of those rules. *We
enforce them unconditionally.* They are:

- The dot character is not allowed. It only indicates the extension.
- A component cannot end with the delimiter symbols. Otherwise, we risk not
  being able to detect the next component.
- Any consecutive delimiter symbol is replaced with a single one, except for
  the title, where we allow "-----" (This is the confusing comment we are
  talking about.)
- Keywords cannot use underscores.

As you can see from the above, these "hard rules" really are minimal and users
will probably never notice them. In particular, these rules allow capital
letters, spaces, underscores, punctuation, consecutive hyphens in titles, etc.
We don't want to limit these here unless it is absolutely necessary. Imposing
further rules is the job of the sluggification, for which we provide a
sensible default.

# The sluggification

On top of those "hard rules", we have the sluggification for which we provide
a sensible default. They are an opinionated set of additional rules that we
apply on file names through the user option `denote-file-name-slug-functions`.
They are:

- Downcasing
- Replace spaces with "-" in titles
- Punctuation and special characters are removed
- Consecutive hyphens are replaced with a single one, even in titles.
- etc.

The default sluggification enforce these additional rules. However, the user
is able to opt-out of them by configuring `denote-file-name-slug-functions`.

# Conclusion

-- is allowed in titles only when the user decides to opt-out of the default
sluggification. The same applies to capital letters, special characters, etc.
In a sense, this is already hidden behind an "opt-in" mechanism.

The comment you are refering to in the code was made in
`denote--replace-consecutive-token-characters` which is used to enforce the
"hard rules", which are as permissive as possible (but not the default).

Hope things are clearer!

jeanphilippegg:
I may have double-posted the previous comment! Sorry

protesilaos:
> Date: Sun, 19 May 2024 07:06:26 -0700

> [... 54 lines elided]

> The comment you are refering to in the code was made in
> `denote--replace-consecutive-token-characters` which is used to enforce the
> "hard rules", which are as permissive as possible (but not the default).
>
> Hope things are clearer!

This is clear, thank you!

-- 
Protesilaos Stavrou
https://protesilaos.com

protesilaos:
> Date: Sun, 19 May 2024 07:13:17 -0700
>
> I may have double-posted the previous comment! Sorry

No worries!

-- 
Protesilaos Stavrou
https://protesilaos.com

jeanphilippegg:
> I just noticed that the fontification does not work in one case:

Just commenting to let you know that I am still investigating the issue. This
is a problem with the first item in a Dired buffer. Generally, there is a
`.git/` or some folder there.


-------

After my currently open pull request, there will only be one more big pull
request to make `denote-directory` accept a list of directories. After that, I
think we will have all the features for Denote 3.0.0. But now, I intend to
take a small pause to see if any issues arise from the latest changes that
were made. I will investigate the current issue in that time.

I think you can close these issues: #277, #279, #287.

protesilaos:
> Date: Sun, 19 May 2024 21:49:16 -0700
>
>> I just noticed that the fontification does not work in one case:
>
> Just commenting to let you know that I am still investigating the issue. This
> is a problem with the first item in a Dired buffer.

Sure, no problem!

> Generally, there is a `.git/` or some folder there.

True for flat listings, though users can have subdirectories.

> -------
>
> After my currently open pull request, there will only be one more big pull
> request to make `denote-directory` accept a list of directories. After that, I
> think we will have all the features for Denote 3.0.0.

Very well!

> But now, I intend to take a small pause to see if any issues arise
> from the latest changes that were made. I will investigate the current
> issue in that time.

Yes, I think it is good practice to have 1-2 weeks of no major changes
before we make a new version.

> I think you can close these issues: #277, #279, #287.

Okay!

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #363, opened 2024-05-19
Bug with `denote-date` + remove code duplication.

jeanphilippegg:
When using `denote-date` with an empty input at the date prompt, the id is
19991231T190000 instead of using the current-time as in `denote-org-capture`.

I fixed `denote-parse-date`. I was also able to remove the code duplication
between `denote` and `denote-org-capture`.


protesilaos:
Merged, thank you!

PR #368, opened 2024-05-25
Bug fixes: Fix the fontification + another bug

jeanphilippegg:
Bug fixes:

- This fixes the fontification. If we have issues with that in the future, we
  can also replace `denote-faces-dired-file-name-matcher` with the function
  [`dired-filename-search-forward`](https://github.com/protesilaos/denote/issues/271#issuecomment-2113961191) that mentalisttraceur shared in #271.
  I tested it as well and it worked. Just mentioning it as a possible alternative.

- I also found another unrelated bug: `denote-org-capture` would always use
  the current time for the id in the front matter irrespective of the value of
  a possible date prompt.

I added some safeguards for when we will allow identifiers to have other
formats. The code is useless for now, but it is ready.

- I made the fontification ready to handle any identifier format.

- Prevent infinite loops in case `denote-id-format` is misconfigured.

- Make sure that only identifiers that have the specific format
  00000000T000000 are allowed to drop their delimiter in file names. This is
  important if we decide at some point that titles can drop their delimiter in
  first position. We don't want ambiguity with an arbitrary identifier. This
  is to keep our options open for the future.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1940,8 +1940,9 @@ which case it is not added to the base file name."
             ((and (eq component 'signature) signature (not (string-empty-p signature)))
              (setq file-name (concat file-name "==" (denote-sluggify 'signature signature))))))
     (setq file-name (concat file-name extension))
-    ;; Do not prepend identifier with @@ if it is the first component.
-    (when (string-prefix-p "@@" file-name)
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you!

PR #383, opened 2024-06-22
Add an invisibility spec to address disabling `denote-fontify-links-mode`

haji-ali:
Follow on from #344 

This addresses one issue that @protesilaos raised about disabling `denote-fontify-links-mode` and invisible text being left behind.

PS: I thought that removing the font-lock keyword would reset the text properties, turns out that is only true for `face` properties. I don't know how org addresses this issue yet, but this PR is one potential fix.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -4246,7 +4246,7 @@ Implementation based on the function `org-activate-links'."
                      (non-sticky-props
                       '(rear-nonsticky (mouse-face highlight keymap invisible intangible help-echo htmlize-link)))
                      (face-property 'link)
-                     (hidden (append '(invisible t) properties)))
+                     (hidden (append '(invisible 'denote-link) properties)))
                 (remove-text-properties start end '(invisible nil))
                 (add-text-properties start visible-start hidden)
                 (add-face-text-property start end face-property)
@@ -4265,9 +4265,6 @@ To be used as a `thing-at' provider."
 
 (defvar thing-at-point-provider-alist)
 
-;; FIXME 2024-06-19: We are missing a function that clean up all those
-;; properties when the mode is disabled.  Otherwise, we are left with
-;; invisible text, which looks broken.
 (define-minor-mode denote-fontify-links-mode
   "A minor mode to fontify and fold Denote links."
   :init-value nil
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged. Thank you!

protesilaos:
> I don't know how org addresses this issue yet, but this PR is one potential fix.

Unrelated to this point, do you have any update on the 'unless' I added to the minor mode so that it does not work in Org? Maybe we can have a better way to deal with this? My concern is to not interfere with Org's fontification because it may cause weird bugs.

haji-ali:
> Unrelated to this point, do you have any update on the 'unless' I added to the minor mode so that it does not work in Org? Maybe we can have a better way to deal with this? My concern is to not interfere with Org's fontification because it may cause weird bugs.

@protesilaos, I understand your concern. More tests and potential modification would be needed to make sure the fontification doesn't interfere with `org-mode` and it could be a moving goal post, in case of changes in `org-mode`.

My reservation about the conditional in `denote-fontify-links-mode` is that the minor mode is enabled even if `org-mode` is non-nil (e.g., would be listed in the list of minor modes), it's just an ineffective minor mode in this case that does absolutely nothing. That is why I think that the user should just not enable the minor mode in this case (encouraged by documentation).

`denote` could  provide a helper function which enforces this recommendation, which the user could call in `text-mode-hook` or wherever they desirer.

```lisp
(defun denote-fontify-links-mode-maybe ()
  "Enable `denote-fontify-links-mode' in a denote file unless in `org-mode'."
  (when (and (buffer-file-name)
             (denote-file-is-note-p (buffer-file-name))
             (not (derived-mode-p 'org-mode)))
    (denote-fontify-links-mode)))
```

protesilaos:
> From: Abdul-Lateef Haji-Ali ***@***.***>
> Date: Sat, 22 Jun 2024 12:16:06 -0700
>
>> Unrelated to this point, do you have any update on the 'unless' I
>> added to the minor mode so that it does not work in Org? Maybe we can
>> have a better way to deal with this? My concern is to not interfere
>> with Org's fontification because it may cause weird bugs.
>
> @protesilaos, I understand your concern. More tests and potential
> modification would be needed to make sure the fontification doesn't
> interfere with `org-mode` and it could be a moving goal post, in case
> of changes in `org-mode`.

I say we leave this for after the release of Denote version 3.0.0. My
plan is to publish the new version before the end of the week. Well,
unless there is some blocking issue.

> My reservation about the conditional in `denote-fontify-links-mode` is
> that the minor mode is enabled even if `org-mode` is non-nil (e.g.,
> would be listed in the list of minor modes), it's just an ineffective
> minor mode in this case that does absolutely nothing. That is why I
> think that the user should just not enable the minor mode in this case
> (encouraged by documentation).

You are right. The current setup is not nice.

> `denote` could provide a helper function which enforces this
> recommendation, which the user could call in `text-mode-hook` or
> wherever they desirer.
>
> ```lisp
> (defun denote-fontify-links-mode-maybe ()
>   "Enable `denote-fontify-links-mode' in a denote file unless in `org-mode'."
>   (when (and (buffer-file-name)
>              (denote-file-is-note-p (buffer-file-name))
>              (not (derived-mode-p 'org-mode)))
>     (denote-fontify-links-mode)))
> ```

I like this! Do you want to send the PR for it?

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #392, opened 2024-07-14
Add indicator for buffer having backlinks to denote-rename-buffer

ashton314:
Hello Prot, thanks for all the work you've done with Denote!

Something I've wanted is some kind of indicator in the mode line that would alert me to the presence of backlinks to a file that I might want to investigate. I've put together a little prototype. It's clumsy; I hope you can give me some pointers on how to improve it. Here is what it does:

It adds `%b` to `denote-rename-buffer-format` which `denote-rename-buffer--format` will replace with `denote-buffer-has-backlinks-string` (defaults to `→ `) when the buffer has backlinks pointing to it.

The way it checks if a buffer has backlinks is possible more expensive than necessary; I call `(length (denote-link-return-backlinks file))` and check if the result is non-zero; that way I know for sure if a buffer has any backlinks Denote would recognize.

It seems to be working fine on my end; I'd be happy to hear any feedback. Let me know if this is welcome, and if it is, how to improve it.

Thanks!

### DIFF
--- a/README.org
+++ b/README.org
@@ -3009,6 +3009,7 @@ are placeholders for Denote file name components ([[#h:4e9c7512-84dc-4dfb-9fa9-e
 - The =%d= is the same as =%i= (=DATE= mnemonic).
 - The =%s= is the Denote =SIGNATURE= of the file.
 - The =%k= is the Denote =KEYWORDS= of the file.
+- The =%b= is an indicator of whether or not the file has backlinks pointing to it.
 - The =%%= is a literal percent sign.
 
 In addition, the following flags are available for each of the specifiers:
@@ -3029,8 +3030,11 @@ include some text that makes Denote buffers stand out, such as
 a =[D]= prefix.  Examples:
 
 #+begin_src emacs-lisp
-;; Use the title (default)
-(setq denote-rename-buffer-format "%t")
+;; Use the title prefixed with the backlink indicator (default)
+(setq denote-rename-buffer-format "%b%t")
+
+;; Customize what the backlink indicator looks like
+(setq denote-buffer-has-backlinks-string "!! ")
 
 ;; Use the title and keywords with some emoji in between
 (setq denote-rename-buffer-format "%t 🤨 %k")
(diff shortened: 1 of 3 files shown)


ashton314:
(Also, I *have* signed the FSF copyright assignment, so there shouldn't be any legal blockers to merging this—just technological blockers. :)

protesilaos:
Hello @ashton314!

This is a good idea. I think it can be merged now, though I wonder if it is better to define a `denote-file-has-backlinks-p`. We can make it simpler than `denote--retrieve-files-in-xrefs`, which does some extra processing. What do you think?

ashton314:
That's a great idea—give me a day or two and I'll see what I can make.

protesilaos:
> From: Ashton Wiersdorf ***@***.***>
> Date: Wed, 17 Jul 2024 08:48:01 -0700
>
> That's a great idea—give me a day or two and I'll see what I can make.

Sure, take your time!

-- 
Protesilaos Stavrou
https://protesilaos.com

ashton314:
I've improved it:

 - I've added a function `denote--file-has-backlinks-p`. This still calls `denote-link-return-backlinks`, as this does all the work I think necessary to ensure that you don't get false positives. (E.g. by finding your identifier in your own file.)
 - The function `denote-rename-buffer--format` now first checks if the `denote-rename-buffer-format` has `%b` in it; if it does, it calls `denote--file-has-backlinks-p`. The reason for this check is because, if the user *doesn't* want the backlink indicator, they shouldn't have to pay the cost of searching for backlinks when they visit a buffer.

Please let me know if there is anything else I can do to improve this. I'm using it right now and I've found it quite helpful; I hope others do too. :)

ashton314:
I just edited the manual to note this change. Again, please let me know if you think the documentation fits elsewhere.

protesilaos:
Merged. Thank you!

protesilaos:
I made two small tweaks on top.

ashton314:
🎉

PR #398, opened 2024-07-18
Support functions in `denote-templates`

### LINKED ISSUE #394: Support using functions in `denote-templates`
Would it be possible to support using arbitrary functions as the content value for `denote-templates`? I have some templates that I would like to use dynamically-generated content in, and this seems like the simplest and cleanest way to do so, since we'd just have to check the type of each value, and, if it's a function, call it to get the template string.

skissue:
Adds support for using templates that dynamically generate content using a function.

Fixes #394

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -499,21 +499,22 @@ it, use the command `org-insert-link'.  Note that `org-capture' uses
 A template is arbitrary text that Denote will add to a newly
 created note right below the front matter.
 
-Templates are expressed as a (KEY . STRING) association.
+Templates are expressed as a (KEY . VALUE) association.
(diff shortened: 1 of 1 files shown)


skissue:
> I wonder if it makes sense to have the function accept arguments.

I suppose we could have the template functions take the other properties of the new note (title, keywords, etc.) as arguments. I'm not sure I see a use case for this though :thinking:. Though I'm sure there *is* one, so I'm open to the idea.

protesilaos:
Merged. Thank you!

I suggest we keep the functions simple for now and change things based on demand.

PR #399, opened 2024-07-21
Add support for function templates in denote-org-capture

jeanphilippegg:
Just adding support for function templates in `denote-org-capture` following the last merged pull request.


protesilaos:
Merged. Thank you!

By the way, we briefly discussed the possibility of having those functions accept arguments. I ultimately decided it was premature to do this at this stage, because we could not come up with a good use-case. If you have any ideas on that, I am happy to see them.

jeanphilippegg:
For now, I don't have any idea either about accepting arguments.

PR #415, opened 2024-08-26
Use absolute file path in denote-link--backlink-find-file

### LINKED ISSUE #412: Backlinks buffer links not working with nested dirs
Recently I have observed an issue. 
I have my notes in subfolders (within `denote-directory`), something like this. Excluding IDs for the simplification.
```
fun/roger_zelazny__author.org
fun/courts_of_chaos__novel.org
fun/amber__novel.org
```

Now `fun/courts_of_chaos__novel.org` and `fun/amber__novel.org` have links to `fun/roger_zelazny__author.org`.  I go to `fun/roger_zelazny__author.org` and call `denote-show-backlinks-buffer`, backlinks buffer opens up and shows the backlinks as:

```
fun/courts_of_chaos__novel.org
fun/amber__novel.org

```

However, when I click on any of the links, emacs doesn't open the correct file. It seems it is trying to open the file in `fun/fun` dir,  i.e. clicking on `fun/amber__novel.org` tries to open `fun/fun/amber__novel.org` which is actually not there. 

My env:
Emacs: `GNU Emacs 29.3 (build 1, x86_64-pc-linux-gnu, cairo version 1.18.0) of 2024-05-26`
Denote: At commit eec6033. I don't use silos.

jeanphilippegg:
I think you mixed two things in commits 7f9d040 followed by 8701908. You removed a line `(setq-local default-directory ...)` and added a line `(setq-local denote-directory ...)`. Those 2 are unrelated and removing the first line causes issue #412.

In the end, it is a good thing that this line is removed though. It was probably there because we open buttons in backlinks buffers by calling `find-find` on the relative path. But we should try to always work with absolute paths in the code.

To fix #412, I made `denote-link--backlink-find-file` work with absolute paths instead of reintroducing the `(setq-local default-directory ...)` line. It is ready to be merged.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -4274,7 +4274,9 @@ major mode is not `org-mode' (or derived therefrom).  Consider using
 
 (defun denote-link--backlink-find-file (button)
   "Action for BUTTON to `find-file'."
-  (funcall denote-link-button-action (buffer-substring (button-start button) (button-end button))))
+  (funcall denote-link-button-action
+           (concat (denote-directory)
+                   (buffer-substring (button-start button) (button-end button)))))
 
 (defun denote-link--display-buffer (buf &optional action)
   "Run `display-buffer' on BUF using optional ACTION alist.


protesilaos:
Merged, thank you!

protesilaos:
I don' remember what it was with the `setq-local` but I think I had to move it after the `erase-buffer`, otherwise I was not getting the expected results.

Working with absolute paths is better though. Should you also remove that `setq-local` then?

jeanphilippegg:
I meant to say that `denote-directory` and `default-directory` may look alike, but they are not the same variable. The remaining `setq-local` is about the `denote-directory` and it is still necessary because enabling `denote-backlinks-mode` seems to reset it when in a silo, ie after calling `denote-backlinks-mode`, `denote-directory` is back to its global value instead of the one in `.dir-locals.el`. You already have a comment for this, but it does not mention silos.

I do not understand why it happens and if it could affect other local variables.

protesilaos:
> Date: Mon, 26 Aug 2024 16:58:53 -0700
>
> I meant to say that `denote-directory` and `default-directory` may
> look alike,

Oh right! I misread them as the same in the diff.

> I do not understand why it happens and if it could affect other local variables.

I am not sure either. Maybe it uses the default settings of the major
mode? If this works and we do not have any downsides, we can keep it for
now. Though I am interested to understand what is happening.

-- 
Protesilaos Stavrou
https://protesilaos.com

Issue #416, opened 2024-08-27
"No links found" when using `denote-find-link` in denote files inside a silo


Kolmas225:
Steps to reproduce:
1. Open any denote file with links to another file inside the same silo
2. `M-x denote-find-link` 
3. "No links found" shows up in stead of a list of forward links of current file

protesilaos:
I just pushed a change. Please let me know if this works for you.

Even if it does work, I suspect there is something deeper going on. We have the same problem with the `denote-find-backlink` command, so I wonder if the core issue here is how the function `denote-directory` works when the silo is defined as a string. This is a change we made fairly recently (cc @jeanphilippegg). In a debug I was doing, the `denote-directory` function did return the silo for the first file in the list, but then it reverted to the non-silo value. I need to explore this further.

protesilaos:
Indeed, if I set the local value to a non-string like we had in the past, then `denote-find-backlink` works as expected. The .dir-local.el:

```elisp
((nil . (;; (denote-directory . "/home/prot/Documents/denote-test-silo/")
         (denote-directory . local))))
```

jeanphilippegg:
I think I have found out the cause the of issue. Execute the following function inside of a silo and test the commented lines:

```emacs-lisp
(defun test-dir-locals-el ()
  (interactive)
  (message "Before with-*-buffer: %s" denote-directory)
  (with-current-buffer (get-buffer-create "test")   ; <---- .dir-locals is forgotten
  ;;(with-current-buffer (find-file-noselect (buffer-file-name)) ; <---- WORKS!
  ;;(with-temp-buffer                       ; <---- .dir-locals is forgotten
    (message "Within with-*-buffer: %s" denote-directory))
  (message "After with-*-buffer: %s" denote-directory))
```

`.dir-locals.el` variables are forgotten inside a `with-temp-buffer` block. The same applies with `with-current-buffer`, but only when the specified buffer is not a buffer within the silo.

In `denote-find-link`, we were in the `with-temp-buffer` case. In `denote-link--prepare-backlinks`, we are in the `with-current-buffer` case with a temporary buffer.

In addition to the above, `(denote-backlinks-mode)` also reset buffer-local variables.

### The solution

Always do computations that depend on variables that may be specified in `.dir-locals.el` outside of a `with-temp-buffer` block and similarly for `with-current-buffer`. We don't have any problem when we use `with-current-buffer` with a buffer that is inside of the silo.

[...]

PR #419, opened 2024-08-31
Rework denote-link-return-links (fix issue #416)

jeanphilippegg:
I have reworked `denote-link-return-links` to limit the use of `with-temp-buffer` to get what we need.

You will notice that I reverted your last change to `denote-link-return-links` because you added a new parameter to `denote-link--expand-identifiers`, but it is used in `denote-org-extras.el`. It must have broken something there. I have kept `denote-link--expand-identifiers` because it is used there, but maybe we can use `denote-link-return-links` instead and remove it.

`denote-link-return-links` should also be more efficient now. Previously, for each notes we looped through all links in the file.

This is ready to be merged. I will get to `denote-link--prepare-backlinks` tomorrow.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -4418,8 +4418,11 @@ non-nil value."
     (with-current-buffer (get-buffer-create backlinks-buffer)
       (erase-buffer)
       (denote-backlinks-mode)
-      ;; Set the `denote-directory' after enabling the major mode,
-      ;; otherwise its value gets overwritten.
+      ;; In the backlinks buffer, the values of variables set in a
+      ;; `.dir-locals.el` do not apply.  We need to set `denote-directory' in
+      ;; the backlinks buffer because the buttons depend on it.  Moreover, its
+      ;; value is overwritten after enabling the major mode, so it needs to be
+      ;; set after.
       (setq-local denote-directory dir)
       (setq overlay-arrow-position nil)
       (goto-char (point-min))
--- a/denote.el
+++ b/denote.el
@@ -3953,9 +3953,9 @@ function."
         (push (match-string-no-properties 1) matches)))
     matches))
 
-(defun denote-link--expand-identifiers (regexp files)
-  "Expand identifiers matching REGEXP in FILES into file paths."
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
I have reviewed `denote-link--prepare-backlinks` and I think it is fine. We only need the value of `denote-directory` because the buttons need it to rebuild the absolute path. There is the `revert-buffer-function` part that I do not understand, but I think it is ok as well.

I just pushed a commit to rewrite the comment. We can consider this issue closed!

protesilaos:
Merged, thank you! I could not see the part related to denote-org-extras.el, but I will check again.

jeanphilippegg:
> Merged, thank you! I could not see the part related to denote-org-extras.el, but I will check again.

You're welcome! `denote-link--expand-identifiers` is used in `denote-org-extras-dblock--get-missing-links`

protesilaos:
> Date: Sat, 31 Aug 2024 05:39:06 -0700
>
>> Merged, thank you! I could not see the part related to denote-org-extras.el, but I will check again.
>
> You're welcome! `denote-link--expand-identifiers` is used in
> `denote-org-extras-dblock--get-missing-links`

Yes, this is fine. It does not use the FILES parameter.

-- 
Protesilaos Stavrou
https://protesilaos.com

jeanphilippegg:
> Yes, this is fine. It does not use the FILES parameter.

That was the point. It does not use the FILES parameter, but this parameter was added in one of the latest commits. But it is all sorted now! :)

protesilaos:
> Date: Sat, 31 Aug 2024 18:50:15 -0700
>
>> Yes, this is fine. It does not use the FILES parameter.
>
> That was the point. It does not use the FILES parameter, but this
> parameter was added in one of the latest commits. But it is all sorted
> now! :)

Yeah, sorry! I don't remember what I was trying to do. Maybe it was some
experiment and then I left it there...

-- 
Protesilaos Stavrou
https://protesilaos.com

Issue #413, opened 2024-08-17
`denote-create-note` silently


kvineet:
Thanks for your excellent package.

I wanted to do org-capture directly to the today's journal file. While this works when the  file already exists, it does not work correctly when today's journal does not exist.

Is there a way for `denote-create-note` to create file silently in background? and not open buffer for it.

Essentially on similar lines of `denote-save-buffers` but to kill buffer after saving.

jeanphilippegg:
The current `denote` command does not have this functionality. Here is how you can get this (interactive) command for now:

```emacs-lisp
(defun denote-create-note-silently ()
  (interactive)
  (when-let ((denote-save-buffers t)
             (new-note (call-interactively #'denote))
             (buffer (find-buffer-visiting new-note)))
    (kill-buffer buffer)))
```

or this may work:

```emacs-lisp
(setq denote-save-buffers t)
(add-hook denote-after-new-note-hook #'kill-buffer)
```

I think having an option to kill a buffer automatically is a good idea. It has already been suggested for renaming commands in #273. It would be good to consider creation commands when/if we get to implementing this feature.

protesilaos:
> From: Jean-Philippe Gagné Guay ***@***.***>
> Date: Sun, 25 Aug 2024 19:35:31 -0700

> [... 19 lines elided]

> I think having an option to kill a buffer automatically is a good
> idea. It has already been suggested for renaming commands in #273. It
> would be good to consider creation commands when/if we get to
> implementing this feature.

Yes, I think this is a nice feature to have.

-- 
Protesilaos Stavrou
https://protesilaos.com

protesilaos:
Closing now that we have the new user option `denote-kill-buffers` by @jeanphilippegg.

Issue #273, opened 2024-03-06
Enhancing `denote-rewrite-front-matter` to save and kill buffer when it's safe to do so


mentalisttraceur:
Here's a huuuge (for me) UX enhancement that I monkey-patch in from my config:

```lisp
(defun fixed-denote-rewrite-front-matter
        (denote-rewrite-front-matter path &rest arguments)
    (let* ((buffers (buffer-list))
           (buffer  (find-file-noselect path))
           (_ (refresh-modified-state buffer))
           (had-unsaved-changes (buffer-modified-p buffer)))
        (apply denote-rewrite-front-matter path arguments)
        (unless had-unsaved-changes
            (with-current-buffer buffer
                (basic-save-buffer))
            (unless (memq buffer buffers)
                (kill-buffer buffer)))))
(advice-add 'denote-rewrite-front-matter
    :around 'fixed-denote-rewrite-front-matter)
```

Big-picture overview:

1. If the buffer and file were the same before we rewrote the frontmatter, it's safe and nicer UX to automatically save it,

2. if the file was also not even open before we rewrote the frontmatter, it's safe and nicer UX to automatically close it after saving it.

Obviously "safe" here just means *as safe as me manually doing that right after*, and "nicer UX" here means that *I always would* manually do that right after.

Now the key piece here for correctness is `refresh-modified-state`. That's [my own invention](https://emacs.stackexchange.com/a/78828/13754) - it forces Emacs to re-check if buffer contents differ from the visited file. This, in my view, is *essential*. If the file changed on disk and `buffer-modified-p` is out of sync, automatically saving would be unacceptably prone to data loss. But once we can force a refresh of that state, we get *better* safety against data-losing race conditions than a manual check of the same (because the time it takes for a human to manually save the file after checking for buffer-vs-file differences is orders of magnitude longer than the time it takes for the code to both modify the frontmatter and save).

Anyway, my joy using Denote (and especially building other things with Denote as a building block) went up a *lot* ever since I implemented this wrapper. I've been using this for several months now with no issues, never looked back, don't want to ever go back. Hopefully it helps others too.

If there's interest in upstreaming this, I have some intent to eventually publish `buffer-differs-from-visited-file-p` and `refresh-modified-state` to ELPA, but I haven't yet signed GNU papers. In the meantime, the default license on code posted to StackOverflow/Exchange is permissive (MIT iirc) and I also offer all of that code under 0BSD, which is widely recognized as public-domain-equivalent.

[...]

PR #426, opened 2024-09-03
Option denote-kill-buffers to automatically kill buffers (issue #413 and #273)

jeanphilippegg:
The options `denote-save-buffers` (existing) and `denote-kill-buffers` (new)
now control what happens after a creation or a rename operation. Their default
is `nil`, meaning that we do not save or kill buffers automatically.

They should behave as expected. You can check their documentation. A few
notes:

- `denote-kill-buffers` is "stronger" in a sense than `denote-save-buffers`.
  When set, it would not make sense to not save the buffer as well.
- We NEVER kill a buffer that was already visited before a rename operation. I
  think this is an expected behavior. It would be strange otherwise.
- In issue #273, mentalisttraceur considered some other cases. For example,
  another process could modify a file and our automatic killing of the buffer
  could lead to data loss. I don't think that we need to worry too much about
  this. The defaults are to not save or kill and we have had
  `denote-save-buffers` for a few months already and no issues were reported.
  I have included a warning in the doc of `denote-save-buffers` if we deviate
  from the default.
- I have noticed that `denote-region` currently does not respect the value of
  `denote-save-buffers` and so will not respect that of
  `denote-kill-buffers`. I will think about this, but this is not a blocking issue.

I have tested this with a few creation/renaming commands and it worked as
expected. It would be ready to be merged.

This would close issues #413 and #273.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -157,10 +157,38 @@ commands (`denote-rename-file' and related).
 If this user option is set to a non-nil value, such buffers are
 saved automatically.  The assumption is that the user who opts in
 to this feature is familiar with the `denote-rename-file'
-operation (or related) and knows it is reliable."
+operation (or related) and knows it is reliable.  Data loss may
+occur if the file is modified externally."
   :group 'denote
   :package-version '(denote . "3.0.0")
   :type 'boolean)
+
+(defcustom denote-kill-buffers nil
+  "Control whether creation or renaming commands kill their buffer.
+
+The default behaviour of creation or renaming commands such as
+`denote' or `denote-rename-file' is to not kill the buffer they
+create or modify at the end of their operation.
+
+If this user option is nil (the default), buffers affected by a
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you!

protesilaos:
About `denote-region`, note that I want to publish version `3.1.0` today or tomorrow. I don't mind if you do not do it by then.

Also, if you have any other changes you want for `3.1.0`, please let me know.

jeanphilippegg:
> About `denote-region`, note that I want to publish version `3.1.0` today or tomorrow. I don't mind if you do not do it by then.
> 
> Also, if you have any other changes you want for `3.1.0`, please let me know.

I am good with 3.1.0 as it is.

I do not know when I will look at `denote-region` and how to fix it, but it is not a big issue. It is okay to go forward with 3.1.0.

Issue #405, opened 2024-08-03
Suggestion  a denote-keyword-append function.


MirkoHernandez:
I recently updated to denote 3 and I'm not getting used to the new` denote-keyword-add `behavior . I understand that it is much faster for adding and removing multiple keywords, but it is less reliable (a typing error can remove or rename a previous keyword).

And there are also problems with some other packages. I'm using citar-denote and the `citar-denote-add-citekey` command is broken since it can't append a keyword now, It calls `denote-keyword-add` with the intention of appending a keyword.

Providing an append function  would trivially allow backward compatibility with other packages. Checking the denote version would allow anyone to set the proper  function for appending keywords.

MirkoHernandez:
I had some trouble creating an append (and remove)  function similar to the previous `denote-keywords-add`.
I'm not sure how to easily tell denote to use the current values, is there a way to do this ?, I could not find a way searching the denote 3 code, I had to manually fetch and pass the values. 

``` emacs-lisp
(defun denote-keywords-append (&optional keywords)
  (interactive)
  (let* ((file (denote--rename-dired-file-or-current-file-or-prompt))
	 (denote-prompts '())
	 (cur-values (denote--rename-get-file-info-from-prompts-or-existing file))
	 (cur-title (cl-first cur-values))
	 (cur-keywords (cl-second cur-values))
	 (cur-signature (cl-third cur-values))
	 (new-keywords (seq-uniq (append cur-keywords
					 (or keywords (denote-keywords-prompt))))))
    (funcall-interactively 'denote-rename-file
			   file 
			   cur-title 
			   new-keywords
			   cur-signature
			   nil)))

```

protesilaos:
> From: Mirko Hernández ***@***.***>
> Date: Sat,  3 Aug 2024 13:34:05 -0700
>
> I recently updated to denote 3 and I'm not getting used to the new`
> denote-keyword-add `behavior . I understand that it is much faster for
> adding and removing multiple keywords, but it is less reliable (a
> typing error can remove or rename a previous keyword).

A typing error is indeed a concern, though this is true for all
functions we provide.

> And there are also problems with some other packages. I'm using
> citar-denote and the `citar-denote-add-citekey` command is broken
> since it can't append a keyword now, It calls `denote-keyword-add`
> with the intention of appending a keyword.

Oh, this is undesirable! I believe our old commands were declared as
"interactive-only", though I think the byte compiler does not actually
complain where such functions are used in Lisp.

[...]

PR #429, opened 2024-09-08
Make denote-rename-file parameters non-optional

jeanphilippegg:
This does not affect the interactive `denote-rename-file`.

When called from lisp, one may believe that `(denote-rename-file
"some-file.org" "new-title")` will only rename the title. However, it also removes keywords and signature.

It may be confusing (see for example #405) to have them optional. I think it is better to make them mandatory for this reason.

### DIFF
--- a/README.org
+++ b/README.org
@@ -1490,9 +1490,14 @@ separated by a comma, else the ~crm-separator~).
 For all prompts, the ~denote-rename-file~ interprets an empty input as
 an instruction to remove that file name component. For example, if a
 =TITLE= prompt is available and =FILE= is =20240211T093531--some-title__keyword1.org=
-then it renames =FILE= to =20240211T093531__keyword1.org=. If a file
-name component is present, but there is no entry for it in
-~denote-prompts~, keep it as-is.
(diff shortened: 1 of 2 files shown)


jeanphilippegg:
I just pushed a change as per the suggestion of MirkoHernandez in #405. You may want to give him credit for the idea in the manual.

`denote--rename-file` now accepts a `keep-current` symbol for its parameters. It should make it easier to use `denote-rename-file` from lisp by package authors and users that need it.

protesilaos:
Merged, thank you!

jeanphilippegg:
You're welcome!

I did not mention Mirko Hernandez in the commit messages, but I see that he is already acknowledged in the manual. Thanks to him for the idea in #405!

protesilaos:
> Date: Mon,  9 Sep 2024 17:00:13 -0700
>
> You're welcome!
>
> I did not mention Mirko Hernandez in the commit messages, but I see
> that he is already acknowledged in the manual. Thanks to him for the
> idea in #405!

Yes, he is there already. I will take a note to mention this in the
release notes.

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #443, opened 2024-09-20
add `denote-journal-extra-path-to-new-or-existing-entry'

### LINKED ISSUE #442: function to get the path of a journal file
I wrote a small function to refile a task to a journal note. To do that, I needed to obtain the path to the note, create it if necessary, or choose one if there are several. Luckily for me, `denote-journal-extras-new-or-existing-entry` does all that, so here is what I wrote:
```emacs-lisp
(defun get-today-journal-file (&optional date)
  (save-window-excursion
    (denote-journal-extras-new-or-existing-entry date)
    (buffer-file-name)))
```

Would it be better to extract the logic that finds the note from `denote-journal-extras-new-or-existing-entry` and `denote-journal-extras-link-or-create-entry` (which is the same code duplicated) to a function `denote-journal-extra-path-to-new-or-existing-entry`, which I could directly use, and refactor `denote-journal-extras-new-or-existing-entry` and `denote-journal-extras-link-or-create-entry` to use it?

This function would look like this:
```emacs-lisp
(defun denote-journal-extra-path-to-new-or-existing-entry (&optional date)
  (let* ((internal-date (denote-parse-date date))
         (files (denote-journal-extras--entry-today internal-date)))
    (cond
     ((length> files 1)
      (completing-read "Select journal entry: " files nil :require-match))
     (files
      (car files))
     (t nil))))

```

brabalan:
This function returns the path to a journal entry with the given date, or nil if there is none.

Refactored `denote-journal-extras-new-or-existing-entry' and `denote-journal-extras-link-or-create-entry' to use it.

Closes #442.

### DIFF
--- a/denote-journal-extras.el
+++ b/denote-journal-extras.el
@@ -172,6 +172,24 @@ DATE has the same format as that returned by `denote-parse-date'."
        (string-match-p keyword file))
      files)))
 
+(defun denote-journal-extra-path-to-new-or-existing-entry (&optional date)
+  "Return the path to the journal file corresponding to DATE, or to
+the current date if DATE in nil.  DATE is a string and has the same
+format as that covered in the documentation of the `denote' function.
+It is internally processed by `denote-parse-date'.
+
+If there are multiple journal entries for the date, prompt for one
+using minibuffer completion.  If there is only one, return it.  If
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you!

PR #444, opened 2024-09-21
Move code

jeanphilippegg:
I just moved the handling of the new `keep-current` parameter out of `denote--rename-file`. This internal function should not contain user logic code.

I also removed a comment about the possibility of regenerating link descriptions automatically on rename (in `denote--rename-file`). I don't think it would be a good idea to do this automatically.


jeanphilippegg:
I just fixed something in my commits. It is ready to be merged!

protesilaos:
Merged, thank you!

> I don't think it would be a good idea to do this automatically.

Agreed. Beside the implementation details of it, I think it also is not a good practice to rewrite potentially lots of files without reviewing them. Sometimes the links are meaningful in context and this is an easy way to make the text make less sense.

Issue #439, opened 2024-09-17
denote-rename-file and associated commands (denote-rename-file-keywords) replace formatting of the title with sluggified title


hapst3r:
Hej prot,

as a new phenomenon, I noticed that when I rename a filebuffer I am currently working on by adding keywords with `denote-rename-file-keywords`, my formatting gets overwritten with the sluggified title. I tried searching if this is intended behaviour (functions/variables, release notes) but couldn't find anything related.

This is what my front matter is supposed to look like:
```
#+title:      AS Gaza, Rudner: Gaza ist in uns, wir sind in Gaza
#+date:       [2024-09-17 Di 09:50]
#+filetags:   :innenpolitik:
#+identifier: 20240917T095025
```

This is what my front matter looks like after adding a keyword using `denote-rename-file-keyword`:
```
#+title:      as-gaza-rudner-gaza-ist-in-uns-wir-sind-in-gaza
#+date:       [2024-09-17 Di 09:50]
#+filetags:   :innenpolitik:krieg:
#+identifier: 20240917T095025
```

Since I use `denote-rename-buffer-mode`, I find this sluggification undesirable, because it takes away from the readability afforded by automatic buffer renaming.

Have a good day :)

jeanphilippegg:
This is not desirable.

What version of Denote are you using? I cannot reproduce this on my end.

Can you give the steps you are doing to get this behavior? For example, here is one of my tests:

1- Call `denote` to create a note with title "A note".
2- Save the note.
3- With the note still open, call `denote-rename-file-keywords` and add a keyword.

Also, can you evaluate this while inside the problematic note (using `M-:`) and share the output?

```emacs-lisp
(denote-retrieve-front-matter-title-value buffer-file-name 'org)
```

protesilaos:
> From: Jean-Philippe Gagné Guay ***@***.***>
> Date: Tue, 17 Sep 2024 17:22:50 -0700
>
> This is not desirable.
>
> What version of Denote are you using? I cannot reproduce this on my end.

I cannot reproduce it on my end either.

-- 
Protesilaos Stavrou
https://protesilaos.com

LukasCBossert:
I had this issue as well, but with the latest update I cannot reproduce it either.

hapst3r:
> This is not desirable.
> 
> What version of Denote are you using? I cannot reproduce this on my end.
> 
> Can you give the steps you are doing to get this behavior? For example, here is one of my tests:
> 
[...]

PR #445, opened 2024-09-21
Use file-truename to handle symbolic links (issue #439)

jeanphilippegg:
This fixes the issue that I could reproduce in #439.


protesilaos:
Merged, thank you!

Issue #346, opened 2024-05-11
denote-dired-mode ("fontification") not triggered in subdirectory if subdirectory is symlinked


julian-hoch:
I am not sure if this is a defect in Denote, or something that is just not possible in Emacs, but if 
you

- set up multiple subdirectories below your main denote directory
- configure the hook for dired-mode
- enable denote-mode in dired for subdirectories using denote-dired-directories-include-subdirectories

you will get dired mode only for subdirectories that are physical subdirectories, not symlinks. I am wondering if there is a way to make sure dired follows the symlink when determining whether to enable the denote mode.

A workaround is to manually add the symlinked directories to "denote-dired-directories", but obviously this only works if there are few subdirectories (and makes it a bit cumbersome to configure for cases where the same config is used on different computers with a different subfolder structure on the respective denote directory).

protesilaos:
> From: Julian Hoch ***@***.***>
> Date: Sat, 11 May 2024 10:46:06 -0700
>
> I am not sure if this is a defect in Denote, or something that is just not possible in Emacs, but if 
> you
>
> - set up multiple subdirectories below your main denote directory
> - configure the hook for dired-mode
> - enable denote-mode in dired for subdirectories using denote-dired-directories-include-subdirectories

Okay, let's try to find out:

- Go to the root of your denote-directory.
- Do M-x eval-expression
- Or type M-: with default key bindings.
- At the eval-expression prompt, write (denote-directory-subdirectories) and hit RET.
- You will get the return value of that function. Does it include the symlinks?

-- 
Protesilaos Stavrou
https://protesilaos.com

julian-hoch:
Hi Prot!
Yeah, the directories are all set up correctly. 
I suspect that the issue lies here:

```
;; denote.el - defun denote-dired-mode-in-directories
[...]
             ((or (member (file-truename default-directory) dirs)
[...]
```

"file-truename" resolves symlinks, so this will compare the directories I configured in my emacs config with the "true" path of the dorectory I am in - which will not resolve to the same path in my case.

In my case, I set up the denote directories as ~/denote/private, ~/denote/customer, which link to ~/git/denote-private-kb and ~/git/denote-customer-kb, just for convenience. So the "member" test will fail if I access them via the symlink (but it works if I dired into the "real" path).

I suspect this could be fixed by also applying file-truename to all the directories in the "dirs" list.
Julian

protesilaos:
Thank you for looking into this!

[...]

PR #447, opened 2024-09-21
Revert "Use file-truename in denote-filename-is-note-p and denote--di…

jeanphilippegg:
I am reverting commit 5fade0c because it will break Denote for some users.

If one's `denote-directory` contains symlink subdirectories (that link to
directories outside `denote-directory`), then some Denote features for those
notes will not work anymore.

We should not use `file-truename` because those directories will become
"outside" `denote-directory`. I don't see a non-costly way of doing this
validation. So I just recommend this warning in the manual:

```
If your `denote-directory` or its subdirectories involve symbolic links, make
sure that you access them using those links (and that
`find-file-visit-truename` is nil). Otherwise, Denote features will not work.
```

For the same reason, we should remove `file-truename` to fix related issue
#346 (with the same warning as above). I will do a separate pull request.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1066,8 +1066,7 @@ extensions are those implied by the variable `denote-file-type'."
 For our purposes, its path must be part of the variable
 `denote-directory', it must have a Denote identifier in its name, and
 use one of the extensions implied by the variable `denote-file-type'."
-  (and (string-prefix-p (file-truename (denote-directory))
-                        (file-truename (expand-file-name filename)))
+  (and (string-prefix-p (denote-directory) (expand-file-name filename))
        (denote-file-has-identifier-p filename)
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you!

> I don't see a non-costly way of doing this validation.

Perhaps we can provide this as an opt-in feature. Though this also comes down to the complexity of a user's setup, so maybe we should not do anything to account for it.

protesilaos:
> I will do a separate pull request.

Thank you!

jeanphilippegg:
> Perhaps we can provide this as an opt-in feature. Though this also comes down to the complexity of a user's setup, so maybe we should not do anything to account for it.

Yes, it could be a good solution.

PR #451, opened 2024-10-01
make denote-journal-extra-path-to-new-or-existing-entry create a missing entry

brabalan:
When using this function, I always created a journal file if it was missing. As this simplifies the other functions that uses it as well, I propose to have `denote-journal-extra-path-to-new-or-existing-entry` create a journal entry if it is missing.


protesilaos:
Makes sense, thank you! I just merged it.

PR #453, opened 2024-10-06
Fix issue reported in #435

jeanphilippegg:
This should fix the issue reported by ssl19 in #435 about digits being sluggified in keywords.

The hyphen character should appear last in a character alternative.


protesilaos:
Merged. Thank you!

PR #456, opened 2024-10-10
Remove redundant "Denote" item from denote--menu-contents

grtcdr:
This pull request removes the redundant "Denote" item from the `denote--menu-contents` variable. 

This item surprisingly does not do anything when clicked, so all the more reason for it to be removed.


protesilaos:
Merged, thank you!

PR #457, opened 2024-10-11
Fix unused let-bound variable warnings

grtcdr:
I noticed some warnings pop up during a reinstallation of Denote and went and fixed them.

Here are the errors for reference:

```
⛔ Warning (comp): denote.el:4347:15: Warning: variable ‘_’ not left unused
⛔ Warning (comp): denote-org-extras.el:640:15: Warning: variable ‘_’ not left unused
⛔ Warning (comp): denote-org-extras.el:637:15: Warning: variable ‘_’ not left unused
```

Unlike `pcase` which holds a special meaning for `_`, there's no way to mark a let-bound variable as ignored but we can omit the SYMBOL name from the VARLIST and the `when-let` expression will still evaluate correctly.

### DIFF
--- a/denote-org-extras.el
+++ b/denote-org-extras.el
@@ -667,10 +667,10 @@ Used by `org-dblock-update' with PARAMS provided by the dynamic block."
 (defun denote-org-extras-dblock--get-file-contents-as-heading (file add-links)
   "Insert the contents of Org FILE, formatting the #+title as a heading.
 With optional ADD-LINKS, make the title link to the original file."
-  (when-let ((_ (denote-file-is-note-p file))
+  (when-let (((denote-file-is-note-p file))
              (identifier (denote-retrieve-filename-identifier file))
              (file-type (denote-filetype-heuristics file))
-             (_ (eq file-type 'org)))
+             ((eq file-type 'org)))
     (with-temp-buffer
       (let ((beginning-of-contents (point))
             title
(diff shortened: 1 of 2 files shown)


grtcdr:
I found out that `denote-get-identifier-at-point` is an unused function through `xref-find-references` so you might consider removing it if no one's making use of it (which is hard to know for me).

protesilaos:
This is fine and I am happy to merge it. Which version of Emacs are you using? I do not get these warnings on my end...

> consider removing it if no one's making use of it (which is hard to know for me).

I know at least one case of someone using it. But we may want to think of how best to organise our files to decide where is the best place to put these sort of functions.

grtcdr:
Yeah, I should've mentioned I was on version `29.4`, if my changes don't cause noise on older versions of Emacs then I'd consider it safe to roll out.

protesilaos:
Merged, thank you!

I remember the standard approach of `when-let` to be like this:

```elisp
(when-let (((stringp a))
           (one a))
  BODY)
```

We have those cases in the Denote source code. Though I do not get any errors on my end with this approach (Emacs 31, but I believe it also worked on Emacs 30):

```elisp
(when-let ((_ (stringp a))
           (one a))
  BODY)
```

The latter seems more readable to me, but I will not argue with the byte compiler.

PR #462, opened 2024-10-16
allow nil type in denote-rename-buffer--format

### LINKED ISSUE #461: `denote-rename-buffer--format` should not use `when-let` for `type`
This line https://github.com/protesilaos/denote/blob/main/denote-rename-buffer.el#L103 for `denote-rename-buffer--format` make the whole function return `nil` if the file type is not in `denote-file-types`. This is inconvenient as the function can deal with arbitrary files using the file name to derive the buffer name.

I propose to move the binding of `type` to a simple `let`. I checked the rest of the function and having `type` being `nil` is fine.

brabalan:
This fixes #461


protesilaos:
This seems right. Merged, thank you!

PR #464, opened 2024-10-20
Some cleanup and fix `denote-link-dired-marked-notes`

jeanphilippegg:
I removed the nil checking of `denote--dir-in-denote-directory-p`. It is not
needed. It is an internal function. Calling it with nil would be a mistake. It
seems like it came from one of my commits (34daa52) where we were using a
`when-let`. It does not look like it was introduced to fix a bug. If there is,
we should ensure that we call it with a non-nil value.

I removed the condition in `denote-link--map-over-notes`. If one open the
parent of denote-directory in Dired and use "i" on the denote-directory, the
notes of this subdirectory will be listed in the current Dired buffer. The
user can mark them and call denote-link-dired-marked-notes. Previously, this
would fail because the current Dired buffer is not in denote-directory.

Speaking of `denote-link-dired-marked-notes`, since commit d942333, it is
broken because the `insert` line still expects `denote-link--prepare-links` to
return a string, but it now returns a list of strings. I fixed it.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -4719,10 +4719,9 @@ This command is meant to be used from a Dired buffer."
       (user-error "The buffer's file type is not recognized by Denote")))
   (when (y-or-n-p (format "Create links at point in %s?" buffer))
     (with-current-buffer buffer
-      (insert (denote-link--prepare-links
-               files
-               (denote-filetype-heuristics (buffer-file-name))
-               id-only)))))
+      (denote-link--insert-links files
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
I have pushed a change. `denote-link--buffer-prompt` did not work with
absolute paths. This command is only used by `denote-link-dired-marked-notes`.
It should now work in subdirectories.

protesilaos:
Merged, thank you!

PR #466, opened 2024-10-22
Fix regression caused by commit 4db0214

jeanphilippegg:
Sorry about this regression. I tested `denote-link-dired-marked-notes`, but not `denote`. I sat on these changes for some time and thought I had checked all locations of `denote--dir-in-denote-directory-p` already. I must have confused it with something else.


protesilaos:
No worries! Thank you for this. Merged.

Issue #467, opened 2024-10-23
denote-file-types: lack of :date-key-regexp and :identifier-key-regexp


sarcom-sar:
Hey, I've began hacking a bit on denote and started wondering why there are no keywords for `:date-key-regexp` and `:identifier-key-regexp` in `denote-file-types`, but there are for `:title-key-regexp` and `:keywords-key-regexp`.

Is it a "political" decision, because ID can be easily found in filename, while the date would be a huge pain?

As for proposed implementation `:identifier-key-reverse-function` the defun for `:title-key-reverse-function` could be easily reused, while the date would need something akin to:

```elisp
;; example for org
(defun denote-extract-org-date-from-front-matter (string)
  (org-read-date nil t string nil))
```

Sadly, this implies _at least_ three separate functions, which might be smelly, but since we have `denote-date-foo`, where foo is rfc3339, iso-8601 or org-timestamp, it would still be symetrical.

jeanphilippegg:
The reason they were not added was that we had no use for them yet.

However, I agree that we should probably include them all. This will make
`denote-file-types` bigger, but this is not really an issue.

We want to include the signature in the front matter (see #366). We will need
to add the relevant functions for signature at some point. The other functions
would also be necessary if we allow changing the identifier/date under some
circumstances.

I am curious: What is your use-case for them?

sarcom-sar:
>What is your use-case for them?

Retrieval of `:date-key-regexp` et al. and `:identifier-key-regexp` et al. would be useful when working with individual notes downstream.  To be precise, it would allow to elegantly pattern match elements according to needs, like:

```elisp
(let ((foo (pcase some-keyword
	     ('title :title-key-regexp)
	     ;; you know the rest
	     ('identifier :identifier-key-regexp)
	     (_ nil)))
      (file-type (denote-filetype-heuristics file)))
  (unless foo
    (re-search-forward (plist-get foo file-type) nil t)))
```

>The other functions would also be necessary if we allow changing the identifier/date under some conditions.

To be honest, I now understand the approach to `denote-file-types` better. The mutable parts are retrievable from front-matter, while non-mutable parts are not.  Under my understanding the `:date-function` made more sense as part of `:date-key-regexp` et al. but now I understand the code's approach.

protesilaos:
I just merged the PR of @jeanphilippegg. Thank you @sarcom-sar for the discussion! Closing now.

PR #468, opened 2024-10-26
Add remaining front matter retrieval functions

jeanphilippegg:
In response to #467 and to allow us to handle the front matter in future development, I have added the missing front matter retrieval functions in `denote-file-types`. I assumed the signature would be included eventually.

Since we still do not use them, this should not break anything. I created and renamed a note of each type to be sure.

Note that I renamed `:date-format-function` to `:date-value-function` to be consistent with the other components.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1464,12 +1464,9 @@ Change the front matter format'.")
 
 (defun denote-format-string-for-md-front-matter (s)
   "Surround string S with quotes.
-If S is not a string, return a literal emptry string.
 
 This can be used in `denote-file-types' to format front mattter."
-  (if (stringp s)
-      (format "%S" s)
-    "\"\""))
+  (format "%S" s))
 
 (defun denote-trim-whitespace (s)
   "Trim whitespace around string S.
@@ -1487,9 +1484,8 @@ This can be used in `denote-file-types' to format front mattter."
   (denote--trim-quotes (denote-trim-whitespace s)))
 
 (defun denote-format-string-for-org-front-matter (s)
-  "Return string S as-is for Org or plain text front matter.
-If S is not a string, return an empty string."
-  (if (stringp s) s ""))
+  "Return string S as-is for Org or plain text front matter."
+  s)
 
 (defun denote-format-keywords-for-md-front-matter (keywords)
   "Format front matter KEYWORDS for markdown file type.
--- a/denote.el
+++ b/denote.el
@@ -1518,10 +1518,15 @@ Split KEYWORDS-STRING into a list of strings.
 Consult the `denote-file-types' for how this is used."
   (split-string keywords-string "[:,\s]+" t "[][ \"']+"))
 
+(defun denote-extract-date-from-front-matter (date-string)
+  "Extract date object from front matter DATE-STRING.
+
+Consult the `denote-file-types' for how this is used."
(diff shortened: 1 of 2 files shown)


jeanphilippegg:
I have added a commit to simplify `denote-format-string-for-{org/md}-front-matter`.
We do not need to check that the parameter is a string. I checked all code
paths and it is not possible for them to be called with a nil value. That
would be a mistake. If it were possible before, it may have been fixed as part
of the refactoring of the renaming commands.

Now, the only remaining possibility is if the user calls `denote-rename-file`
or `denote-add-front-matter` from lisp with a nil title, but that would be a
mistake. The docstring says that it is a string parameter. Interactively, all
is fine.

This is ready to be merged!

protesilaos:
> from lisp with a nil title, but that would be a mistake

Do you think we should leave this as-is or interpret nil as an empty string in this case? I am fine with what we have. Just asking.

protesilaos:
Merged, thank you!

jeanphilippegg:
I think a nil value for a parameter makes sense when it means "not set", such
as with optional parameters. In the case of these commands, the parameter is
mandatory. It seems reasonable to me to expect users to call them with
arguments of the appropriate type.

-----

Since we are at it, in commit 55e923f, about `denote-format-file-name`, you
made this comment in git:

> We do not want to limit this to empty strings, because there are cases
we will need to arrange for the return value to be that. Whereas we
get nil for free.

In this case, I can understand that the title and signature can have this "not
set" meaning. My preference is still to enforce a string because the
parameters are mandatory. But I am conflicted, because at some point, I may
suggest to make the id optional and nil may be better than an empty string to
mean an absent identifier. It would not really be an issue, but it may be
confusing that nil is permitted for a paramater, but not the others.

You can remove it if you want. It would be safe: we never call it with nil in
the code (I just checked.).

-----

Also, I am okay if you would decide to revert my last commit (about
`denote-format-string-for-{org/md}-front-matter`).

protesilaos:
> Date: Sat, 26 Oct 2024 06:40:10 -0700
>
> I think a nil value for a parameter makes sense when it means "not set", such
> as with optional parameters. In the case of these commands, the parameter is
> mandatory. It seems reasonable to me to expect users to call them with
> arguments of the appropriate type.

This is fine, thank you!

> Since we are at it, in commit 55e923f, about `denote-format-file-name`, you
> made this comment in git:
>
>> We do not want to limit this to empty strings, because there are cases
> we will need to arrange for the return value to be that. Whereas we
> get nil for free.
>
> In this case, I can understand that the title and signature can have this "not
> set" meaning. My preference is still to enforce a string because the
> parameters are mandatory. But I am conflicted, because at some point, I may
> suggest to make the id optional and nil may be better than an empty string to
> mean an absent identifier. It would not really be an issue, but it may be
> confusing that nil is permitted for a paramater, but not the others.

Indeed, it is better to have this nuanced treatment of the value.

> You can remove it if you want. It would be safe: we never call it with nil in
> the code (I just checked.).
>
> -----
>
> Also, I am okay if you would decide to revert my last commit (about
> `denote-format-string-for-{org/md}-front-matter`).

No, let's keep it as-is. I just want to ask to be sure I am on the same
page as you.

-- 
Protesilaos Stavrou
https://protesilaos.com

jeanphilippegg:
Perfect!

PR #472, opened 2024-11-02
Small cleanup of the front matter functions

jeanphilippegg:
In this pull request:

### Add `denote--retrieve-front-matter-*-from-content` functions

These are just like `denote-retrieve-front-matter-*` functions, but they work on a content string instead of a file. (Maybe there would a way to remove the code duplication further.)

With them, we can remove the two functions `denote--get-{title/keywords}-line-from-front-matter`.

### Clean up denote--format-front-matter

- The date parameter is now a date object and `denote-format-front-matter` handles its formatting through `denote-date`, like other components.
- The identifier line is now handled like the other components in md front matters.
- Removed `denote--format-front-matter-{title/keywords}` as they do not need a function of their own. All front matter components are handled within `denote--format-front-matter`.

### Some internal support of signatures in front matter

Two internal functions handle signature lines in the front matter. There are still a few things to adjust and signatures are still not included in the default front matter templates. So they do nothing by default for now.


jeanphilippegg:
I just pushed a small change to make sure the signature will be sluggified in the front matter.

It is ready to be merged!

protesilaos:
Merged, thank you!

Issue #469, opened 2024-10-26
Use format specifiers for the link description


protesilaos:
I think we need to review how we generate link descriptions. The idea is to have an option with format specifiers for the various file name components (like what we do with `denote-rename-buffer-format`). This way, users can set their own style with ease. Plus, it makes it easier to write custom wrappers which `let` bind the format.

The code is in the branch `link-description-format` until I merge it into `main` (subject to feedback): <https://github.com/protesilaos/denote/tree/link-description-format>.

@jeanphilippegg Since we had worked on this before, do you think my suggested change is a step in the right direction? If not, do you prefer the status quo or something else?

jeanphilippegg:
That's good!

But are you aware that you now changed the default link description to
title-only? This is fine with me, just making sure.

In particular, links produced by `denote-link-with-signature` will also be
title-only, by default. This means that the only specific behavior of this
command is to restrict the completion candidates to files that have a
signature, which is still something. Users can always let-bind link
description options to configure the actual links.

One possibility is to reintroduce
`denote-link-description-with-signature-and-title` as the default and also
keep the new mechanism. But I am fine with the new default!

protesilaos:
> From: Jean-Philippe Gagné Guay ***@***.***>
> Date: Sat, 26 Oct 2024 06:05:59 -0700
>
> That's good!

Thank you!

> But are you aware that you now changed the default link description to
> title-only? This is fine with me, just making sure.

Yes, I know. This is not finalised. I just wanted to have a prototype to
get the discussion going.

> In particular, links produced by `denote-link-with-signature` will also be
> title-only, by default. This means that the only specific behavior of this
> command is to restrict the completion candidates to files that have a
> signature, which is still something. Users can always let-bind link
> description options to configure the actual links.

I have to think about this. One possibility is to keep its current
behaviour while 'let' binding the new option. But that is probably not a
good precedent.

> One possibility is to reintroduce
> `denote-link-description-with-signature-and-title` as the default and also
> keep the new mechanism. But I am fine with the new default!

Yes, this is also fine. Maybe we can add the new user option with a nil
value. A function like 'denote-get-link-description' will use a non-nil
value, but will otherwise delegate to 'denote-link-description-with-signature-and-title'.

Though the 'denote-link-with-signature' needs to be considered either way.

-- 
Protesilaos Stavrou
https://protesilaos.com

jeanphilippegg:
[...]

PR #473, opened 2024-11-03
Link descriptions

jeanphilippegg:
See my comment in https://github.com/protesilaos/denote/issues/469#issuecomment-2453499835.


protesilaos:
Merged, thank you!

PR #476, opened 2024-11-09
Cleanup of date handling function and more

jeanphilippegg:
This pull request is mostly about date handling, but there are a few other
cleanups as well.

- `denote-get-identifier`: The `date` parameter is no longer optional. This a
  a function (not a command) and we only used it without a date once. Also, it
  does not check `denote-valid-date-p` anymore. Within the scope of
  `denote-get-identifier`, it is too late for this validation. The callers
  should have done this. We almost always did this already. I patched the only
  place where we did not.

- `denote-valid-date-p`: Unchanged. This is the function that returns a valid
  date object (or nil). It is good!

- `denote--date`: Renamed to `denote--format-front-matter-date`.

- `denote-parse-date`: Obsoleted. It was just like `denote-valid-date-p`, but
  it converted nil dates to `(current-time)`. I think this is not really
  desirable. See below.

- `denote-create-unique-file-identifier`: Obsoleted. We used it only inside
  `denote--rename-file`, but it was not really the best place. We don't need
  it in the end.

- I removed some limitations of `denote-format-file-name`. This is a general
  function. As such, it is good that it can do things that we don't use
  (yet). For example, accept identifiers of different formats or empty
  identifiers. It is easily able to handle these cases. We enforce the format
  of the identifiers in other parts of the code, as necessary.

- I removed the id format checking from the
  `denote-{title/signature/keywords}-regexp`. We don't need to check the
  format of the identifier here. In fact, if we do, it would mean that a file
  "--title@@123.org" would yield the title "title@@123". We probably want to
  stop parsing any component when another delimiter is found. At some point,
  we may be more flexible with identifiers formats.

### Internal dates

Internally, we now always work with date objects (possibly `(current-time)`).
A nil value now always means "no date".

"No date" is still not publicly supported, i.e. Denote still always create notes
with a date/id. However, internally, functions are now capable of working with
nil dates. They do not need to be limited.

### DIFF
--- a/README.org
+++ b/README.org
@@ -5751,8 +5751,8 @@ might change them without further notice.
 
 #+findex: denote-retrieve-filename-identifier
 + Function ~denote-retrieve-filename-identifier~ :: Extract identifier
-  from =FILE= name, if present, else return nil.  To create a new one, refer to the
-  ~denote-create-unique-file-identifier~ function.
+  from =FILE= name, if present, else return nil.  To create a new one
+  from a date, refer to the ~denote-get-identifier~ function.
 
 #+findex: denote-retrieve-filename-title
 + Function ~denote-retrieve-filename-title~ ::  Extract Denote title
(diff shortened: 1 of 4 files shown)


jeanphilippegg:
# Additions:

### Small fix to an existing bug

I noticed an existing bug (not related to my changes): If the id needs to be
made unique, then the date in the front matter is not updated accordingly. I
added a commit to fix that.

### New defvar `denote-generate-identifier-automatically` (default `t`)

Eventually, we may want to make this a user option. However, for now, this is
only used internally and the value is always t. You can try setting it to
nil. You will notice that Denote is able to create/rename notes without an id
and without generating one! There are features that need to be arranged for it
to become a user option (fontification, front matter, linking). Check its docstring.

protesilaos:
Merged, thank you!

> ### New defvar `denote-generate-identifier-automatically` (default `t`)
> 
> Eventually, we may want to make this a user option. However, for now, this is only used internally and the value is always t. You can try setting it to nil. You will notice that Denote is able to create/rename notes without an id and without generating one! There are features that need to be arranged for it to become a user option (fontification, front matter, linking). Check its docstring.

Sounds good! Though I wonder if this makes more sense as a variable that users will `let` bind. So then we will have convenience commands that do the usual work minus the identifier. But this is an implementation detail, anyway: I do not mind a user option.

jeanphilippegg:
> Though I wonder if this makes more sense as a variable that users will `let` bind. So then we will have convenience commands that do the usual work minus the identifier. But this is an implementation detail, anyway: I do not mind a user option.

The difference is between a `defvar` and a `defcustom`?

I have these use-cases in mind:
- A user wants the file naming scheme of Denote, but never use the date/id. In that case, a global value of `nil` for `denote-generate-identifier-automatically` would make sense.
- A user wants identifiers in new notes. However, he wants the renaming commands to be more "explicit", ie do not add an identifier to a note that is missing one and also make it possible to input a nil date to remove the identifier. A global value of `on-creation` would make sense here.

PR #478, opened 2024-11-10
Fix denote-rename-file when called from lisp

jeanphilippegg:
I moved code from `denote--rename-get-file-info-from-prompts-or-existing` back to `denote--rename-file` because it would cause issues when `denote-rename-file` is used from lisp.

When calling `denote-rename-file` from lisp with a nil date, it would remove the date, ignoring `denote-generate-identifier-automatically`. Also, using keep-current in `denote-rename-file` would remove the date.


protesilaos:
Merged, thank you!

PR #479, opened 2024-11-11
Refactored `denote-rewrite-front-matter`

jeanphilippegg:
This new `denote-rewrite-front-matter` handles all front matter
lines/components. It works on each of them individually. If a line has no
value, it is removed. Removed lines are added again automatically on rename,
in a good position, if they have a non-empty value.

The new `denote-always-include-all-front-matter-lines` controls whether all
lines are present, even if the corresponding value is empty. I made the
default be `t` because that is currently what we are doing for keywords (ie
there is a keywords line even when there is no keywords).

------

This is a big pull request, so you may want to experiment with it before
merging it. You can try removing and reordering lines in a note and in the
front matter template and call something like this:

```
(denote-rewrite-front-matter buffer-file-name "A test" '("journal") "" nil "")
```

You should see the front matter be rebuilt correctly. Try with
`denote-always-include-all-front-matter-lines` set to `nil`.

I also made a comment in the patch about a technicality.

------

After this, it should be easy to add the signature and allow changing the
date/id. `denote-rewrite-front-matter` should now have no issue handling them.

### DIFF
--- a/README.org
+++ b/README.org
@@ -5901,11 +5901,11 @@ might change them without further notice.
 #+findex: denote-rewrite-front-matter
 + Function ~denote-rewrite-front-matter~ :: Rewrite front matter of
   note after ~denote-rename-file~ (or related) The =FILE=, =TITLE=,
-  =KEYWORDS=, and =FILE-TYPE= arguments are given by the renaming
-  command and are used to construct new front matter values if
-  appropriate. If ~denote-rename-confirmations~ contains
-  ~rewrite-front-matter~, prompt to confirm the rewriting of the front
-  matter. Otherwise produce a ~y-or-n-p~ prompt to that effect.
+  =KEYWORDS=, =SIGNATURE=, =DATE=, =IDENTIFIER=, and =FILE-TYPE=
+  arguments are given by the renaming command and are used to construct
+  new front matter values if appropriate. If ~denote-rename-confirmations~
+  contains ~rewrite-front-matter~, prompt to confirm the rewriting of
+  the front matter. Otherwise produce a ~y-or-n-p~ prompt to that effect.
 
 #+findex: denote-add-front-matter-prompt
 + Function ~denote-add-front-matter-prompt~ :: Prompt to add a
--- a/denote.el
+++ b/denote.el
@@ -1850,6 +1850,17 @@ this list for new note creation.  The default is `org'.")
   "Return all `denote-file-types' keys."
   (delete-dups (mapcar #'car denote-file-types)))
 
+(defun denote--get-component-key-regexp-function (component)
+  "Return COMPONENT's key regexp function.
+
+COMPONENT can be one of `title', `keywords', `identifier', `date', `signature'."
+  (pcase component
+    ('title #'denote--title-key-regexp)
+    ('keywords #'denote--keywords-key-regexp)
+    ('signature #'denote--signature-key-regexp)
+    ('date #'denote--date-key-regexp)
+    ('identifier #'denote--identifier-key-regexp)))
+
 (defun denote--format-front-matter (title date keywords id signature filetype)
   "Front matter for new notes.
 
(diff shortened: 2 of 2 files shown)


protesilaos:
Thank you! I will have time to test it either this evening or tomorrow morning.

* * *

Maybe related to this, I want to discuss with you how we can have nil values for many of the keywords in `denote-file-types`. For example, the user wants a Markdown file with just this:

```
# This is the title I wrote
```

Now this will not be handled gracefully by several of our commands. I think they should if the user registers a new `denote-file-types` like, say, `markdown-plain`.

Just mentioning this here for you to keep in mind. It does not affect the PR.

jeanphilippegg:
> Thank you! I will have time to test it either this evening or tomorrow morning.

Take your time! I also found a small regression. If you create a new note and do not save it, `denote-rename-file` will fail. The fix is easy. We should change `insert-file-contents` to `find-file-noselect` in `denote--file-has-front-matter-p`. I will fix it. Make sure to save newly-created notes in your tests before renaming.

### The markdown-plain file type

> Maybe related to this, I want to discuss with you how we can have nil values for many of the keywords in `denote-file-types`. For example, the user wants a Markdown file with just this:
> 
> ```
> # This is the title I wrote
> ```

Try this:

```emacs-lisp
(add-to-list
 'denote-file-types
 '(markdown-plain
   :extension ".md"
   :front-matter "# %s\n\n"
   :title-key-regexp "^# "
   :keywords-key-regexp "^<XF86Travel>" ; Some random string that does not exist in a file.
   :keywords-value-function identity
   :keywords-value-reverse-function identity
   :signature-key-regexp "^<XF86Travel>"
   :signature-value-function identity
   :signature-value-reverse-function identity
   :identifier-key-regexp "^<XF86Travel>"
   :identifier-value-function identity
   :identifier-value-reverse-function identity
   :date-key-regexp "^<XF86Travel>"
   :date-value-function identity
   :date-value-reverse-function identity
   :link denote-md-link-format
```

It works!

But I agree that we should integrate default values and avoid doing unnecessary work when configs are missing. I will look into it!

protesilaos:
Merged, thank you!

jeanphilippegg:
You're welcome!

In case you missed it, I had made a comment in the patch about the way I replicated the behavior of `delete-line` (Emacs 29).

protesilaos:
> Date: Thu, 14 Nov 2024 19:25:41 -0800
>
> You're welcome!
>
> In case you missed it, I had made a comment in the patch about the way
> I replicated the behavior of `delete-line` (Emacs 29).

I did miss it, indeed. Can you point me to it?

-- 
Protesilaos Stavrou
https://protesilaos.com

jeanphilippegg:
I am not able to link to it, but it is visible if you scroll up from this page (on github). Here is what I wrote:

In `denote-rewrite-front-matter`, I wrote

```emacs-lisp
(delete-region (line-beginning-position) (line-beginning-position 2))
```

In Emacs 29, there is a `delete-line` function, implemented like this:

```emacs-lisp
(delete-region (pos-bol) (pos-bol 2))
```

Reading the doc of `pos-bol` (Emacs 29) and `line-beginning-position`, it seems like there is a small difference, but I don't understand it and I don't know if it is relevant in denote's code. My tests of `denote-rewrite-front-matter` worked.

protesilaos:
> Date: Mon, 18 Nov 2024 17:54:23 -0800
>
> I am not able to link to it, but it is visible if you scroll up from this page (on github). Here is what I wrote:
>
> In `denote-rewrite-front-matter`, I wrote
>
> ```emacs-lisp
> (delete-region (line-beginning-position) (line-beginning-position 2))
> ```
>
> In Emacs 29, there is a `delete-line` function, implemented like this:
>
> ```emacs-lisp
> (delete-region (pos-bol) (pos-bol 2))
> ```

This is fine. I also use the 'line-beginning-position' in such cases.
The 'pos-bol' will be relevant for right-to-left scripts like Arabic. I
do not know any right-to-left script, so I am not sure I can test it
properly. If anyone reports a bug, then we can fix it.

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #480, opened 2024-11-16
Signatures in front matter, change id/date and bug fix

jeanphilippegg:
### Signatures in front matters!

I added the signature in the front matter. I think we should keep
`denote-always-include-all-front-matter-lines` to `t` because it is the
current behavior for the keywords line when there are none. It is also less
surprising to users if they get a new signature line than if they have a
keywords line disappear. Users have these options:

- Set `denote-always-include-all-front-matter-lines` to `nil`.
- Remove the signature line from their front matter template.

### Modify the date/id

The new convenience `denote-rename-file-date` allows changing the date/id, as
long as we find no links to the note. A user-error is thrown otherwise. I
think it is better to throw a user-error instead of ignoring silently. When
renaming multiple files in a single operation, it may be an inconvenience, but
it is probably not a good idea to have a date prompt anyway, so the user-error
should not matter here.

### Default values in front matter specification, for markdown-plain

I added default values in `denote-file-types`. You can notice that I opted for
a regexp of a random string in `denote--*-key-regexp`. It leads to the desired
results. I prefer this solution to complicating the rest of the code. This
means that anywhere in the code, we do not have to worry about a possibly
undefined value (we can assume that the value exists). The other keys do not
need to be given a default value.

Now here is the spec of `markdown-plain`:

```emacs-lisp
(add-to-list
 'denote-file-types
 '(markdown-plain
   :extension ".md"
   :front-matter "# %s\n\n"
   :title-key-regexp "^# "
   :link denote-md-link-format
```

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -3377,12 +3377,9 @@ renaming commands."
                           signature
                           (format "Rename `%s' with SIGNATURE (empty to remove)" file-in-prompt))))
         ('date
-         ;; TODO: We currently prompt only if the current file has no
-         ;; identifier. Eventually, we may want to allow modifying the
-         ;; date/id. Then, it will be better to prompt according to
-         ;; `denote-prompts`, like other components (ie remove this
-         ;; condition).
-         (unless (denote-file-has-identifier-p file)
+         (if (and (denote-file-has-identifier-p file)
+                  (denote--file-has-backlinks-p file))
+             (user-error "The date cannot be modified because the file has backlinks")
            (setq date (denote-valid-date-p (denote-date-prompt)))))))
     (list title keywords signature date)))
 
--- a/denote.el
+++ b/denote.el
@@ -1479,6 +1479,7 @@ Denote file-naming scheme."
 #+date:       %s
 #+filetags:   %s
 #+identifier: %s
+#+signature:  %s
 \n"
   "Org front matter.
 It is passed to `format' with arguments TITLE, DATE, KEYWORDS,
@@ -1491,6 +1492,7 @@ title:      %s
 date:       %s
 tags:       %s
 identifier: %s
+signature:  %s
 ---\n\n"
   "YAML (Markdown) front matter.
 It is passed to `format' with arguments TITLE, DATE, KEYWORDS,
@@ -1503,6 +1505,7 @@ title      = %s
 date       = %s
 tags       = %s
 identifier = %s
+signature  = %s
 +++\n\n"
   "TOML (Markdown) front matter.
 It is passed to `format' with arguments TITLE, DATE, KEYWORDS,
@@ -1514,6 +1517,7 @@ Change the front matter format'.")
 date:       %s
 tags:       %s
 identifier: %s
+signature:  %s
 ---------------------------\n\n"
   "Plain text front matter.
 It is passed to `format' with arguments TITLE, DATE, KEYWORDS,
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
Just fixed one commit. It is ready to be merged!

protesilaos:
Merged, thank you!

jeanphilippegg:
> In my discussions with people who use Denote, those who rely on signatures are in the minority. I think it is better to make this an opt-in feature for those who need. This way we will not default to filling notes with empty front matter.

I will replace `denote-always-include-all-front-matter-lines` with `denote-front-matter-components-present-on-empty-value` (or some other name) that would default to '(title keywords date identifier). In fact, it looks like a small adjustment to the code and allows for even more flexibility.

> In the case of multiple files, maybe we can use 'condition-case' to eventually collect a list of files with errors and then report that to the user? I have not tried this before, but I think it is possible.

Yes. I don't know if it is worth to do now. It seems like a bad idea to be prompted for a date for a series a files. Also, I have not mentionned it, but changing the date comes with a performance cost as we effectively call `denote-link-return-backlinks` for each note if there is a date prompt. Maybe we should wait for user feedback and see the use-cases.

> To make it self-documenting, we can change the string to something like "^denoteuserwantsemptyfieldsohandleit". Of course, we can also write a comment/docstring.

I agree!

> So we may check if this should go in denote-md-extras.el and be named 'markdown-obsidian'.

This is fine with me!

protesilaos:
> Date: Mon, 18 Nov 2024 17:40:48 -0800
>
>> In my discussions with people who use Denote, those who rely on
>> signatures are in the minority. I think it is better to make this an
>> opt-in feature for those who need. This way we will not default to
>> filling notes with empty front matter.
>
> I will replace `denote-always-include-all-front-matter-lines` with
> `denote-front-matter-components-present-on-empty-value` (or some other
> name) that would default to '(title keywords date identifier). In
> fact, it looks like a small adjustment to the code and allows for even
> more flexibility.

Good idea! This makes it more flexible and is consistent with what we do
elsewhere.

>> In the case of multiple files, maybe we can use 'condition-case' to
>> eventually collect a list of files with errors and then report that
>> to the user? I have not tried this before, but I think it is
>> possible.
>
> Yes. I don't know if it is worth to do now. It seems like a bad idea
> to be prompted for a date for a series a files. Also, I have not
> mentionned it, but changing the date comes with a performance cost as
> we effectively call `denote-link-return-backlinks` for each note if
> there is a date prompt. Maybe we should wait for user feedback and see
> the use-cases.

Yes, let's wait. I think there would be less of a need for this feature
if our renaming mechanism could use the file's creation time. But I do
not know if there is something else beside what the docs of 'file-attributes'
cover.

>> To make it self-documenting, we can change the string to something
>> like "^denoteuserwantsemptyfieldsohandleit". Of course, we can also
>> write a comment/docstring.
>
> I agree!
>
>> So we may check if this should go in denote-md-extras.el and be named
>> 'markdown-obsidian'.
>
> This is fine with me!

Do you want to include those in the change that will add
'denote-front-matter-components-present-on-empty-value'?

If not, I will make the changes from here.

-- 
Protesilaos Stavrou
https://protesilaos.com

protesilaos:
> Date: Tue, 19 Nov 2024 18:32:15 -0800
>
>> Do you want to include those in the change that will add
>> 'denote-front-matter-components-present-on-empty-value'? If not, I
>> will make the changes from here.
>
> Go ahead! I am less familiar with this part of Denote. You will
> integrate this better than me.
>
> I will include `denote-front-matter-components-present-on-empty-value`
> in a few days.

Thank you!

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #482, opened 2024-11-21
fix obsolete alias

brabalan:
I just saw a compilation warning about this being an obsolete alias.


protesilaos:
Merged, thank you!

PR #483, opened 2024-11-22
Add multiple keywords option for journal

cosmicz:
This pull will allow denote-journal-extras users to add multiple keywords to their journal files

### DIFF
--- a/denote-journal-extras.el
+++ b/denote-journal-extras.el
@@ -52,11 +52,12 @@ the `denote-journal-extras-keyword'."
                  (const :tag "Use the `denote-directory'" nil)))
 
 (defcustom denote-journal-extras-keyword "journal"
-  "Single word keyword to tag journal entries.
+  "Single word keyword or lists of keywords to tag journal entries.
 It is used by `denote-journal-extras-new-entry' to add a keyword
 to the newly created file."
   :group 'denote-journal-extras
-  :type 'string)
+  :type '(choice (string :tag "Keyword")
+                 (repeat :tag "List of keywords" string)))
 
 (defcustom denote-journal-extras-title-format 'day-date-month-year-24h
   "Date format to construct the title with `denote-journal-extras-new-entry'.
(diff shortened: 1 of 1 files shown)


cosmicz (commented):
Happy to address any issues and get this over the finish line. Thanks for creating denote!

protesilaos:
Merged, thank you!

> denote-journal-extras--filename-date-regexp

This function hardcodes the pattern for a file name. It is fine with the out-of-the-box settings but will be problematic for users who configure the `denote-file-name-components-order`. The function `denote-format-file-name` takes care to put things in the right order. Maybe we can adapt that approach?

> These should maybe check if the file is in the journal directory

I do not mind this. Though I imagine if the user has moved their journal entry into another directory they probably had a good reason for it? Let me also ping @brabalan who uses the journaling feature and may have a better intuition about it.

brabalan:
I have all my journal entries in a `journal` subdirectory (so that I can easily browse them using `dired`), and they also have a `journal` tag. Looking at the code, this commit will not change anything for me. If I notice that something is amiss, I’ll report linking this PR.

protesilaos:
> From: Alan Schmitt ***@***.***>
> Date: Sat, 23 Nov 2024 00:51:06 -0800
>
> I have all my journal entries in a `journal` subdirectory (so that I
> can easily browse them using `dired`), and they also have a `journal`
> tag. Looking at the code, this commit will not change anything for me.
> If I notice that something is amiss, I’ll report linking this PR.

Fine. Then @cosmicz please go ahead with the extra check for the
directory.

Also @cosmicz, I think you have not assigned copyright to the Free
Software Foundation, right? You will need it eventually to continue
sending PRs. In the manual we provide the form you need to send to the
address specified therein: https://protesilaos.com/emacs/denote#h:1ebe4865-c001-4747-a6f2-0fe45aad71cd

The process is fairly simple: you send them the message and they then
ask you to sign a PDF and you file back to them.

-- 
Protesilaos Stavrou
https://protesilaos.com

Issue #474, opened 2024-11-04
Make it possible to rename/change the identifier


protesilaos:
We have mentioned the titular idea before and we have the building blocks to make it happen. Modifying the identifier can be problematic, because it may result in duplicates (we have code to avoid this inside of the `denote-directory`, but it can still happen if the user renames files elsewhere and then consolidates them in a single directory). If there are such cases, we have to trust the user to resolve them or, maybe, we can provide a command that does that automatically.

I think this is the right time to provide the relevant points of entry. @jeanphilippegg You recently made changes to the `denote-files-types`. Are you also planning something along those lines? If yes, then I will wait for it. Otherwise, I will go ahead with some code and then we can iterate on it.

jeanphilippegg:
Actually, yes!

I had the following 2 changes in mind before that:

- Cleanup the date handling functions. Since it is closely related to the id,
  I would make it consistent. There are some unecessary functions that I would
  like to address.

- Refactor `denote-rewrite-front-matter` and add the signature. The refactor
  involves handling all front matter components consistenty and address the
  points you mentionned in #366. We should be able to add/remove individual
  lines automatically when they are present. The signature can then be added
  to the front matter! This will also be useful to modify the front matter
  id/date.

After the above changes, I would have worked on allowing the modification of
the identifier (under the condition that there are no links to the note) and
maybe allow its absence. I want to make this possible using what we already
have for other components. Changing the date with `denote-rename-file` should
just change the date/id as expected (it does nothing right now). I would not
add any new commands specific to date/id, except maybe the convenience
`denote-rename-file-date`!

I planned on pushing some of these changes next weekend.

Did you have anything specific already on this?

juh2:
In this context I would like to ask if you plan to make the date identifier more precise. Some years ago I bulk created 10.000 org-files with a script and I had to add milliseconds to the filename part that took the current date+time as identifier of the filename. Otherwise I would end up with files with the same timestamp. These are older org-files I manage in org-roam. Now that I want to migrate to denote, I run into problems as the migration skript at hand (https://github.com/bitspook/notes-migrator.git) does not support milliseconds and if I am unaware of a millisecond option in denote anyway.

I may solve this issue by reimporting the files from a sqlite database where I also store the information with a wait one second function, but then the script would run for several hours. 

If denote had an option to evaluate a longer timestring in the filename I could just try to import them with the bitspook script or to newly export them from the database.

jeanphilippegg:
[...]

Issue #366, opened 2024-05-20
Suggestion where to place `SIGNATURE` in the front matter


maikol-solis:
Hi Prot!

I'm an org-roam user which is trying denote, and so far, I like a lot. 

My suggestion is to put the `SIGNATURE` in the subtitle of the front matter. I use it this way in org-roam and worked for me perfectly. 

For example. 

Markdown 
```markdown
---
title:      "My title"
subtitle:   "7a1"
date:       2024-05-14
tags:       []
identifier: "20240514T135224"
---

```

org-mode

```org
#+title:      My title
#+subtitle:   7a1
#+date:       [2024-05-14 Tue 13:52]
#+filetags:   
#+identifier: 20240514T135224
```

WDYT? 

Best.

maikol-solis:
Thanks Prot. Fair enough. Yes, I agree that version 3.0 has a lot of exciting features to add a new one.

With respect to the subtitle, I normally use my notes as intermediate products before start drafting a real work. But I could understand that other could have other workflows.

Maybe I could tweak the capture functions to add the subtitles in my case.

Thanks.

Best.

MirkoHernandez:
@maikol-solis  Could you elaborate why there is a need to have the signature in the front-matter? I'm currently using `denote-renamer-buffer-mode` to display the signature along the title (If you just need a visual cue, this might be the solution).

Also, I've been using  the package `citar-denote` which seamlessly is to be able to add entries to the front-matter. Maybe you could check that package's source.

maikol-solis:
My main goal is to put the fogelzettel in the front matter. This helps me to identify where I am in my own zettelkasten. The citar-denote it's a good resource. I'll check it out.

MirkoHernandez:
@maikol-solis  `denote-rename-buffer-mode` can be used for that particular use case, it can display the signature in the modeline. Unless your are using unusually long signatures I think that it should work. Here is how to change the format to display the signature.

	`(setq denote-rename-buffer-format  "%s %t")`

protesilaos:
> From: Maikol Solís ***@***.***>
> Date: Tue, 25 Jun 2024 12:52:45 -0700
>
> My main goal is to put the fogelzettel in the front matter.

I think it will be a nice addition to have the SIGNATURE in the front
matter. The tricky part is to make this backward compatible with the
existing front matter. We also want to insert a signature only if there
is one, rather than have the field there unconditionally and leave it
without a value.

-- 
Protesilaos Stavrou
[...]

Issue #278, opened 2024-03-08
Hacking `denote-rename-file` to allow changing the ID


mentalisttraceur:
I often want to change the ID part of a Denote-style file name while reusing all of the logic in `denote-rename-file`.

---

Currently I achieve this by advising `denote-retrieve-filename-identifier` to return the desired ID during the call to `denote-rename-file`. In other words:

```lisp
(with-advice ('denote-retrieve-rename-file :override (lambda (&rest _) new-id))
    (denote-rename-file ...))
```

(`with-advice` calls `advice-add` and its body in an `unwind-protect`, with the corresponding `advice-remove` in its cleanup clause.)

Of course, this only replaces the datetime in the file name. For files with front matter, this does *not* update the front matter's `date` and `identifier` fields. That would require much more involved code and I personally haven't needed it.

---

My use case for things like this is:

1. I have a basic task list system based on Denote where each task is a note file. For scheduled-for-a-specific-time tasks I use the datetime in the file name to indicate when the task is to be done. For unscheduled tasks, I just use this to sort tasks relative to each other. (Here, I benefit most heavily from the automatic Dired refresh, and sometimes from the ability to do this from the task's note or to have the task's not open in the background but have the buffer refresh automatically. My code disables Denote's confirmation prompts in these "task reschedule" functions, so I wouldn't miss that.)

2. Fixing typos in the names of pictures or other files where I might need to manually type the date when first converting them to Denote's file naming convention. More generally: the more I go "all in" on using Denote's file naming convention for all sorts of files, the more relevant it is for me to have quick and easy ways to add, change, or remove the ID in a Denote-style file name.

3. It's nice to have as an option precisely because having it available helps me creatively notice future situations where it would be useful.

4. Importing my blog posts into Denote notes. Right now I have a lot of blog posts which were written before I switched to Denote, which only exist online. With my current setup, it's not extremely ergonomic to create a new note without even thinking about the date yet, and *then* paste in the post, add the tags, and change the ID to reflect the original publish date.

The benefit of going through `denote-rename-file` is I get all the other nice features: the visiting buffer renames to match, all Dired buffers that have the file's directory open refresh, and so on (with my other hack to enable `denote-rename-file` to act even if the file doesn't exist yet but there's a visiting buffer for it, a newly created note doesn't even have to be saved yet for me to change any part of its name - and with this hack, that really means *any* part of its name).

[...]

PR #484, opened 2024-11-23
Small fixes, `denote-rename-file-using-front-matter` and signature line only when there is a value

jeanphilippegg:
In this pull request:

- `denote-front-matter-components-present-even-if-empty-value`
  defaulting to `'(title keywords date identifier)`. The signature is
  not included so things should keep working as before for users.

- Fixed smalls issues in `denote--get-file-components-for-rewrite` and
  `denote-rewrite-front-matter`.

- Moved the backlinks check in `denote--rename-file`.

- Refactored `denote-rename-file-using-front-matter`. It was not
  handling all front matter components. Though it was a general issue,
  I think you noticed it when trying to integrate `markdown-obsidian`.
  It should work now. Check its docstring!

------

### markdown-obsidian

I have seen your obsidian branch. I have two small comments:

1- The `denote-*-key-regexp` should not be empty strings for the
missing components, or else it will match. The default value ensures
that there is no match.

2- `#'identity` is not a suitable function for the
`denote-keywords-value-function` and `denote-date-value-function`. The
reason is that the functions should take a list or a date object as
input and return a string for the front matter. `#'identity` does not
do that. I know I gave this example in a previous comment. I realized
it was not valid when I tried to give them default values. It may work
anyway because there is no date or keywords in the front matter
template.

So, a correct definition would be:

```emacs-lisp
(add-to-list
 'denote-file-types
 '(markdown-plain
   :extension ".md"
   :front-matter "# %s\n\n"
   :title-key-regexp "^# "
   :link denote-md-link-format
```

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -3256,6 +3256,8 @@ prompt to confirm the rewriting of the front matter."
                            (goto-char (line-beginning-position))
                            (insert new-line)
                            (delete-region (point) (line-end-position))
+                           (goto-char (line-beginning-position 2)))
+                          (t
                            (goto-char (line-beginning-position 2))))))))))))))
 
 ;;;;; The renaming commands and their prompts
--- a/denote.el
+++ b/denote.el
@@ -5027,7 +5027,7 @@ Place the buffer below the current window or wherever the user option
 ;;;;; Add links matching regexp
 
 (defvar denote-link--prepare-links-format "- %s\n"
-  "Format specifiers for `denote-link-add-links'.")
+  "Format specifiers for `denote-add-links'.")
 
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
Just pushed one last commit.

Ready to be merged!

jeanphilippegg:
I refactored `denote--format-front-matter` to remove lines with empty
values (if necessary) from new front matters.

I rebased the branch on top of main.

I think this completes these issues: #474, #366 and #278.

protesilaos:
Merged, thank you!

> I have seen your obsidian branch.

Thank you for the comments! I realised I was using an older version of Denote, which was messing up with my experiment. I just pushed it to the branch to not lose my progress. Will pick it up again as soon as possible.

PR #486, opened 2024-12-02
fix org dblock definition of missing-links

EFLS:
The Org dynamic block defining function mentioned the wrong function to create the block. A simple mistake that is now corrected.


protesilaos:
Merged, thank you!

PR #495, opened 2024-12-14
Small adjustments for notes without identifier

jeanphilippegg:
In this pull request:

- I have made adjustments to the recent code about `initial-date`. It
  was not working with `denote-generate-identifier-automatically` and
  notes without identifiers.

- In a recent pull request (#476), I reworked the internal date
  handling functions so that we always only have to work with dates
  that are represented as date objects (or nil for "no date"). I think
  it makes sense to keep this consistent throughout the code. With
  that in mind, `denote--date-convert` can benefit from this
  simplification, as it is an internal function.

- `denote-date-prompt` now accepts an `initial-date` parameter. The
  docstring says that the type of this parameter should depend on the
  value of `denote-date-prompt-use-org-read-date`. This is not really
  true, as the current code will work with strings or lists,
  regardless of `denote-date-prompt-use-org-read-date`. As such, I
  have adapted the docstring to just say that it accepts lists or
  strings. As a public function, it is also okay that it accepts both
  strings and lists, but I made sure to convert it to an internal date
  object right away, to avoid spreading strings down to internal
  functions.

### DIFF
--- a/README.org
+++ b/README.org
@@ -5949,11 +5949,10 @@ might change them without further notice.
   ([[#h:e7ef08d6-af1b-4ab3-bb00-494a653e6d63][The denote-date-prompt-use-org-read-date option]]). With optional
   =INITIAL-DATE= use it as the initial minibuffer text. With optional
   =PROMPT-TEXT= use it in the minibuffer instead of the default
-  prompt. When ~denote-date-prompt-use-org-read-date~ is non-nil, the
-  value of =INITIAL-DATE= is of the format understood by
-  ~org-read-date~. Otherwise, it is a string that can be processed by
-  ~denote-valid-date-p~. [ The =INITIAL-DATE= and =PROMPT-TEXT= are
-  part of {{{development-version}}}. ]
+  prompt. =INITIAL-DATE= is a string that can be processed by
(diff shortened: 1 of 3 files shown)


protesilaos:
Merged, thank you!

Not related to this particular changeset, I think we should have an entry in the manual about this new direction with the identifiers. It will be for advanced users, of course. The idea is to cover the possibillities that are now available. What do you think? Or do we leave it to experts to figure it out by reading the source code?

jeanphilippegg:
Yes, it makes sense to include it in the manual. But there are still issues with notes without identifiers. For example, such notes are not included in `denote-directory-files`, which makes them unlisted in backlinks buffers. I will look into fixing this soon.

protesilaos:
> Date: Sat, 14 Dec 2024 20:19:47 -0800
>
> Yes, it makes sense to include it in the manual. But there are still
> issues with notes without identifiers. For example, such notes are not
> included in `denote-directory-files`, which makes them unlisted in
> backlinks buffers. I will look into fixing this soon.

Sure, this is not a pressing issue. We do not have to even mention it
right now and let it be an undocumented feature for experts until we
think we are ready for the next step.

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #496, opened 2024-12-14
Fix tests

jeanphilippegg:
The regressions come from #480 where I added the signature in the front matter and #476 where I made `denote-format-file-name` allow nil date/id.

I fixed the tests!

I also added another validation in `denote-format-file-name` and `denote-file-has-denoted-filename-p`.


protesilaos:
Merged, thank you!

PR #497, opened 2024-12-15
Small adjustments

jeanphilippegg:
In this pull request:

- `denote--rename-file` is now able to rewrite a front matter even if the file name stays the same. This is useful to fix a front matter.

- I made the fontification use `denote-file-has-denoted-filename-p` instead of `denote-file-has-identifier-p`. This makes files without identifiers fontified. I don't think there would be performance issues, even though it involves more computation in the fontification. It is easy to revert if users report issues.

- Other minor tweaks (see the commits)


jeanphilippegg:
Just added a few minor commits. I added `denote-file-is-in-denote-directory` and replaced `denote-filename-is-note-p` with `denote-file-has-supported-extension-p` where it made sense.

It is ready to be merged!

protesilaos:
Merged, thank you! If we notice anything if the fontification performance, then here we are to deal with it. No problem!

Issue #500, opened 2024-12-21
`denote-rename-file` and `denote-rename-file-using-front-matter` behave incorrectly in `org-capture` buffer


jeffvalk:
First of all, thanks for Denote and all the thought that has gone into it!

Here is what I'm seeing...

### Unexpected behavior

When in an `org-capture` buffer:

- Invoking `denote-rename-file` produces an unexpected directory prompt, and
- Invoking `denote-rename-file-using-front-matter` errors with `(wrong-type-argument stringp nil)`.

I think these functions do not check whether the buffer currently has a file (e.g. `denote--file-type-org-extra-p`).

### Steps to reproduce with `emacs -Q`:

```elisp
(require 'denote "path/to/denote-3.1.0/denote.el")

;; example org-capture template from manual
(with-eval-after-load 'org-capture
  (add-to-list 'org-capture-templates
               '("n" "New note (with Denote)" plain
                 (file denote-last-path)
                 #'denote-org-capture
                 :no-save t
                 :immediate-finish nil
                 :kill-buffer t
                 :jump-to-captured t)))

```

1. `M-x org-capture`, and select "New note (with Denote)"
2. Prior to finishing the capture with `org-capture-finalize` (`C-c C-c`), call either of Denote's rename commands:

   - `M-x denote-rename-file` brings up a minibuffer prompt: "Rename FILE Denote-style: "
   - `M-x denote-rename-file-using-front-matter` produces an error: `(wrong-type-argument stringp nil)`

### Expected result

I would expect the following behavior in an `org-capture` buffer, which is not yet saved to a file:

- `denote-rename-file`
   - Prompt according to `denote-prompts`
   - Update front matter
   - Update the value of `denote-last-path`

- `denote-rename-file-using-front-matter` 
   - Update `denote-last-path`

Then when `org-capture-finalize` is called, the buffer should be written by `org-capture` to a file with the updated file name stored in `denote-last-path`.

jeanphilippegg:
Thank you for the detailed issue!

> Then when `org-capture-finalize` is called, the buffer should be written by `org-capture` to a file with the updated file name stored in `denote-last-path`.

I have tried to make renaming commands work in org-capture buffer on my fork (branch `denote-org-capture`), but it seems like updating `denote-last-path` in `denote-rename-file` is too late, meaning that the initial path is always used.

[...]

PR #502, opened 2024-12-22
Make renaming commands work in denote-org-capture

jeanphilippegg:
For issue #500. See discussion there.


protesilaos:
Merged, thank you!

We could simply say that we do not cover this use-case, though that would have felt like an arbitrary restriction. This is better.

PR #505, opened 2024-12-28
Update README.org

lgatto:
- From the text, there seem to be more than five (I count seven) to write a note with Denote.
- Typo - remove 's'.


protesilaos:
Merged, thank you!

PR #514, opened 2025-01-03
In `denote-rename-file', validate a given date before renaming.

wlharvey4:
If the user supplies a string calendar date to `denote-rename-file`, the date is sent through without validation.  If the user has supplied a string calendar date, such as `2025-01-03T08:00`, there will be an error because the process expects a timestamp instead.  This change first validates the date by using `denote-valid-date-p`, which turns the string date into a timestamp, and then proceeds with the file renaming process.  If the provided date is invalid, an error will result.


protesilaos:
Thank you @wlharvey4! I just merged [PR 515](https://github.com/protesilaos/denote/pull/515) which does the same thing slightly differently. I will still add your name to the manual for code-related contributions.

Issue #513, opened 2025-01-02
Error when renaming an Org file into a Denote file with Date being a string of the form "YYYY-MM-DDTHH:MM"


wlharvey4:
In my case, an Elisp program saves an Org file without an identifier or front matter in a Denote directory and then renames the file to be a Denote note file with an identifier and front matter by calling the function `denote-rename-file` with the required arguments.  Until recently, the program worked with the Date argument to `denote-rename-file` being a string of the form "YYYY-MM-DDTHH:MM".  Now, an error is raised, saying that the Date is not valid.

    denote-rename-file("/path/to/file.org" "New Title" ("keyword1" "keyword2") "new-signature" "2016-02-01T12:00")
    denote--rename-file("/path/to/file.org" "New Title" ("keyword1" "keyword2") "new-signature" "2016-02-01T12:00")
    denote-get-identifier("2016-02-01T12:00")
    (if date (format-time-string denote-id-format date) "") ;; <<== Error because 'date' is a calendar string, not a timestamp
    Debugger entered--Lisp error: (error "Invalid time specification")

Date used to be converted into a timestamp by `denote-valid-date-p`.  However, on Nov 8 and 9, `denote-create-unique-file-identifier` was obsoleted, and in its place `denote-get-identifier` was substituted.  `denote-get-identifier` does not use `denote-valid-date-p` and so the string date is never converted into a timestamp.  `format-time-string` requires a timestamp, and raises an error if the Date argument is a string instead.

See commit a20455 and commit 191643 for the changes.

The error is eliminated when `denote-get-identifier`  is written with `denote-valid-date-p` because this function still converts a string calendar date into a timestamp of the form `(dddddd,dddddd)`:

    (defun denote-get-identifier (date)
       "Convert DATE into a Denote identifier using `denote-id-format'.
         If DATE is nil, return an empty string as the identifier."
        (if date
            (format-time-string denote-id-format (denote-valid-date-p date)) ;; <<== the fix
          ""))

However, since Date is already in the correct form (the form that `format-time-string` is creating), it might make sense to check that it is a calendar string and then simply use it as is, instead of making two conversion to get right back where it started.

jeanphilippegg:
I will make a pull request tomorrow to include `denote-valid-date-p` in `denote-rename-file` (among other minor changes).

In the meantime, you can modify your script to use `denote-valid-date-p`.

PR #515, opened 2025-01-04
Some fixes

jeanphilippegg:
In this pull request:

- Fixed issue #513.
- `denote-file-has-denoted-filename-p` does not validate the presence of an identifier. So we also have to check that `denote-file-has-identifier-p` in `denote-link--map-over-notes` because linking requires identifiers.

I wanted to include other small changes. For example, handle the `keep-current` parameter in `denote--rename-file` instead of its current location, but I did not have time. I will do it later.


protesilaos:
Merged, thank you!

protesilaos:
> I will do it later.

Sure, no pressure!

PR #517, opened 2025-01-04
Add identifier comparison sort function in denote-sort-files

maikol-solis:
The function `denote-sort-dired` wasn't working due to the `identifier` missing case in `denote-sort-files`. This is useful when you customize `denote-file-name-components-order`.


protesilaos:
Merged, thank you!

maikol-solis:
You're welcome.

Issue #516, opened 2025-01-04
Wrong number of arguments in `sort` function call in `denote-sequence.el`


migliio:
Hello Prot!
When I launch the `denote-sequence` command over a note in `dired`, I get a `wrong-number-of-arguments` error, specifically for the call to the built-in `sort` function. Here's the stack trace:
```
Debugger entered--Lisp error: (wrong-number-of-arguments #<subr sort> 5)
  sort(("1" "2" "4" "5" "6" "3" "7" "8" "9") :lessp #f(compiled-function (s1 s2) #<bytecode 0x10c401d9605ae5ef>) :reverse t)
  denote-sequence--get-largest(("1" "2" "4" "5" "6" "3" "7" "8" "9") parent)
  denote-sequence--get-new-parent()
  denote-sequence-get(parent nil)
  denote-sequence(parent nil)
  funcall-interactively(denote-sequence parent nil)
  command-execute(denote-sequence record)
  #<subr execute-extended-command>(nil "denote-sequence" nil)
  ad-Advice-execute-extended-command(#<subr execute-extended-command> nil "denote-sequence" nil)
  apply(ad-Advice-execute-extended-command #<subr execute-extended-command> (nil "denote-sequence" nil))
  execute-extended-command(nil "denote-sequence" nil)
  funcall-interactively(execute-extended-command nil "denote-sequence" nil)
  command-execute(execute-extended-command)
```
By looking at the `sort` function prototype in the documentation, I see
```
sort is a built-in function in ‘C source code’.

(sort SEQ PREDICATE)
```
while in `denote-sequence` it is called with three arguments. Maybe a version mismatch between my Emacs (29.4) and the one you used to test the package? At least, this is my guess. Let me know if I can give you more details to reproduce the bug.

Thanks,

Claudio

protesilaos:
> From: Claudio Migliorelli ***@***.***>
> Date: Sat,  4 Jan 2025 03:52:04 -0800
>
> Hello Prot!

Hello Claudio!

> When I launch the `denote-sequence` command over a note in `dired`, I get a `wrong-number-of-arguments` error, specifically for the call to the built-in `sort` function. Here's the stack trace:

> [... 16 lines elided]

> By looking at the `sort` function prototype in the documentation, I see
> ```
> sort is a built-in function in ‘C source code’.
>
> (sort SEQ PREDICATE)
> ```
[...]

PR #518, opened 2025-01-04
Use older-compatible `sort` call in `denote-sequence--get-largest`

migliio:
As pointed out in #516, the `sort` function seems to have a mismatching paradigm depending on the Emacs version the code is run on. However, the `denote-sequence--get-largest` implementation was simply using the `:in-place` key as third argument, while the original implementation of `sort` only accepted two arguments, which is not necessary. The sorting is done in place anyhow. This commit modifies the `denote-sequence--get-largest` function to be compatible with older Emacs versions.


protesilaos:
Merged, thank you!

PR #522, opened 2025-01-11
Adoption procedure for existing Denote files

pprevos:
Hi Prot,

I have added two functions to adopt existing files into a sequence family. Either add it as a new parent or as a sibling / child of an existing note.

denote-sequence-adopt: Adopt the current Denote buffer as part of a sequence family.

denote-sequence--adopt-file: Add Denote FILE as new parent or add as child or sibling or existing sequence.

Hope you find it useful.

### DIFF
--- a/denote-sequence.el
+++ b/denote-sequence.el
@@ -575,5 +575,39 @@ the target sequence."
          (new-sequence (denote-sequence--get-new-child target-sequence)))
     (denote-rename-file current-file 'keep-current 'keep-current new-sequence 'keep-current)))
 
+(defun denote-sequence--adopt-file (file)
+  "Add Denote FILE as new parent or add as child or sibling or existing sequence.
+Three options:
+- parent: Add a root sequence to the Denote file
+- child: Add the Denote file as a child of chosen sequence
+- sibling: Add the Denote file as a sibling of chosen sequence"
+  (let* ((file-type (denote-filetype-heuristics file))
+         (selected-type (denote-sequence-type-prompt))
+         (relative (when (not (equal selected-type 'parent))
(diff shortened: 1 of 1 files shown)


protesilaos:
> From: Peter Prevos ***@***.***>
> Date: Sat, 11 Jan 2025 12:22:07 -0800
>
> Hi Prot,

Hello Peter,

> I have added two functions to adopt existing files into a sequence
> family. Either add it as a new parent or as a sibling / child of an
> existing note.
>
> denote-sequence-adopt: Adopt the current Denote buffer as part of a
> sequence family.
>
> denote-sequence--adopt-file: Add Denote FILE as new parent or add as
> child or sibling or existing sequence.

Thank you! Please wait a little longer before adding any extra
functionality to denote-sequence.el. I want to get the basics right so
that we can then extend everything with consistency.

Anything that involves re-sequencing will get tricky once we want to
deal with entire trees. For example, re-parenting '1=1' to become '2=1'
must eventually apply to all children of '1=1', such that '1=1=1' is
turned into '2=1=1'. We are not there yet and we need to be careful with
how much power we give users at this early stage.

I plan to work on such a recursive operation. Once it is done, we are
ready to proceed.

-- 
Protesilaos Stavrou
https://protesilaos.com

pprevos:
Hi Prot,

Deal with entire trees is indeed complex. Even Luhmann would not dare to renumber his system.

My main interest is the ability to adopt existing notes into a sequence. Adoption should not require renumbering.

I will update my visualisation tools after you have released `sequence.el`.

P:)

protesilaos:
Since `denote-sequence` is its own package now, I am closing this PR. We can review it there, if you want.

Issue #526, opened 2025-01-14
Maybe bug in `denote-sequence--get-new-child`


migliio:
Hello Prot!

I noticed a strange behavior in `denote-sequence` when trying to add a new child of a note. As far as I understand, when calling every function that creates a new child (e.g., `denote-sequence-new-child-of-current`), the function that actually gets called is `denote-sequence--get-new-child`. I noticed that, sometimes, the creation of the signature goes as intended, while in some cases it seems a sibling gets created, and not a child.
Let me make an example. Creating a new child for a note with signature "1a1" would imply that signature "1a1a" should be created (assuming "1a1" doesn't have any child already). However, I noticed that the created signature is "1a2". By inspecting the `denote-sequence--get-new-child` function, I noticed that the `all-unfiltered` plays an important role:
```
(defun denote-sequence--get-new-child (sequence &optional sequences)
  "Return a new child of SEQUENCE.
Optional SEQUENCES has the same meaning as that specified in the
function `denote-sequence-get-all-sequences-with-prefix'."
  (if-let* ((depth (+ (denote-sequence-depth sequence) 1))
            (all-unfiltered (denote-sequence-get-all-sequences-with-prefix sequence sequences))
            ...
```
In particular, `denote-sequence-get-all-sequences-with-prefix` returns the same signature twice (if the note is opened in a buffer), and I think that later in the code this is interpreted as there are children already for note "1a1". In fact, if the note "1a1" I am about to create the child for is not opened in any buffer, the signature creation goes as intended. To me, simply adding a `delete-dups` may solve the bug:
```
(defun denote-sequence--get-new-child (sequence &optional sequences)
  "Return a new child of SEQUENCE.
Optional SEQUENCES has the same meaning as that specified in the
function `denote-sequence-get-all-sequences-with-prefix'."
  (if-let* ((depth (+ (denote-sequence-depth sequence) 1))
            (all-unfiltered (delete-dups (denote-sequence-get-all-sequences-with-prefix sequence sequences)))
```
In case this makes sense to you, I can create a pull request for it.

Let me know if I need to share more details to make you reproduce the bug.
Best,

Claudio

protesilaos:
Thank you @migliio! Yes, I noticed a regression this morning, but I could not reproduce it after updating the package. Check that you are on the latest commit. Otherwise, I think we need the following diff:

```diff
 denote-sequence.el | 18 +++++++++---------
 1 file changed, 9 insertions(+), 9 deletions(-)

[...]

PR #528, opened 2025-01-14
Fix the regression bug discussed in #526

migliio:
This patch fixes the regression bug in
`denote-sequence-get-all-files`, as discussed in #526. The core idea behind this fix is to delete duplicates within the list of files and buffers in the `denote-directory` matching a certain signature.


protesilaos:
Merged, thank you!

Issue #532, opened 2025-01-17
Adjusting Denote front matter layout


hyperfocus1337:
Hey Prot!!

When updating my Doom Emacs I can see a lot of activity (ie. commits) in the Denote repository, which is nice to see. Hopefully all is going well with you (and your coaching!).

Due to all the updates I'm wondering if I can adapt Denote to the following use case. Consider this my first attempt to at least write down the requirements and consequently report my research findings to see whether it's (already) possible or not (and keep track of it long term).

As described in README.org, notes have their own [front matter](https://github.com/protesilaos/denote/blob/main/README.org#6-front-matter). This is the markdown front matter:
```
---
title:      "This is a sample note"
date:       2022-06-30T16:09:58+03:00
tags:       ["denote", "testing"]
identifier: "20220630T160958"
---
```

I would like to adapt it to the following layout, as that is required for the Obsidian compatibility we spoke about earlier. Also, it seems to be a more common format in static site builders, as well as other documentation.
```
---
title: This is a sample note
date: 2022-06-30T16:09:58+03:00
tags: 
  - denote
  - tags
identifier: 20220630T160958
---
```

Now back to Denote, we seem to have `denote-file-type` and `denote-yaml-front-matter` to potentially manipulate to get the desired outcome.

When looking at `denote-file-type`, there are three entries for `org`, `markdown-yaml`, `markdown-toml` and `text`.

We are only interested in `markdown-yaml`, thus I am only displaying the relevant portion of `denote-file-type` in this issue. Now maybe I need to look a bit deeper, but this is where I am a bit lost (and could use some guidance). The entry seems to refer to a lot of other variables in the codebase, that would also need to be looked into.

```
[...]

Issue #521, opened 2025-01-09
Accessing the title in `denote-org-capture-specifiers`


756D61:
Hello, I'm trying to access the title in `denote-org-capture-specifiers` but it doesn't work

```elisp
(use-package org-capture
  :config
  (add-to-list
   'org-capture-templates
   '("t" "New note (with Denote)" plain
     (file denote-last-path)
     (function
      (lambda ()
        (let ((denote-org-capture-specifiers "%i\n* TODO %\1 \n%?"))
          (denote-org-capture))))
     :no-save t
     :immediate-finish nil
     :kill-buffer t
     :jump-to-captured t)))
```

But it produces this

```org-mode
#+title:      Demo title
#+date:       [2025-01-09 gio 14:14]
#+filetags:   :aw:
#+identifier: 20250109T141450


* TODO %
```

A screenshot is better  

![image](https://github.com/user-attachments/assets/8822c999-15b3-4934-bb89-b710f289feac)

jeanphilippegg:
> After fiddling a little more and reading again the docs and source code, I came up with following solution. Is this a correct approach?

I would say it will work, but it is not really in the "spirit" of how
`denote-use-title` is intended to be used. In principle,
`denote-use-title` is meant to prevent any prompting and use a
specific title. With your code, you will always be prompted for a
title, ignoring the value of `denote-prompts`.

However, I do not have another solution to suggest, so you may prefer
to keep what you have, despite the above.

If you don't mind using `denote` instead of `denote-org-capture`, you
can use the the following code to do what you want:

```emacs-lisp
(defun my-new-note-add-org-todo ()
  (let ((type (denote-filetype-heuristics buffer-file-name)))
    (when (eq type 'org)
      (insert (concat "* TODO " denote-current-title)))))

(add-hook 'denote-after-new-note-hook #'my-new-note-add-org-todo)
```

You may be able to do something similar with
`org-capture-after-finalize-hook`, but I have not tested.

protesilaos:
I made a follow-up to the contribution of @jeanphilippegg which replaces `denote-current-title` with `denote-current-data`, just in case anybody needs to access something other than the title. Here is the variable with its documentation string:

```elisp
(defvar denote-current-data nil
  "Store the current unprocessed data passed to `denote'.
This is an alist where each `car' is one among `title', `keywords',
`signature', `directory', `date', `id', `file-type', `template'.  The
value each of them contains is the unprocessed input (e.g. the title
before it is sluggified).

[...]

PR #537, opened 2025-01-20
New denote-current-title variable

jeanphilippegg:
I have added `denote-current-title` so that the non-sluggified title
is accessible in `denote-after-new-note-hook` and
`denote-after-rename-file-hook`. This can be useful to those with
a custom front matter (not the default one).

This is not needed for other components because they appear identical
in the file name and the front matter and so they can be retrieved
from the file name.

I will give examples in #532 and #521.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -2508,6 +2508,13 @@ no matter the value of `denote-save-buffers'."
       (when do-save-buffer (with-current-buffer buffer (save-buffer)))
       (when do-kill-buffer (kill-buffer buffer)))))
 
+(defvar denote-current-title nil
+  "Store the current non-sluggified title.
+
+This may be used by the hooks `denote-after-new-note-hook' and
+`denote-after-rename-file-hook' to access the current non-sluggified
+title, even if there is no front matter present.")
+
 (defvar denote-use-title nil
   "The title to be used in a note creation command.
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you! Maybe we could replace this with an alist for all the data we read? I cannot think of a use-case right away, but it seems consistent with your idea.

jeanphilippegg:
> Maybe we could replace this with an alist for all the data we read?

I am not sure I understand.

The new variable `denote-current-title` is not meant to be let-bound like `denote-use-*` variables and user options. It is more like a "read-only" variable accessible to hooks, never to be set by users.

protesilaos:
> Date: Mon, 20 Jan 2025 20:13:29 -0800
>
>> Maybe we could replace this with an alist for all the data we read?
>
> I am not sure I understand.
>
> The new variable `denote-current-title` is not meant to be let-bound
> like `denote-use-*` variables and user options. It is more like a
> "read-only" variable accessible to hooks, never to be set by users.

The idea is if we want to have an alist of "current
title/keywords/signature/date/id/whatever". I am not sure we will need
it though, so it is fine as-is.

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #540, opened 2025-01-30
Fix false matches in 'denote-sequence-get-all-sequences-with-prefix'

frnsd:
Hi Prot,

This pull request fixes what I believe to be a bug involving the 'denote-sequence' command.

When I have at least 10 top-level sequence notes in my 'denote-directory' and call 'denote-sequence' to create a new child for the parent with sequence "1", the actual parent ends up being that with sequence "10" if there are exactly ten top-level sequence notes, "12" if there are 12, "100" if there are 100, etc.

This all comes down to 'denote-sequence-get-all-sequences-with-prefix' using 'string-match-p' to test for matches. I think it should instead match the components individually.

This is all with the latest main. If there are other branches that I don't know about which address this, apologies!

Thanks!

### DIFF
--- a/denote-sequence.el
+++ b/denote-sequence.el
@@ -380,10 +380,18 @@ With optional SEQUENCES operate on those, else use the return value of
 `denote-sequence-get-all-sequences'.
 
 A sequence is a Denote signature that conforms with `denote-sequence-p'."
-  (seq-filter
-   (lambda (string)
-     (string-prefix-p sequence string))
-   (or sequences (denote-sequence-get-all-sequences))))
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you!

protesilaos:
By the way, do you want to prepare a test for this case? I already have a few in denote-test.el.

frnsd:
Hi Prot, I added some tests but found a number of issues in the process. ...Particularly, I forgot to update 'denote-sequence-get-all-files-with-prefix' as well! I'm submitting a new pull request momentarily, since I'm not familiar with how to update closed pull requests on Github.

PR #541, opened 2025-02-01
Add tests for false matches addressed by pull request #540 and fix related issues

frnsd:
This adds tests as mentioned in pull request #540. It also fixes similar false matches in 'denote-sequence-get-all-files-with-prefix'.

The latter function is apparently mainly used by 'denote-sequence-find-dired', and in the process of trying that command out, I found a number of bugs involvings its use of ''denote-sequence-get-relative' --- which I fix here as well.


frnsd:
P.S. I do have FSF copyright assignment (you can check my recent contribution to Emacs core).

protesilaos:
Merged, thank you!

PR #542, opened 2025-02-02
fix regex for word separator in keyword

QiangF:


jeanphilippegg:
It is not possible to have underscores present in keywords because the single underscore is used to separate keywords.

As an alternative to the underscore as the word separator in multi-word keywords, I suggest hyphens, spaces or using CamelCase. This requires modifying the default sluggification (`denote-file-name-slug-functions`).

-------

For your info, I once had the idea of changing the separator between keywords to allow for this current use-case, but I changed my mind because:

- The change would not be backward-compatible.
- The single underscore looks better between keywords than the alternative I had in mind.

You can check #271 where this was discussed (among many other things).

jeanphilippegg:
The extra `replace-regexp-in-string` is necessary to ensure that there are no underscores in keywords. Without it, you can create a note `20250101T000000--my-note_tag_1_tag_2.org`. Then `denote-rename-file` (and other commands) would see 4 keywords instead of the intended 2.

> I am planning to get rid of the tag in note filename all together. I think it's probably a bad idea to keep tags in filenames, ones the notes number goes big, and different order of tags will be chaotic to even look at.

Denote is mostly about the file naming scheme. It can be used with image files and PDFs. Such files can only save tags in the file name. However, you can "disable" keywords in Denote if you do this:

- Remove `keywords` from `denote-prompts`.
- Remove the keywords lines from the front matter templates.

Then you can use mode-specific functions to handle tags independently from Denote, and you will not have them in file names.

Issue #172, opened 2023-10-08
Executing `denote-keyword-add` on an encrypted file only keeps the .gpg extension


iensu:
When I run `denote-keyword-add` on an encrypted denote file (for instance `20221114T142453--some-note__work.org.gpg`) the keyword is added correctly, but the `.org` extension is removed. I can work around it by manually changing the extension to `.org.gpg`, but it would be great if Denote handled this use-case as well. 

In order to support encrypted Org files I added the following file type. 

```
(add-to-list 'denote-file-types
               '(org-gpg 
                        :extension ".org.gpg"
                         :date-function denote-date-org-timestamp
                         :front-matter denote-org-front-matter
                         :title-key-regexp "^#\\+title\\s-*:"
                         :title-value-function identity
                         :title-value-reverse-function denote-trim-whitespace
                         :keywords-key-regexp "^#\\+filetags\\s-*:"
                         :keywords-value-function denote-format-keywords-for-org-front-matter
                         :keywords-value-reverse-function denote-extract-keywords-from-front-matter
                         :link denote-org-link-format
                         :link-in-context-regexp denote-org-link-in-context-regexp))
```

I noticed in the source code that `denote-keywords-add` calls `denote-filetype-heuristics` to figure out the file type and that function in turn relies on `file-name-extension` to figure out the extension. This function only takes the part after the final `.` into account. It might be better to look at the extensions in `denote-file-types` first, and then fall back to `file-name-extension`, but I'm not sure what other consequences that would have. 

Anyway, thank you very much for a great package. I use it daily and I feel it has really helped me improve my note taking.

iensu:
I fixed it locally by applying the changes below. It got a bit messy though and I haven't looked into how a change like this would affect other functionality.

```diff
diff --git denote.el denote.el
index 9bc9f5c..179843e 100644
--- denote.el
+++ denote.el
@@ -2123,11 +2123,11 @@ the new front matter, per `denote-rename-file-using-front-matter'."
 
 ;;;;; Common helpers for note modifications
 
-(defun denote--file-types-with-extension (extension)
-  "Return only the entries of `denote-file-types' with EXTENSION.
+(defun denote--file-types-with-extension (file)
+  "Return only the entries of `denote-file-types' with extension of FILE.
 See the format of `denote-file-types'."
   (seq-filter (lambda (type)
-                (string-equal (plist-get (cdr type) :extension) extension))
+                (string-suffix-p (plist-get (cdr type) :extension) file))
[...]

PR #557, opened 2025-03-17
Use denote-get-file-extension in denote--rename-file

eum3l:
Hi, this preserves the inner extension when renaming encrypted files again (#172).


protesilaos:
Merged, thank you!

Issue #550, opened 2025-02-23
Calling `denote-link` in an org-capture buffer does not use the title of the note for the link


brabalan:
When I create a link (with `denote-link`) in a normal org buffer, I get something like this (with the note title):
```
 [[denote:20210528T115530][GPG tricks]]
```

However, creating the same link in an org-capture buffer generates a description that is not the note title:
```
 [[denote:20210528T115530][gpg-tricks]]
```

I suspect the issue lies with `denote-retrieve-title-or-filename`, but I don’t understand why it would work in one case and not the other. Doing an edebug on it reveals that the call `(denote-retrieve-front-matter-title-value file type)` returns nil, but as it is defined as a macro I do not know how to debug it.

jeanphilippegg:
You are probably not using the latest version if you have this behavior. However, the current version has an even worse behavior: it always uses the current file title instead of the linked file.

This is because `denote--file-with-temp-buffer` was changed recently to make it possible to call `denote-rename-file` on an org-capture note. @protesilaos , my idea to fix this would be to introduce a private global variable `denote--linked-file` that `denote-link` would set and unset. When set (and in the context of an org-capture buffer), `denote--file-with-temp-buffer` would use it instead of the current buffer. But maybe there is a better way.

protesilaos:
> From: Jean-Philippe Gagné Guay ***@***.***>
> Date: Sun, 23 Feb 2025 20:28:32 -0800

> [... 7 lines elided]

> This is because `denote--file-with-temp-buffer` was changed recently
> to make it possible to call `denote-rename-file` on an org-capture
> note. @protesilaos , my idea to fix this would be to introduce a
> private global variable `denote--linked-file` that `denote-link` would
> set and unset. When set (and in the context of an org-capture buffer),
> `denote--file-with-temp-buffer` would use it instead of the current
> buffer. But maybe there is a better way.

This seems reasonable. Though I wonder if we should step back and
instead of accommodating this use-case of renaming a capture buffer to
throw an error instead telling the user to first confirm the capture and
then do whatever they want. This way we avoid all these edge cases.

But, anyway, I do not mind if we have a workaround for it.

-- 
Protesilaos Stavrou
https://protesilaos.com

jeanphilippegg:
> This seems reasonable. Though I wonder if we should step back and
> instead of accommodating this use-case of renaming a capture buffer to
> throw an error instead telling the user to first confirm the capture and
[...]

PR #558, opened 2025-03-22
Revert commits about renaming org-capture buffers

jeanphilippegg:
This is following the discussion [here](https://github.com/protesilaos/denote/issues/550#issuecomment-2693343371) where we decided to take a step back about the renaming of notes inside an org-capture buffer because it causes issues.

I reverted the relevant commits.

Issue #550 may still not be resolved, but I cannot reproduce it.


protesilaos:
Merged, thank you!

PR #562, opened 2025-03-24
Respect encryption-file-extensions in get-path-by-id

eum3l:
This fixes a problem with opening links to encrypted `org` files.
For example, if the target file has been exported with the same name, the following error occurs:

```
denote-link-ol-follow: Wrong type argument: stringp, nil
```


protesilaos:
Merged, thank you!

Issue #548, opened 2025-02-13
Is there a `create-if-interactive` option for denote `org-store-link` handling mechanism ?


lyndhurst:
### This is a duplicate

I just found out this issue is a duplicate of #267 , sorry I did not find it earlier (or looked hard enough in the doc).
Closing the issue since it has been dealt with and documented apparently.

---

Hi,

I originally wanted my first post to be a big thank you to Protesilaos and all the contributors of `denote` for building something truly great for many of us who have been wandering for years to find a home for their notes (and idiosyncrasies). Unfortunately I am having a configuration issue, so it will be a cry for help instead :)

I am having trouble understanding how to configure denote identifiers and org ID variables so they work well together. I read the documentation (many times over I think) but I might still be missing something. I am trying to stick to the defaults for my first setup iteration, so it should be rather straightforward to explain what I am failing to achieve.

I have local configurations for my non `denote` Org files, so all configurations I am describing here belong to my global `config.el` (Doom user here), and are designed to manage Org mode behaviour in my `denote-directory` only. By default, I have `denote-org-store-link-to-heading` set to `id`, and, as recommended in the documentation `org-id-link-to-org-use-id` set to nil.

This configuration behaves exactly like I want so far:
1. I have `denote` links stored to conveniently insert them later,
2. along with a `file` link suffixed with the heading CUSTOM_ID which:
  a. lets me navigate to specific heading from my non `denote` org files (agenda mostly),
  b. ignores the `denote-directory` variable which means I can easily insert links in my agenda files to any Denote silo I might need.
3. All of this happens without any additional user solicitation like the `org-store-link` prompt to select an handler, and I can keep "clean" property drawers with only one CUSTOM_ID where it is needed.

The problem arises with my `org-capture` templates. They are pretty simple, and designed to capture mostly TODO items to my agenda files. The way I planned my migration to `denote` I do not intend to mix things up and I will use Denote templates for my notes, and `org-capture` for my tasks, so there should be no interference or unexpected complications. I am trying to be a good boy for now ;) and follow the recommendations:

> If you want my opinion though, be more forceful with the separation of concerns. Decouple your knowledge base from your ephemeral to-do list: Denote (and others) can be used for the former, while you let standard Org work splendidly for the latter—that is what I do, anyway.

Previously (before `denote`), and locally in my agenda files directory, my `org-id-link-to-org-use-id` setting was (and locally is) `create-if-interactive` exactly for the reason stated in the variable help:

[...]

Issue #267, opened 2024-03-03
`org-capture` creates CUSTOM_ID property when `denote-org-store-link-to-heading` is non-nil


krisbalintona:
Hello, I am reporting behavior that I believe is an issue, though I am not entirely sure if change is called for within denote or not.

The behavior is this: when calling `org-capture` in an org file, if `denote-org-store-link-to-heading` is non-nil, then denote will create a new CUSTOM_ID (assuming there isn't already one). This is undesirable, since most times I am calling `org-capture` from a file, I do not want a new CUSTOM_ID to be created.

After some debugging, I realized this was because `org-capture` calls `org-store-link`:
```
...
(annotation (if (and (boundp 'org-capture-link-is-already-stored)
				org-capture-link-is-already-stored)
			   (plist-get org-store-link-plist :annotation)
			 (ignore-errors (org-store-link nil))))
...
```
And `denote-link-ol-store` creates a CUSTOM_ID if one isn't already present (via `denote-link-ol-get-id`).

I'm not sure if this is idiosyncratic behavior or not. In my experience, I've tried no other packages that have produced this unintended behavior. As such, I'm not sure if this calls for a change in my configuration or in denote. Perhaps denote should adhere to the value of `org-id-link-to-org-use-id`? But again, I'm not sure what the proper interaction is to begin with. (I also don't recall, off the top of my head, other packages that implement an equivalent feature, so I can't refer to other existing codebases.)

Also, regarding the versions of these packages, I am on the master branch for both Org and Denote.

protesilaos:
Hello @krisbalintona and thank you for taking the time to report this!

> This is undesirable, since most times I am calling `org-capture` from a file, I do not want a new CUSTOM_ID to be created.

Are these files Denote files or does this happen with all Org files? Is setting `denote-org-store-link-to-heading` to nil a solution?

> I'm not sure if this is idiosyncratic behavior or not. In my experience, I've tried no other packages that have produced this unintended behavior. As such, I'm not sure if this calls for a change in my configuration or in denote. Perhaps denote should adhere to the value of `org-id-link-to-org-use-id`?

[...]

PR #565, opened 2025-03-26
Bypass `org-store-link` called by `org-capture`

TomoeMami:
As the issues (https://github.com/protesilaos/denote/issues/548 and https://github.com/protesilaos/denote/issues/267) mentioned, when we set `denote-org-store-link-to-heading` to non-nil and run `org-capture` in a denote file, the heading where cursor stays will be add a property `CUSTOM_ID`.

According to the `org-capture` function(https://github.com/emacs-mirror/emacs/blob/2d278a0f2e945eef30752550f900c1c88367fb6b/lisp/org/org-capture.el#L678-L684):

```lisp
(let* ((orig-buf (current-buffer))
	   (annotation (if (and (boundp 'org-capture-link-is-already-stored)
				org-capture-link-is-already-stored)
			   (plist-get org-store-link-plist :annotation)
			 (ignore-errors (org-store-link nil))))
	   (entry (or org-capture-entry (org-capture-select-template keys)))
	   initial)
```

 It calls `org-store-link` to get the annotation, and `org-store-link` will call `denote-link-ol-store` in denote files.

So we can add a conditional `unless` function to detect the local variable `orig-buf` set by `org-capture`. 

If it exists, `denote-link-ol-store` should skip creating links (especially `CUSTOM_ID`).

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -5747,6 +5747,7 @@ Also see the user option `denote-org-store-link-to-heading'."
                                 (derived-mode-p 'org-mode)
                                 (denote--org-capture-link-specifiers-p)))
             (heading (denote-link-ol-get-heading)))
+       (unless (consp orig-buf)
         (org-link-store-props
          :type "denote"
          :description (if (and heading-links heading)
@@ -5763,7 +5764,7 @@ Also see the user option `denote-org-store-link-to-heading'."
                  (format "denote:%s::#%s" file-id (denote-link-ol-get-id)))
                 (t
                  (concat "denote:" file-id))))
-        org-store-link-plist))))
+        org-store-link-plist)))))
 
 ;;;###autoload
 (defun denote-link-ol-export (link description format)


TomoeMami:
Sorry, we should use `boundp` instead of `consp`.

TomoeMami:
I realized where I went wrong. Variables created by the `let*` function are not passed into nested functions, but can only be used as parameters for functions within the current scope. The reason why modifying local configurations through `let*` succeeds is because those configurations already exist as global variables - the `let` scope simply temporarily overrides their values within that specific context.

protesilaos:
How does this relate to what `org-id-link-to-org-use-id` says? In principle, we want to create a `CUSTOM_ID` if the user opts into that. Though I have been thinking of setting `denote-org-store-link-to-heading` to `nil` as its default value.

TomoeMami:
🙏Apologies, I just realized that the version of `denote` in this GitHub repository does not match the one I installed via `GNU ELPA`. All my previous responses were based on outdated code. (https://github.com/protesilaos/denote/commit/3bb05f212cc29fa6953e6d703e5e0c2e982882a9)

PR #567, opened 2025-03-28
Fix Org links with location identifiers

eum3l:
Links with location identifiers like `denote:20240611T111809::/apple/` are currently treated as query links.
This fixes the opening of these links and makes them use the correct face.


protesilaos:
Merged, thank you!

PR #569, opened 2025-03-29
Add info in docstring of denote-file-name-components-order

jeanphilippegg:
This documents that `denote-file-name-components-order` should probably not be used in custom code.


protesilaos:
Yes, this is better. Thank you!

Issue #561, opened 2025-03-24
Make `denote:' links support "Query links"


protesilaos:
Currently the `denote:` link mechanism always resolves to a file. This is done by formatting a link with an identifier, which should lead to a unique match.

We do, however, already have the infrastructure to make this more general, such that a match that is NOT direct can be treated as the start of a search query.

I have implemented the PROOF-OF-CONCEPT for Org files in the `query-links` branch, in case anybody wants to take a look.

In short:

- `denote:IDENTIFIER` (like `denote:20250324T074132`) is a unique match so go to the given file path (or return the file path upon Org export).
- `denote:QUERY` (like `denote:Prot`) is not meant to be a unique match, so produce an Xref buffer (basically a `grep` like what we do with backlinks) showing all the results for `QUERY`.

I still need to make further modifications, as explained in the commit below. But this is a good starting point. Any feedback is welcome.

```
commit a612e42567bffc7ef468fb63c24f83897e479a1e
Author: Protesilaos Stavrou <info@protesilaos.com>
Date:   Mon Mar 24 07:35:51 2025 +0200

  Add initial support for query links
  
  I still need to work on the link implementation for file types other
  than Org (in 'denote-fontify-links' and related) and then expand the
  manual to document all the changes herein. Though I want to make this
  commit first to give others the chance to review the code.

 denote.el | 341 +++++++++++++++++++++++++++++---------------------------------
 1 file changed, 157 insertions(+), 184 deletions(-)
```

protesilaos:
Sample screenshots

## Query link with a single word

![Image](https://github.com/user-attachments/assets/6ea8dbb5-cf4a-4d39-808c-bf3b5d571856)

## Query link with multiple words

![Image](https://github.com/user-attachments/assets/3df33501-25ef-47ed-b1cb-ac37038dc539)

## The Org export will not create a link but add plain text instead (exact format to be discussed)

![Image](https://github.com/user-attachments/assets/a15808c3-9c30-4797-8f06-ddea97719e87)

gamecubate:
Hi Prot.

Here are 2 cases I'd like to support that might inform your implementation, in case you had not considered them and as long as they don't fall out of your chosen scope. Note: the second case is more important to me, to the point that I am currently working on a [Deft](https://github.com/jrblevin/deft) remake.

1. Creating a query link for all notes created in January 2025, i.e., those notes w. filename starting with `202501`
`denote:^202501`

[...]

PR #571, opened 2025-03-29
Add some useful features to query buffers

lmq-10:
Hi Prot!  These are the changes I talked about in #561.  Basically, I'm merging most of the features currently provided by `denote-search` into Denote, which (I think) will be useful now that searching is a core part of the package.

A brief overview of what is new:

* There's now an interactive command for searching in file contents.  I called it `denote-grep` so as to prevent name collisions for users of my package, and also because it is somewhat more accurate.  It uses `denote-make-links-buffer` under the hood, so it does the same thing as following a link created by `denote-query-contents-link`.
* There are now several commands for filtering the files displayed in query buffers.  This is so users can easily find what they are searching for.  So, for instance, `denote-grep-exclude-files` prompts for a regexp and hides files that match that regexp; `denote-grep-only-include-files` restricts the search to files matching a regexp instead.  I think the most common use case is to filter based on keywords, so `denote-grep-exclude-files-with-keywords` and its sibling `denote-grep-only-include-files-with-keywords` do that, with completion.
   * The command `denote-grep-clear-all-filters` removes the filters in query buffers.
   * There's also a command for searching in files matched by a previous search: `denote-grep-refine`.
   * I tried to give sane keybindings to all these commands in `denote-query-mode`, but you tell me what do you think.
* There's now a variable governing how file headings in query buffers are displayed: `denote-query-format-heading-function`.  I created a function for it called `denote-query-extract-title` which displays its "pretty" titles instead of the bare file name.  The default, however, remains the same.
* All query buffers now use outline-minor-mode so as to fold files interactively.  I find it very useful to quickly see files matched by a search with `outline-cycle-buffer` (I took the freedom to bind it in `denote-query-mode`).  Also, it allows to use `consult-outline` in the buffer.
* As a side effect, backlinks have gained all of these features too (as they also use `denote-make-links-buffer`).  This allows useful things I haven't thought about, such as displaying the real title of backlinks, filtering the backlinks, searching something in the backlinks, and so on.

I haven't updated the manual as this is a big change and maybe you would want to tweak something.  But I can take care of it.


protesilaos:
Thank you @lmq-10! I am merging it now. There are some tweaks I will make on top. I will take care of the manual as well.

A question about `denote-grep`:

```elisp
  (let (denote-query--omit-current)
    (denote-make-links-buffer query)))
```

I think it makes sense in this case. Do you think we should be doing something similar in other queries?

protesilaos:
I just made some tweaks to `denote.el`. I will update the manual either later today or tomorrow. Now I must stop because the weather is bad.

protesilaos:
I just finished making several tweaks and documenting everything. Thank you!

lmq-10:
Thank you for all your work! Sorry if you had to do a lot of code style fixes.

> I think it makes sense in this case. Do you think we should be doing something similar in other queries?

No, I think it makes sense to exclude current file for most queries. `denote-grep` is an exception because when doing an interactive search the user most likely expects to see all matches, even the ones found in current file (and maybe overlooked).

-----

By the way, something I didn't address in the pull request was how the buffer created by `denote-grep` should be displayed. I wasn't sure whether to reuse `denote-query-links-display-buffer-action` or to create another such variable. Right now it doesn't have a display action associated. What do you think?

protesilaos:
> Thank you for all your work! Sorry if you had to do a lot of code style fixes.

This is fine. Thanks again for your contribution!

> Right now it doesn't have a display action associated. What do you think?

I think this is fine. The other commands generate what I consider an "ancillary window": it is useful in the context of the current file, but may not be as useful in its own right. Whereas the `denote-grep` stands on its own.

This is the same idea for not using a `display-buffer` for `denote-sort-dired` (stands on its own) as opposed to the ancillary window of `denote-query-filenames-link`.

lmq-10:
> The other commands generate what I consider an "ancillary window":
> it is useful in the context of the current file, but may not be as
> useful in its own right. Whereas the denote-grep stands on its own.

I agree! Given that `denote-sort-dired` replaces current window, what
do you think about this in `denote-grep`:

```elisp
(denote-make-links-buffer query nil nil '(display-buffer-same-window))
```

Replacing current window was what `denote-search` did originally.

protesilaos:
> (denote-make-links-buffer query nil nil '(display-buffer-same-window))

This hardcodes the behaviour. It is better to let users have the flexibility. With `denote-dired` users get whatever their `display-buffer-alist` has for that. Is there a use-case to have a different behaviour? I cannot think of one.

I agree though that making `denote-grep` and `denote-dired` behave the same way is better. We can always have a user option for this, if needed.

lmq-10:
> Is there a use-case to have a different behaviour? I cannot think of one.

I cannot either.

> I agree though that making denote-grep and denote-dired behave the same way is better. We can always have a user option for this, if needed.

Well, if flexibility is important, we can create another display-buffer variable for it which defaults to replacing current window. What do you think?

PR #573, opened 2025-04-02
Implement functions for searching in dired files and in a region

lmq-10:
The last search features, I promise :)

This pull request implements the last few things `denote-search` has but Denote hasn't.

* A command for searching something in marked dired files (`denote-grep-marked-dired-files`). This allows great composability with `denote-dired`.
* A command for searching among files whose IDs are referenced in region (`denote-grep-files-referenced-in-region`). Specially useful for "See also" sections and the like, but it really is a versatile command.

I also implemented a display buffer action for both of them and `denote-grep`, which defaults to replacing current window.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -5328,18 +5328,42 @@ means of e.g. `denote-query-exclude-files')."
   "Minibuffer history of content searches performed by `denote-grep'.
 Also see `denote-grep-file-regexp-history'.")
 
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you @lmq-10! I will merge it later today because it started raining now.

protesilaos:
Merged, thank you! I will update the documentation a bit later.

PR #576, opened 2025-04-14
Small fixes

jeanphilippegg:
- I changed the order of the `or` subexpressions in `denote-date-prompt`. It was probably unintended.

- I changed the default value of `denote-use-keywords` to the `default` symbol, because `nil` (ie the empty list) is a valid value for keywords.


protesilaos:
Merged, thank you!

Issue #577, opened 2025-04-14
Windows compatibility of `denote-link-ol-follow`


TomoeMami:
https://github.com/protesilaos/denote/blob/2062286ee90d62d8d7da37725d12e0ff7f01e258/denote.el#L6075

```lisp
(defun denote-link-ol-follow (link)
  "Find file of type `denote:' matching LINK.
LINK is the identifier of the note, optionally followed by a file search
option akin to that of standard Org `file:' link types.  Read Info
node `(org) Query Options'.

If LINK is not an identifier, then it is not pointing to a file but to a
query of file contents or file names (see the commands
`denote-query-contents-link' and `denote-query-filenames-link').

Uses the function `denote-directory' to establish the path to the file."
  (if-let* ((match (denote-link--ol-resolve-link-to-target link))
            (_ (file-exists-p (string-trim-right match ":.*"))))
      (org-link-open-as-file match nil)
    (denote--act-on-query-link match)))
```

On Windows, given the `link` as an identifer `"20250312T161853"` , the result of `(denote-link--ol-resolve-link-to-target 
 link)` is a Windows path `"c:/path/to/the/file.extension"`

However, the result of `(file-exists-p (string-trim-right match ":.*"))` is `nil`, because `(string-trim-right "c:/path/to/the/file.extension" ":.*")` returns `"c"` ,which is the driver letter on Windows.

Maybe we can fix it by this: `(string-trim-right "c:/path/to/the/file.extension" ":[^/]+.*")`

protesilaos:
> Maybe we can fix it by this: `(string-trim-right "c:/path/to/the/file.extension" ":[^/]+.*")`

This seems right to me. I made that change recently to cover the case of "query links".

protesilaos:
Do you want to prepare the pull request?

TomoeMami:
PR created.

protesilaos:
Just merged the PR. We can safely close this. Thank you!

PR #578, opened 2025-04-14
Fix Windows compatibility of `denote-link-ol-follow`

TomoeMami:
On Windows, given the `link` as an identifer `"20250312T161853"` , the result of `(denote-link--ol-resolve-link-to-target 
 link)` is a Windows path `"c:/path/to/the/file.extension"`

However, the result of `(file-exists-p (string-trim-right match ":.*"))` is `nil`, because `(string-trim-right "c:/path/to/the/file.extension" ":.*")` returns `"c"` ,which is the driver letter on Windows.

We can fix it by this: `(string-trim-right "c:/path/to/the/file.extension" ":[^/]+.*")`

https://github.com/protesilaos/denote/issues/577

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -6082,7 +6082,7 @@ query of file contents or file names (see the commands
 
 Uses the function `denote-directory' to establish the path to the file."
   (if-let* ((match (denote-link--ol-resolve-link-to-target link))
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you!

protesilaos:
I made the same change to `denote-get-link-face`.

Issue #592, opened 2025-04-29
denote-context-menu disables context menu entirely


artelse:
In my denote `use-package` config I have this hook:
`:hook (context-menu-functions . denote-context-menu)`

When this is enabled, it disables the entire context menu and throws it back to the `mouse-save-then-kill` functionality.

protesilaos:
Hello @artelse!

This happens on my end as well. I have no idea why! The code looks exactly like what I am using in my `beframe` package: the context menu works for that one.

artelse:
This problem has been intermittent; i.e. sometimes after changing (fixing) something in my config, the context menu got lost. So perhaps it could be a loading order issue? But am unsure of that. At least narrowed it down to the `denote-context-menu` as a reproducible issue.

lmq-10:
This is interesting; normally, the function `context-menu-map` returns a menu map. However, after doing:

```elisp
(add-hook 'context-menu-functions #'denote-context-menu)
```

`context-menu-map` throws a backtrace instead:

```
Debugger entered--Lisp error: (wrong-type-argument char-or-string-p ["Create a note" denote :help "Create a new note in the `denote-directory'"])
```

This is why right-clicking fallbacks to `mouse-save-then-kill` then.

Debugging this was hard because (at least for me) edebug behaves weirdly for mouse clicks. However, I was able to find de culprit. The following code always errors out (you can try it yourself):

```elisp
(let ((easy-menu (make-sparse-keymap "Denote")))
  (easy-menu-define nil easy-menu nil
    denote--menu-contents))
```

Now, it is weird that this errors out, because Prot said it works for `beframe`. So I looked at that package, and I found `beframe--menu-contents` is defined diferently. Compare:

```elisp
(defvar beframe--menu-contents
  '("BEFRAME buffers"
    ["Switch to beframed buffer" beframe-switch-buffer
     :help "Switch to a buffer that belongs to the current frame"
     :enable (beframe--multiple-frames-p)]
```

Versus:

```elisp
(defvar denote--menu-contents
  '(["Create a note" denote
     :help "Create a new note in the `denote-directory'"]
```

One has a string first, the other starts right away with a vector.

So, this patch:

```diff
diff --git a/denote.el b/denote.el
index 9ff12ab..79d1414 100644
--- a/denote.el
+++ b/denote.el
@@ -6069,7 +6069,8 @@ This command is meant to be used from a Dired buffer."
 ;;;; Define menu

 (defvar denote--menu-contents
[...]

PR #594, opened 2025-05-03
Add support for sorting files in query buffers

lmq-10:
Hi Prot! I saw that you recently implemented a new random sorting option for `denote-sort-dired` and Org dynamic blocks. As this is a good feature, I thought it would be a nice chance to finally try to bring sorting capabilities to our query buffers!

An overview:

* There's now a user variable that governs how files in all query buffers are sorted (`denote-query-sorting`). It is passed internally to `denote-sort-files`, so all present features of that function are available right away, and any future development will propagate as well.

* More importantly, there's a command to change sorting interactively, `denote-query-sort-last-search`. I bound it to `S` in `denote-query-mode`.

* I made sure all of this is compatible with backlinks, query links, and `denote-grep`.

The changes are rather small, but I know little time passed since 4.0.0 - so no offence if you want to wait a little more in order to implement those :D


protesilaos:
Thank you @lmq-10! I am merging it right away.

> The changes are rather small, but I know little time passed since 4.0.0 - so no offence if you want to wait a little more in order to implement those :D

The idea was to wait ~2 weeks to check if there are no obvious bugs with the last release. I think we are good now. There is the issue with the context menu, but (i) it will not break most people's configuration and (ii) I am not sure what is happening with it: <https://github.com/protesilaos/denote/issues/592>.

PR #595, opened 2025-05-04
Add optional ‘buffer-name’ argument to ‘denote-sort-dired’

mjkalyan:
Users/developers may wish to supply their own name for the resulting sorted dired buffer, so it makes sense to give them the option here.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1668,7 +1668,8 @@ OMIT-CURRENT have been applied."
     (list sort-by-component reverse-sort exclude-rx)))
 
 ;;;###autoload
-(defun denote-sort-dired (files-matching-regexp sort-by-component reverse exclude-regexp)
+(defun denote-sort-dired (files-matching-regexp sort-by-component reverse exclude-regexp
+                          &optional buffer-name)
   "Produce Dired buffer with sorted files from variable `denote-directory'.
 When called interactively, prompt for FILES-MATCHING-REGEXP and,
 depending on the value of the user option `denote-sort-dired-extra-prompts',
@@ -1712,7 +1713,8 @@ When called from Lisp, the arguments are a string, a symbol among
               ;; it upstream, but even if it is fixed we cannot use it
               ;; for now (whatever fix will be available for Emacs 30+).
               (denote-sort-dired-buffer-name (format "Denote sort `%s' by `%s'" files-matching-regexp component))
-              (buffer-name (format "Denote sort by `%s' at %s" component (format-time-string "%T"))))
+              (buffer-name (or buffer-name
+                               (format "Denote sort by `%s' at %s" component (format-time-string "%T")))))
         (let ((dired-buffer (dired (cons buffer-name (mapcar #'file-relative-name files)))))
           (setq denote-sort--dired-buffer dired-buffer)
           (with-current-buffer dired-buffer


mjkalyan:
For example I have the following function to help me view files with matching keywords:
```lisp
(defun denote-dired-filter-by-keyword ()
  "Open a dired buffer with all Denote notes that contain the
interactively-selected keyword(s). Sort the buffer with the most recent
notes at the top (i.e. by identifier)."
  (interactive)
  (let ((keywords (completing-read-multiple "Filter by keyword(s): "
                                            (denote-keywords))))
    (denote-sort-dired (denote--match-tags-rx keywords)
                       'identifier
                       t
                       nil
                       ;; HERE I'M USING THE PATCHED VERSION OF `denote-sort-dired' TO SUPPLY THIS 5TH ARGUMENT
                       (format "%s Files with keywords: %s"
                               denote-buffer-name-prefix
                               keywords))))```

**Comment by mjkalyan** [comment]
I am waiting on copyright assignment to the FSF.

**Comment by protesilaos** [comment]
> Users/developers may wish to supply their own name for the resulting sorted dired buffer, so it makes sense to give them the option here.

The idea is fine. Though because `denote-sort-dired` is interactive, we want the parameters to be reflected there as well.

I made a change now which (i) takes the different approach of introducing a variable that advanced users can modify and (ii) fixes an issue we have had for a couple of years and was noted there in the comment.

What do you think?

> I am waiting on copyright assignment to the FSF.

Good to know! Though you do not need it for changes that are ~15 lines in total.

**Comment by mjkalyan** [comment]
Your solution is much better for customization, thank you! I can still achieve what I want (for my `denote-dired-filter-by-keyword` example above) by shadowing `denote-sort-dired-buffer-name-function` with a lambda as follows:
```lisp
(defun denote-dired-filter-by-keyword ()
  "Open a dired buffer with all Denote notes that contain the
interactively-selected keyword(s). Sort the buffer with the most recent
notes at the top (i.e. by identifier)."
  (interactive)
  (let* ((keywords (completing-read-multiple "Filter by keyword(s): "
                                             (denote-keywords)))
         (denote-sort-dired-buffer-name-function
          (lambda (&rest _args)      ;; <-------- HERE
            "Use a custom buffer name for the result of ‘denote-sort-dired’ that
shows the user's selected keywords (the filter)."
            (format "%s Files with keywords: %s" denote-buffer-name-prefix keywords))))
    (denote-sort-dired (denote--match-keywords-rx keywords) 'identifier t nil)))
```
which would produce the denote-sort-dired buffer with, for example, the name: "Files with keywords: (recipe spicy)".

Note: here is not exactly the best place to discuss it, but in that example I define `denote-buffer-name-prefix` which is intended for all Denote-related buffer names. I use "‹D›", but the default would be "[D]". Is there any interest in adding that as a user option as well?

mjkalyan:
I just noticed an issue with this change!

We cannot use `(denote-directory)` (or possibly any static name for that matter) as the name of the dired buffer, otherwise calling `denote-sort-dired` **a second time** will produce an error complaining about a killed buffer. I think this is because we kill the **statically-named** buffer which the next invocation attempts to use.

Here is a hack-y patch to fix it, though I'm not great with the specifics of Emacs buffers so maybe there's a better way:

```diff
From 39f7837b4fe49ab917963ee49e9e5738fe769cb0 Mon Sep 17 00:00:00 2001
From: James Kalyan <mjkalyan@proton.me>
Date: Mon, 5 May 2025 16:32:07 -0600
Subject: [PATCH] Patch showing example fix for the static dired buffer name

---
 denote.el | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/denote.el b/denote.el
index 5f6d32d..73972a1 100644
--- a/denote.el
+++ b/denote.el
@@ -1722,7 +1722,7 @@ When called from Lisp, the arguments are a string, a symbol among
         (exclude-rx (or exclude-regexp nil)))
     (if-let* ((default-directory (denote-directory))
               (files (denote-sort-get-directory-files files-matching-regexp component reverse-sort nil exclude-rx)))
-        (let ((dired-buffer (dired (cons (denote-directory) (mapcar #'file-relative-name files))))
+        (let ((dired-buffer (dired (cons (format "denote-dired-%s" (gensym)) (mapcar #'file-relative-name files))))
               (buffer-name (funcall denote-sort-dired-buffer-name-function files-matching-regexp component reverse-sort exclude-rx)))
           (setq denote-sort--dired-buffer dired-buffer)
           (with-current-buffer dired-buffer
-- 
2.49.0
```

protesilaos:
Thank you for noticing this! Your approach should do the trick, though I wonder if we need the custom `revert-buffer-function` anymore. We had another FIXME for it, so maybe this is enough?

```diff
 denote.el | 10 +---------
 1 file changed, 1 insertion(+), 9 deletions(-)

diff --git a/denote.el b/denote.el
index 5f6d32d..9a53802 100644
--- a/denote.el
+++ b/denote.el
@@ -1726,15 +1726,7 @@ (defun denote-sort-dired (files-matching-regexp sort-by-component reverse exclud
               (buffer-name (funcall denote-sort-dired-buffer-name-function files-matching-regexp component reverse-sort exclude-rx)))
           (setq denote-sort--dired-buffer dired-buffer)
           (with-current-buffer dired-buffer
-            (rename-buffer buffer-name :unique)
-            (setq-local revert-buffer-function
-                        (lambda (&rest _)
-                          ;; FIXME 2025-01-04: Killing the buffer has
-                          ;; the unintended side effect of affecting the
-                          ;; window configuration when we call
-                          ;; `denote-update-dired-buffers'.
-                          (kill-buffer dired-buffer)
-                          (denote-sort-dired files-matching-regexp component reverse-sort exclude-rx))))
+            (rename-buffer buffer-name :unique))
           buffer-name)
       (message "No matching files for: %s" files-matching-regexp))))
 ```

It seems to work fine on my end, but I have not tested it thoroughly.

protesilaos:
> Note: here is not exactly the best place to discuss it, but in that example I am using `denote-buffer-name-prefix` which I've intended for all Denote-related buffer names. I've defined it as "‹D›", but the default would be "[D]". Is there any interest in adding that as a user option as well?

You mean to have just the prefix as separate from the `denote-rename-buffer-format`? I see we are doing it for the `denote-rename-buffer-backlinks-indicator`, though that is appropriate because it is added conditionally. If anything, I would rather reduce the options here to just `denote-rename-buffer-format`: it is flexible and we can make it more flexible if we need to.

mjkalyan:
> maybe this is enough?

Yes, that seems to be working just fine for me. I didn't have a need for any special revert behaviour anyway. Not sure if it breaks other people's expectations.

mjkalyan:
> You mean to have just the prefix as separate from the `denote-rename-buffer-format`?

Essentially, yes. The idea is to use this prefix in all the Denote-related buffers, for example:
- regular Denote note files (per `denote-rename-buffer-format`)
- denote-query-mode buffers (backlinks, denote-grep, etc.)
- denote-dired buffers (including those of the kind `denote-sort-dired` produces)
- whatever else I'm missing

It's obviously of low-priority, but it would be nice to have this bit of visual harmony. Reference photos:
Unharmonious:
image
Harmonious:
image

In any case, this is just my opinion, and discussion should probably be moved to an appropriate place. 

Thank you for your time, Prot!

protesilaos:
> The idea is to use this prefix in all the Denote-related buffers

Oh, this is a good idea! I was only thinking in terms of `denote-rename-buffer-mode`. Let's discuss it separately. I am interested to see what we can do.

protesilaos:
Closing this now, since we are done with it. Thank you!

PR #596, opened 2025-05-05
Remove include-date parameter from linking functions

jeanphilippegg:
It does not appear to be used in the code and it is probably better to configure `denote-link-description-format` to include the date in the link description.

I can provide an example if you want!

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -4742,31 +4742,24 @@ The format of such links is `denote-md-link-format'.")
   "Regexp to match an identifier-only link in its context.
 The format of such links is `denote-id-only-link-format'."  )
 
-(defun denote-format-link (file description file-type id-only &optional include-date)
+(defun denote-format-link (file description file-type id-only)
   "Prepare link to FILE using DESCRIPTION.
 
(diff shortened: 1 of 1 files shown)


protesilaos:
Hello @jeanphilippegg!

> It does not appear to be used in the code [...]

It is used in the `denote-org` Org dynamic blocks that insert links.

[ This is one of the disadvantages of splitting the code into many packages, but it is manageable. ]

> it is probably better to configure `denote-link-description-format` to include the date in the link description.

Yes, this is better and is consistent with how we use `let` to extend our code.

jeanphilippegg:
I am not familiar with the code of `denote-org`, but I just checked and I think the `include-date` parameter could be removed there as well. It would probably be simpler to not duplicate the means of customizing the description format. The main package already provides `denote-link-description-format` for that (and it is more flexible).

After that, this pull request could be merged.

protesilaos:
> After that, this pull request could be merged.

Very well! Let me look into that and then proceed with this change.

protesilaos:
Merged, thank you! I will make the changes to `denote-org` right away.

PR #597, opened 2025-05-06
Implement ‘denote-buffer-name-prefix’ discussed as an aside in #595

mjkalyan:
This approach defines ‘denote-buffer-name-prefix’ as a customizable variable defaulting to "[D] ".

Then I define ‘denote-buffer-name’ as a function to wrap buffer names in such that they use ‘denote-buffer-name-prefix’. ‘denote-buffer-name’ also takes the opportunity to allow designating buffers as "special/temporary" or not which adds surrounding asterisks.

Lastly, I have wrapped all the Denote-related buffer names with ‘denote-buffer-name’ so commands like ‘denote-sort-dired’, ‘denote-backlinks’, and ‘denote-grep’ will all use ‘denote-buffer-name-prefix’.

BEFORE MERGING:

- Fix my placement of the definitions inside denote.el. I'm not sure what makes most sense to you.

- decide on a package-version for the introduction of ‘denote-buffer-name-prefix’. I have it set to "TODO" at the moment.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1677,12 +1677,14 @@ must return a string that is appropriate for a buffer name.")
 The FILES-MATCHING-REGEXP, SORT-BY-COMPONENT, REVERSE, and
 EXCLUDE-REGEXP all have the same meaning as `denote-sort-dired'.
 Process them to return the buffer name."
-  (format-message
-   "*denote-dired: regexp `%s'; sort `%s'%s%s*"
-   files-matching-regexp
-   sort-by-component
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you! I am merging this and will then do the tweaks on top.

mjkalyan:
Great, thanks :)

protesilaos:
I made the stylistic change. Will update the manual later, as I need to go now.

PR #598, opened 2025-05-07
Add note about denote-use-template duplicate

### LINKED ISSUE #545: Org Capture duplicating `%?` string at the end of buffer
First of all, thanks for the great work with Denote, I have been finding it intuitive and fun to make use of my notes!

I just wanted to report how I am seeing an odd behaviour where the `%?` cursor position identifier in Org Capture gets duplicated.

When using `denote-use-template` with `%?` as a part of the setup, like below:

```elisp
(setq org-capture-templates
      '(("n" "New note" plain
        (file denote-last-path)
        #'(lambda ()
            (let ((denote-use-title (format-time-string "%b, %Y - Week %V"))
                  (denote-use-keywords '("goal"))
                  (denote-use-template "* Goals\n%?\n\n* Notes"))
              (denote-org-capture)))
        :no-save t
        :immediate-finish nil
        :kill-buffer t
        :jump-to-captured t)))
```

I get the cursor correctly positioned at where `%?` is. But I also get an extra `%?` at the bottom of the capture buffer.

Image

It is possible that I have some other Org Mode setup interfering with this, but could not find anything obvious from my end. Is this potentially an issue with the Org Capture support?

rytswd:
When using `denote-use-template`with `let` binding, it is advisable to set `denote-org-capture-specifiers` to nil or an empty string.

This fixes #545.

### DIFF
--- a/README.org
+++ b/README.org
@@ -893,6 +893,12 @@ sets the Org file extension for the created note to ensure that the
 capture process works as intended, especially for the desired output of
 the ~denote-org-capture-specifiers~.
 
+Also, using ~denote-use-template~ with ~let~ bind could potentially interfere
+with the user option ~denote-org-capture-specifiers~. Along with the ~let~
+bind for ~denote-use-template~, it is advisable to bind
+~denote-org-capture-specifiers~ to nil or an empty string in case of template
+getting duplicated.
+
 [ You may not need ~org-capture~ to do what you want ([[#h:11946562-7eb0-4925-a3b5-92d75f1f5895][Write your own convenience commands]]). ]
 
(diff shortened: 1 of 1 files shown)


rytswd:
I wasn't sure where to leave a comment about this -- I suppose it would make more sense to document near `denote-use-template`, but other notes around each function didn't go into the details of how other variables may need to be taken into account. As `denote-org-capture-specifiers` was already documented extensively, I added a section in there.

I'm happy for anyone to rework this. I just wanted to make sure to provide some clarity in case anyone faces a similar issue.

protesilaos:
Merged. This is fine, thank you!

PR #599, opened 2025-05-07
De-duplicate code in ‘denote-format-buffer-name’

mjkalyan:
i.e. Just make the format string conditional, not the whole format.


protesilaos:
Merged, thank you!

This is a matter of style and I have no strong preference either way.

PR #606, opened 2025-05-19
Add user option 'denote-sluggify-signature-in-frontmatter'

hanwenguo:
This user option controls whether signature should be sluggified or keep the user input as-is in frontmatter (not filename), as some users may prefer.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -496,6 +496,17 @@ and/or when the user invokes the command `denote-date'."
   :package-version '(denote . "0.6.0")
   :type 'boolean)
 
+(defcustom denote-sluggify-signature-in-frontmatter t
+  "Control whether signature should be sluggified in frontmatter.
+
+When non-nil, the signature in frontmatter would be sluggified.
+Otherwise, the signature in frontmatter keeps the same as user input.
+
+The default value is t."
+  :group 'denote
+  :package-version '(denote . "4.1.0")
+  :type 'boolean)
+
 (defcustom denote-org-store-link-to-heading nil
   "Determine whether `org-store-link' links to the current Org heading.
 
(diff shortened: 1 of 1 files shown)


protesilaos:
Hello @hanwenguo!

I am curious why you think we need a user option specifically for the signature. If we are to have one, why not for others as well?

At any rate, you can already inhibit sluggification with something like this:

```elisp
;; Global value...
(setq denote-file-name-slug-functions
      '((title . denote-sluggify-title)
        (signature . identity) ; keep as-is
        (keyword . denote-sluggify-keyword)))

;; OR for this command only...
(defun my-denote-wrapper ()
  (interactive)
  (let ((denote-file-name-slug-functions
         '((title . denote-sluggify-title)
           (signature . identity) ; keep as-is
           (keyword . denote-sluggify-keyword))))
    (call-interactively 'denote)))
```

But maybe I have not understood the use-case, so please let me know.

hanwenguo:
Hi Prot, the use case I was thinking was that: the user might want to refer the raw, unsluggified value in the file. For example, one might want to custom exporting in org-mode to prepend the signature (which can contain uppercase letters or some special symbols) to the title in exported file format. However, the user might still want to keep it sluggified in filename for better compatibility with filesystem and integration with other tools.

But after more carefully thinking about this, now I feel that, if the semantics difference made by unsluggifying is important, user should custom the sluggify function to also reflect that in the filename, and carefully choose the allowed symbols to avoid issue with filesystem. There is no much meaning to add such specific options. If you think so, please close this PR. Thanks for your reply.

protesilaos:
I forgot to close this PR at the time. Sorry! Thank you for your time!

Issue #228, opened 2024-01-22
Links to files not in denote-directory


x61:
Hi All:
I started using denote and have difficulties creating links to files that are not in my default denote-directory. I don't want my default  home folder to be my denote-directory. I just want `~/A/notes` to be my denote directory but I want to create links to `~/A/Photos`, `~/B/Media`, `~/A/Videos`. Is this possible? I know we can create silos but I want to be able to reference/link files in any of the directories. Note that `~/A/notes` is my main directory and is where most of the notes are written.

protesilaos:
Hello @x61!

This will not be possible because `denote:` links are designed to search for the identifier anywhere inside the `denote-directory`. Otherwise, it is impossible to (i) know where the file with the given identifier is and, more generally, (ii) if the identifier we created is unique in the `denote-directory`.

x61:
thank you. is it possible to define multiple `denote-directories` without creating `silos`?  This will allow one to use `denote` to manage files in multiple directories without using the `home` folder as a `denote-directory`.

protesilaos:
It is not possible right now, but I think it is a good idea to work on.

x61:
Thank you, @protesilaos . Perhaps, we could provide an option within silo directories to have its contents visible to other directories.

jeanphilippegg:
In my fork of the code, I have added `denote-directories` which will allow notes to be spread across multiple locations.

Code of my fork will be public this weekend. Pull requests will be made in May.

protesilaos:
> From: Jean-Philippe Gagné Guay ***@***.***>
> Date: Mon, 25 Mar 2024 19:11:35 -0700
>
> In my fork of the code, I have added `denote-directories` which will allow notes to be spread across multiple locations.

Sounds promising! I think the 'denote-directories' is the natural
extension of what we have and will be useful to a lot of people.

Silos will continue to exist for the isolation they provide.

> Code of my fork will be public this weekend. Pull requests will be made in May.

This is good. In the meantime, we can focus on fixing any bugs that may
be reported.

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #608, opened 2025-05-26
Implement denote-extra-directories variable

lmq-10:
Currently, Denote can only link to other files in `denote-directory`. That effectively means every valuable file must be kept in that directory in order to reference it. This is not always desirable or possible. Here I craft a simple solution to the issue, which I hope makes everybody happy: a new variable named `denote-extra-directories`, containing a list of directories with files the user can link to. Any link to an identifier not found in `denote-directory` is then searched among these files and followed if found there.

I created these helper commands:

- `denote-link-external-file`: Like `denote-link`, but only considers files in `denote-extra-directories`.
- `denote-backlinks-extra`: Like `denote-backlinks`, but also considers files in `denote-extra-directories`. A user variable exists so this behaviour is available in `denote-backlinks` too, `denote-backlinks-include-extra-directories`.

I tried to keep things simple. Only links and backlinks are implemented; it's possible to add new functionality on top (such as enabling `denote-grep` and `denote-query` to search in the extra directories), but I wanted to stop here and see if this is the right path to follow.

This should solve issue #228. Pinging @jeanphilippegg as he may have something to say about this.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1340,12 +1340,16 @@ something like .org even if the actual file extension is
       extension)))
 
 (defun denote-get-path-by-id (id)
-  "Return absolute path of ID string in `denote-directory-files'."
+  "Return absolute path of ID string in `denote-directory-files'.
+When `denote-extra-directories' is set, also consider files in such
+directories."
   (let ((files
          (seq-filter
           (lambda (file)
             (string= id (denote-retrieve-filename-identifier file)))
-          (denote-directory-files))))
+          (if (not denote-extra-directories)
+              (denote-directory-files)
+            (denote-extra-directories-files :include-main)))))
     (if (length< files 2)
         (car files)
       (seq-find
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
In my experiments, I ended up making `denote-directory` accept a list of directories. I believe all functionalities could be adapted to work with them.

I did not end up making a PR for this because I thought it added a bit of complexity in the code for the feature it added. But maybe this should be revisited. If Protesilaos thinks it is desirable to have this feature, I can recover what I did and make a PR for this.

protesilaos:
@lmq-10

> but I wanted to stop here and see if this is the right path to follow.

Reimplementing functionality will involve a lot of work and is prone to errors. To me this is the test if we go with the idea of `denote-extra-directories` or what @jeanphilippegg says about making `denote-directory` also accept a list.

At any rate, I think this is a good idea.

jeanphilippegg:
Ok, I will dig up the code I had and keep you posted.

jeanphilippegg:
See #609 to allow `denote-directory` to be a list of directories. You can test that `denote-grep` and related commands work.

PR #613, opened 2025-06-09
Make the date and identifiers independent and new identifier prompt

jeanphilippegg:
Hello Prot,

In this pull request, I make the date and identifier independent at
the core. The default is still unchanged, ie the identifier is derived
from the date.

There is a new identifier prompt. For now, the format of the string
should be the one we know (20240101T010101). Users do not need to use
this new prompt, though it is there.

Most users should not notice much changes. The only thing is that
users that have started to modify their identifiers should probably
now use `denote-rename-file-identifier` instead of
`denote-rename-file-date`. The latter would only change the date in
the front matter (as the date component is technically not in the file
name anymore).

Also, the date and identifier in the front matter may not always be
the same, which is probably a good thing, in fact.

-------

You may notice that we should now be very close to supporting
arbitrary identifiers. You can already experiment with it by removing
the restriction in `denote--creation-prepare-note-data` and evaluate
`(setq denote--identifier-generation-function #'denote-generate-identifier-as-number)`
and your identifiers will be simple increasing numbers. Everything
should work, except links, which I need to review.

### DIFF
--- a/README.org
+++ b/README.org
@@ -430,6 +430,10 @@ of the following:
 
   [[#h:e7ef08d6-af1b-4ab3-bb00-494a653e6d63][The denote-date-prompt-use-org-read-date option]].
 
+- =identifier=: Prompts with completion for the identifier of the new
+  note. It expects a string that has the format of
+  ~denote-date-identifier-format~.
+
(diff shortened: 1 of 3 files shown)


protesilaos:
Merged, thank you!

PR #617, opened 2025-06-20
Fix denote--rename-file for files without an identifier

### LINKED ISSUE #616: denote-rename-file no longer working on files without a Denote Time stamp
I used to be able to rename files that don't have Denote front matter (e.g. PDF or xls files using the denote-rename-file function. It would add a time stamp with the file's creation date and prompt me for title, keywords etc.

E.g. calling denote-rename-file on Document.PDF would give something like 20250617T213211--Document.PDF assuming I didn't specify any keywords and left the title unchanged.

However, when I run the denote-rename-file function now, I always get "denote-retrieve-filename-identifier-with-error: Cannot find <path to>/Document.PDF’ as a file with a Denote identifier" in the Message buffer

I can work around this by manually renaming Document.PDF with a fake time stamp, e.g. "20250101T000000--Document.PDF", I can then use denote-rename-file to change the title, add keywords etc. 

However, I still can't use denote-rename-file-date to change the date to anything other than what I manually assigned.

I can't pinpoint exactly when the denote-rename-file function stopped working for me, but it must have been within the last 2 weeks or so

jeanphilippegg:
This fixes #616.

Later, I will also make the other change I mentioned about `denote-get-backlinks`. I will probably just remove `denote-retrieve-filename-identifier-with-error` completely. This can be done later.

On an unrelated note, I think commit 9549d97 should be reverted. The functions `denote-generate-identifier-as-{date/number}` are not user commands and they are also not meant to be called from lisp directly. Even `denote-get-identifier-function` should probably not be used in packages. They are example identifier generation functions and users have to make a function with 2 parameters if they create their own.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -4081,7 +4081,7 @@ Respect `denote-rename-confirmations', `denote-save-buffers' and
                        identifier))
          (identifier (cond ((string-empty-p identifier) identifier)
                            ((string= old-identifier identifier) identifier)
-                           ((denote--file-has-backlinks-p file)
+                           ((and (not (string-empty-p old-identifier)) (denote--file-has-backlinks-p file))
                             (user-error "The identifier cannot be modified because the new identifier has backlinks"))
                            (t (funcall denote-get-identifier-function identifier nil))))
          (new-name (denote-format-file-name directory identifier keywords title extension signature))


protesilaos:
> This fixes #616.

Merged, thank you!

> Later, I will also make the other change I mentioned about `denote-get-backlinks`. I will probably just remove `denote-retrieve-filename-identifier-with-error` completely. This can be done later.

Very well!

> On an unrelated note, I think commit [9549d97](https://github.com/protesilaos/denote/commit/9549d977d5173243f5f25ccdf5051c9aa9d613ad) should be reverted. The functions `denote-generate-identifier-as-{date/number}` are not user commands and they are also not meant to be called from lisp directly. Even `denote-get-identifier-function` should probably not be used in packages. They are example identifier generation functions and users have to make a function with 2 parameters if they create their own.

Then I misunderstood the intent. Maybe we should remove them from the code base altogether then? Instead, have a section in the manual which explains what advanced users need to do to get custom identifiers.

jeanphilippegg:
> Then I misunderstood the intent. Maybe we should remove them from the code base altogether then? Instead, have a section in the manual which explains what advanced users need to do to get custom identifiers.

Users who want to customize their identifiers will have to create a function and assign it to `denote-get-identifier-function`, like this:

```emacs-lisp
(defun my-denote-generate-identifier-as-letter (identifier _date)
   ;; custom code)

(setq denote-get-identifier-function #'my-denote-generate-identifier-as-letter)
```
The 2 parameters (identifier and date) always have to be part of the function definition, even if unused. In commit 9549d97, you added a `&optional` in `denote-generate-identifier-as-{date/number}, but I think it is not necessary because only Denote's code should call `(funcall denote-get-identifier-function id date)` in the end. We will also never omit a parameter, when called, because we want it to work with custom functions made by users and we don't want them to worry about declaring their parameters `&optional`.

`denote-generate-identifier-as-{date/number} are good examples for users to create their own. `denote-generate-identifier-as-date` is the default.

PR #619, opened 2025-06-26
Take date into account when renaming a file without identifier

brabalan:
`denote-rename-file` no longer uses file modification time when called from lisp with no identifier: the last argument when calling `denote-get-identifier-function` is `nil` instead of the `date` that was computed earlier


jeanphilippegg:
The identifier parameter is new and it is mandatory. It is always a string (ex "20250101T000000"). `nil` is not a valid value, though it does not throw an error. To get the behavior that you want, you should use the empty string (""), like this:

```emacs-lisp (denote-rename-file "your-file" "your-title" nil "your-sig" nil "") ```

Issue #533, opened 2025-01-17
Should we have a database?


lmq-10:
[I know this would be better as a discussion, but I guess more people will see it if I make it an issue instead]

Org Roam, arguably the most popular note taking program developed for Emacs, uses a SQLite database to cache some data about notes. This significantly speeds up operations such as linking and jumping to notes.

Denote addresses this by having all relevant metadata imbued in file names. This way, there's no need to parse the files themselves, but only a single list of file names. Which turns out to be really fast, that being the reason why it scalates so well, and why it can work just fine without caching content.

However, there are some areas where this approach may be insufficient, and several issues in this repo seems to point that. For instance, we have situations where users want to use the real titles somewhere (#440, #527) or expand words into links using available titles (#508). 

So, the question naturally arises: should Denote keep a cache using a SQLite database?

That, of course, would require a lot of work. I wrote a proof-of-concept extension (not finished nor polished, but it works):

- https://github.com/lmq-10/denote-db

It's mainly for showing that implementing a database is doable. I don't think it is a great implementation on my side. Though I plan to keep working on it.

So, I just leave the question here, to discuss or drop ideas!

protesilaos:
> From: Lucas Quintana ***@***.***>
> Date: Fri, 17 Jan 2025 12:14:08 -0800
>
> [I know this would be better as a discussion, but I guess more people
> will see it if I make it an issue instead]

I prefer it here, anyway. The "discussions" tab is hard to find.

> [... 15 lines elided]
>
> So, the question naturally arises: should Denote keep a cache using a
> SQLite database?

In principle "yes". It has to (i) be optional and (ii) not enabled by default.

> That, of course, would require a lot of work. I wrote a
> proof-of-concept extension (not finished nor polished, but it works):
>
> - https://github.com/lmq-10/denote-db

Well done!

> It's mainly for showing that implementing a database is doable. I
> don't think it is a great implementation on my side. Though I plan to
> keep working on it.
>
> So, I just leave the question here, to discuss or drop ideas!

For the database to be useful at all times, I imagine you will want to
keep track of (i) files, (ii) headings, (iii) links, (iv) backlinks.
With the exception of headings, this data is independent of the file
type and we already have the means to retrieve it.

-- 
Protesilaos Stavrou
https://protesilaos.com

lmq-10:
> In principle "yes". It has to (i) be optional and (ii) not enabled by default.

[...]

PR #624, opened 2025-07-13
Custom identifier formats

jeanphilippegg:
Hello Prot!

In this pull request, I made the final changes to allow custom
identifier formats.

For your info (and anyone interested), the challenge was to find the
backlinks correctly. I needed to be able to retrieve links according
to the format of each supported file type. For example, for org, I
need to look for "[denote:some-identifier]" instead of only
"some-identifier". This is because if an id is "11", we cannot just
look for "11". We also cannot just look for "denote:11" because it
will also match "denote:11111".

I did the requisite changes. In fact, I also created the function
`denote--get-all-backlinks` that may be useful eventually. Right now,
each call to `denote-backlinks` reads all notes. With
`denote--get-all-backlinks`, we scan all files once and we have ALL
backlinks, for each identifier. We may want to cache this.

### Important notes

- Greping is not done by Xref anymore. Instead, it is done by calling
  the existing `denote-link--collect-identifiers` (which should
  probably be renamed) on each note. This may be slower than before
  depending on how xref is working. If there are performance issues,
  this PR is easy to revert. So it is probably best to wait a little
  before telling users that identifiers can have any format (or you
  may want to keep commit f704b49 for later). I still use Xref to
  display the final results, so things should still look as before.

- `denote-backlinks` does not use `denote-query-buffer-function`
  anymore. We cannot just grep for the identifier if identifiers can
  be anything, as explained above.

- Only supported file types will appear in backlinks buffer. This is
  probably acceptable and as intended.

- A note can now link to itself and should appear in the backlinks
  buffer.

- `denote-grep/query` should be unaffected by these changes.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -3350,10 +3350,6 @@ instead of that of the parameter."
                        template
                      (or (alist-get template denote-templates) "")))
          (signature (or signature "")))
-    ;; TODO: Remove this when ready to allow custom identifiers.
-    (unless (and (string-match-p denote-date-identifier-regexp identifier)
-                 (date-to-time identifier))
-      (user-error "The identifier must have the format of `denote-date-identifier-format'"))
     (list title keywords file-type directory date identifier template signature)))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you!

> I did the requisite changes. In fact, I also created the function
> `denote--get-all-backlinks` that may be useful eventually. Right now,
> each call to `denote-backlinks` reads all notes. With
> `denote--get-all-backlinks`, we scan all files once and we have ALL
> backlinks, for each identifier. We may want to cache this.

Caching makes sense. I suspect we will need to handle that state though. How does a database fit into that consideration? If it is an opt-in feature, ideally a separate package, then I think it is worth developing it to its fullest potential. There is this issue on the matter, as well as the `denote-db` package:

- <https://github.com/protesilaos/denote/issues/533>
- <https://github.com/lmq-10/denote-db>

PR #632, opened 2025-08-16
Various precisions in the manual

### LINKED ISSUE #586: File-naming scheme specification + Allow custom ID formats
Hello Prot!

Below is a specification of the file naming scheme of Denote that
includes all the fundamental rules that I can think of. It includes
generalized identifiers, that I discuss after. It also does not
mention sluggification, which is optional (not a rule).

It may be useful to include it in the manual eventually.

### The file naming scheme specification

A Denote file name has this form:

```
@@ID==SIGNATURE--TITLE__KEYWORD1_KEYWORD2.EXT
```

The rules (enforced by Denote commands):

- The delimiter characters are: `-`, `_`, `=`, and `@`.

- File name components are: identifier, signature, title, keywords.

- All components are optional (including the ID).

- All components can appear in any order.

- All components can contain any character, with these restrictions:

  - The dot character is not allowed in any component.

  - Consecutive sequences of the same delimiter character are not
    allowed. For example, `===` is not allowed in any component. One
    exception is that `---` (consecutive hyphens) is allowed in the
    title component.

  - `_` is not allowed in individual keywords, because it is used as
    the keywords separator.

  - The delimiter characters cannot appear at the end of any
    component.

  - The identifier component cannot contain parenthesis, square
    brackets, "query-contents:", or "query-filenames:".

- The delimiters are always present, except for the identifier
  component when it is in first position and it has the specific
  format "00000000T000000".

Notes for developers:

- Packages should not assume a specific order of the components. The
  notes "20240101T111111--note1.org" and
  "--note2@@20240101T222222.org" can both exist and a package
  should recognized them both. The current value of
  `denote-file-name-components-order` should not be considered by
  packages.

- Packages should not assume a specific format for the identifier or
  they should mention it as a limitation if they do. Also, packages
  should be able to work with all identifier formats, ie the current
  value of `denote-get-identifier-function` should not be checked.

### Custom identifiers

As you can see from the above specification, my intention would be to
make the identifier more general and allow it to be any string. I have
looked at the code and I think it is doable.

The idea is for advanced users to be able to customize the variable
`denote-identifier-generation-function`. The default would remain what
we know, ie identifiers would be automatically generated from the
date.

Advanced users would have to handle uniqueness of IDs (ID collisions)
themselves in their custom function and we should recommend them to
avoid using fancy characters in their identifiers. We know that square
brackets would break links in org notes, but we cannot know about
unknown file types.

The date and identifier would become separate concepts everywhere. The
default would remain to use the date to generate the id, but it would
not be mandatory.

What do you think?

jeanphilippegg:
Hello Prot!

I have updated the manual to be more precise about the file-naming
scheme and to make it acknowledge the latest development about
identifiers.

While reviewing the manual, I thought it seemed strange that an
example file produced by Denote is not shown until after all creation
and renaming commands are explained. So I added a couple lines in the
Overview section.

This would close #586 for me. Let me know if you want to include a
section in the manual about custom identifier formats or you prefer
keeping it on GitHub as documentation. I am fine with either.

Also, in case you missed it, I also replied to your question
[here](https://github.com/protesilaos/denote/commit/947f0be070c686f2b253498d6eb65ac2abcaee8e).

### DIFF
--- a/README.org
+++ b/README.org
@@ -254,6 +254,10 @@ other metadata.  Denote basically streamlines the creation of such
 files or file names while providing facilities to link between them
 (where those files are editable).
 
+A typical note produced by Denote looks like:
+
+: 20240322T131856--some-title__topic1_topic2.org
+
 Denote's file-naming scheme is not limited to "notes".  It can be used
 for all types of file, including those that are not editable in Emacs,
 such as videos.  Naming files in a consistent way makes their
@@ -2045,6 +2049,8 @@ minibuffer prompts:
 :CUSTOM_ID: h:4e9c7512-84dc-4dfb-9fa9-e15d51178e5d
 :END:
 
+[Customization of the identifier format is part of {{{development-version}}}.]
+
 #+vindex: denote-directory
 Notes are stored in the ~denote-directory~.  The default path is
 =~/Documents/notes=.  The ~denote-directory~ can be a flat listing,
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you @jeanphilippegg! Mergin now. And yes, I did miss that reply. It seems some notifications from GitHub never reach my inbox. I should get in the habit of checking GitHub directly. At any rate, I get your point and will deprecate `(denote-directory)` now.

jeanphilippegg:
> And yes, I did miss that reply. [... ] At any rate, I get your point and will deprecate `(denote-directory)` now.

No worries! I saw you deprecated it for 4.2.0. Isn't the next version going to be 4.1.0 instead?

PR #637, opened 2025-10-05
Check in denote--file-has-backlinks-p whether link matches file id

dizzyyogi:


protesilaos:
Merged, thank you @dizzyyogi!

protesilaos:
Looking at it again, I wonder if we can get away with something like this:

```elisp
(defun denote--file-has-backlinks-p (file)
  "Return non-nil if FILE has backlinks."
  (when-let* ((id (denote-retrieve-filename-identifier file))
              (files (denote-directory-files nil :omit-current :text-only)))
    (catch 'has-backlinks
      (dolist (file files)
        (with-temp-buffer
          (insert-file-contents file)
          (goto-char (point-min))
          (while (search-forward (format "[denote:%s]" id) nil t)
            (throw 'has-backlinks t)))))))
```

dizzyyogi:
Hi Prot, 
Sure, the change from your last comment is indeed simpler and clearer. The "while" could even be replaced by a "when", to avoid searching for more  backlinks once the one matching id has been found.

protesilaos:
Even better @dizzyyogi! I just pushed the change. Let's see if it works as intended.

PR #658, opened 2026-01-01
Fix trimming of Org links in `denote-link-ol-follow'

alvmts:
The current regexp only checks for a single colon, whereas search options in Org links are always followed by a double colon.

This is specially problematic in the Android port of Emacs, where external directories accessible to the Emacs app (see ```(info "(emacs) Android Document Providers")```) have paths like
```
/content/storage/com.android.externalstorage.documents/primary:emacs/denote
```
which get truncated by this function. If ```denote-directory``` is in one these directories, it makes it impossible to open links to other notes.


protesilaos:
Merged, thank you!

PR #661, opened 2026-01-05
Fix #654

### LINKED ISSUE #654: Consider removing `directories` parameter of `denote-directory-get-files`
Hello Prot!

You can check your commit message in c0df642 and the referenced issue where you decided to revert this `directories` parameters that you had previously added in March.

In short, there is no need for a `directories` parameter because users and package authors can already let-bind `denote-directory` to achieve the same result.

jeanphilippegg:
Hi Prot!

In version 4.1.0, you made `denote--directory-get-files` a public function. However, the intended public function was `(denote-directory-files)`, that does the same thing. So I made it private again for the next version. If other Denote packages started using the duplicate function, they should be fixed.

I have removed the `directories` parameter from some functions. As explained in #654, users and packages can let-bind `denote-directories` if necessary. I have added the more general function `denote--get-file-name-relative-to-directories`. This one takes a `directories` parameter and should be used to not recompute `(denote-directories)` for each note, as was done before.

You will need to revert the last 2 commits of the `denote-journal` package.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1248,18 +1248,19 @@ what remains."
   (and (file-writable-p file)
        (denote-file-has-supported-extension-p file)))
 
-(defun denote-get-file-name-relative-to-denote-directory (file &optional directories)
-  "Return FILE relative to the variable `denote-directory'.
-With optional DIRECTORIES, as a list of directories, return FILE
-relative to whichever one of those it belongs to."
+(defun denote--get-file-name-relative-to-directories (file directories)
+  "Return FILE relative to one of the DIRECTORIES."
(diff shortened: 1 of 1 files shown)


jeanphilippegg:
> Does it avoid the issue where `denote-directories` was called multiple times when its value should already be a given? I do not remember the exact scenario I was considering at the time. This is a good reminder for me to write more tests.

Calling it once or twice per command should not be an issue. However, Denote was calling it once for *each note* when `denote-get-file-name-relative-to-denote-directory` was used, which is now fixed.

Issue #662, opened 2026-01-05
denote-sort-dired errors if regexp matches a subdirectory of denote-directory


aurtzy:
Hello! I believe I may have discovered a bug with `denote-dired` that occurs when the regexp exactly matches a subdirectory in `denote-directory` where notes are nested.

### Reproduction steps

I've tested this on master commit cf2a6b082b0b792cb52b54fd3642c48b805ff071, Emacs version `GNU Emacs 30.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.51, cairo version 1.18.4)`.

1. Run `emacs -Q`.
2. Evaluate the code below.
3. Run `reproduce-error`, and see that an error occurs with the following message: `dired-insert-directory: Reading directory: No such file or directory, subdir/20260101T000000--title`

<details>

<summary>code</summary>

```emacs-lisp
(require 'denote)

(defun reproduce-setup ()
  (mkdir "/tmp/denote-test/subdir" t)
  (shell-command
   (concat
    "touch " (denote-format-file-name
              "/tmp/denote-test/subdir/" "20260101T000000" nil "title" nil nil))))

(defun reproduce (regexp)
  (interactive "MEnter regexp: ")
  (reproduce-setup)
  (let* ((default-directory "/tmp/denote-test/")
         (denote-directory default-directory))
    (denote-dired regexp 'title nil nil)))

(defun reproduce-error ()
  (interactive)
  (reproduce "subdir"))
```

</details>

### Some diagnosis

Some experimenting seems to point to the `dired` call. If I'm understanding correctly, `dired` is interpreting `DIRED-NAME` as a subdirectory name to start from, so it expects the files to be relative to that subdirectory and fails due to trying to find `/tmp/denote-test/subdir/subdir/20260101T000000--title`. If there is no matched subdirectory, it appears to fall back to `denote-directory` so the error doesn't normally happen.

I included a few more tests from my testing below, including some that isolate the error to a call to `dired`, in case they are of interest.

<details>

<summary>code</summary>

```emacs-lisp
(defun reproduce-no-error ()
  (interactive)
  (reproduce "subdi"))

(defun reproduce-with-dired-error ()
  (interactive)
  (reproduce-setup)
  (let ((default-directory "/tmp/denote-test/"))
    (dired '("subdir" "subdir/20260101T000000--title"))))

(defun reproduce-with-dired-no-error ()
  (interactive)
  (reproduce-setup)
[...]

PR #664, opened 2026-01-08
Fix error in denote-sort-dired when regexp is a subdirectory

aurtzy:
In the DIRNAME list argument to `dired`, the first element is treated as a
directory name, with the rest being file paths relative to that directory.
`dired` appears to fall back to the parent directory when DIRED-NAME does not
match any subdirectory, resulting in expected correct behavior; however, when
there is a matched directory, it will attempt to search for files relative to
that directory instead, which may result in file-not-found errors.

Thus, we remove and replace DIRED-NAME such that we always use the directory
that the files are supposed to be relative to.

Fixes: #662

---

Not sure if the README changes count as part of the total for legally-significant contributions, but just in case: I have already completed copyright assignment paperwork :)

### DIFF
--- a/README.org
+++ b/README.org
@@ -4635,9 +4635,8 @@ Empty files are those that satisfy `my-denote--note-has-no-contents-p'."
                              (car (denote-directories))
                            (denote-directories-get-common-root)))
               (files (funcall files-fn))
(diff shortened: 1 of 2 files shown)


protesilaos:
Merged, thank you!

PR #666, opened 2026-01-14
Refactor denote-sort-dired and remove the helper function

kilesduli:
This function is extremely slow because calling (dired (cons directory files)) triggers ls for every file. When there are a large number of files, The `call-process' becomes expensive. There is no other way to change this behavior in dired. We should use Emacs’s built-in ls implementation (ls-lisp). On my machine, with around 200 files, this reduces execution time by about 6x.

Additionally, prepare-buffer seems unnecessary. It is a private function (indicated by the double dash) and is not intended for user access or other use cases. It calls files-fn twice, which is redundant. We should instead set the revert-function directly in a single function to make its purpose explicit.

* (denote-sort-dired): refactor
* (denote-sort-dired--prepare-buffer): remove

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1938,30 +1938,6 @@ If REVERSE is nil, use the value of the user option
             "\n"
             (propertize "No more matching files" 'face 'warning))))
 
-(defun denote-sort-dired--prepare-buffer (directory files-fn buffer-name)
-  "Prepare buffer for `denote-sort-dired'.
-DIRECTORY is an absolute path to the `default-directory' of the Dired
-listing.
-
-FILES-FN is a function that returns the files to be listed in the Dired
-buffer.
-
-BUFFER-NAME is the name of the resulting buffer."
-  (let* ((default-directory directory)
-         (files (funcall files-fn))
-         (dired-buffer (dired (cons directory files))))
-    (with-current-buffer dired-buffer
-      (rename-buffer buffer-name :unique)
-      (setq-local revert-buffer-function
-                  (lambda (&rest _)
-                    (if-let* ((default-directory directory)
-                              (files (funcall files-fn)))
-                        (progn
-                          (setq-local dired-directory (cons directory files))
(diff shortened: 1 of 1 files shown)


denote.el
```diff
@@ -1938,30 +1938,6 @@ If REVERSE is nil, use the value of the user option
...
-
-BUFFER-NAME is the name of the resulting buffer."
-  (let* ((default-directory directory)
-         (files (funcall files-fn))
-         (dired-buffer (dired (cons directory files))))
-    (with-current-buffer dired-buffer
-      (rename-buffer buffer-name :unique)
-      (setq-local revert-buffer-function
-                  (lambda (&rest _)
-                    (if-let* ((default-directory directory)
```
kilesduli: We should not set `default-directory` here. `denote-revert` doesn't handle this; only `denote-dired` does. This may cause revert to fail, which I will describe below.

protesilaos:
Thank you @kilesduli! I will need to review this more closely. In the meantime, can you please tell me what you used to produce [the number of calls you demonstrate here](https://github.com/protesilaos/denote/pull/666#issuecomment-3748515113)?

kilesduli:
> Thank you @kilesduli! I will need to review this more closely. In the meantime, can you please tell me what you used to produce [the number of calls you demonstrate here](https://github.com/protesilaos/denote/pull/666#issuecomment-3748515113)?

I should upload this picture. My mistake.
image

The way dired retrieves file information is simply by calling insert-directory (which is a wrapper around ls). This works well for directories because it doesn't take much time. However, when we provide dired with a large number of files, as in the case of denote, dired ends up calling it for each file individually. In this scenario, switching to Emacs' built-in ls-lisp (a library used in environments like Windows that don't have an ls executable file) can actually speed things up, because call-process is too expensive in this context.

image
image

kilesduli:
You could use debug-on-entry and elp-instrument-function to verify this.

Issue #652, opened 2025-12-01
denote opens file three times


oliverepper:
Hi Prot!

I've noticed another little annoyance in the otherwise fantastic denote package:

I have a few notes that are encrypted via gpg. When I open them I need to enter the password to my keychain three times! (I don't cache the password at all, and the password actually unlocks the keychain which then enables access to my gpg passphrase.) Anyways this happens only with files that have the denote naming scheme and immediately goes away when I evaluate:

```elisp
(remove-hook 'find-file-hook #'denote-rename-buffer-rename-function-or-fallback)
```

After removing this from the find-file-hook entering my password once is sufficient.

So my guess is, that all of this has nothing to do with GPG at all, but with the simple fact that denote actually opens the file three times. Is that correct?

I might cache my passphrase for a few seconds to prevent the annoyance but I wanted to check in with you in order to understand what is going on.

Thanks a lot
Oliver

protesilaos:
Hello @oliverepper!

What is the value of the `denote-rename-buffer-format`? I suspect the issue is with `denote-rename-buffer--format`, which reads the file contents to get the file title and to check for backlinks. We should only do this using the current buffer, rather than call the functions which operate with the file contents in another buffer.

protesilaos:
Thank you @oliverepper! I have not been able to work on this due to my limited electricity these past days. I hope to do it soon. It requires a few changes.

oliverepper:
Take your time. I love denote. It's very valuable to me. Thanks for creating such a wonderful tool. I really like how much it tries to stay out of my way and not force me to do things a certain way.

PR #669, opened 2026-01-14
Fix denote--define-retrieve-front-matter not reusing opened buffer

kilesduli:
In earlier commits, denote-rename-buffer could open the same file multiple times when retrieving front matter, as described in https://github.com/protesilaos/denote/issues/652 .

This occurred because the subroutine denote--file-with-temp-buffer-subr, used in the definition of denote--define-retrieve-front-matter, contained incorrect conditional logic. As a result, files were reopened instead of reusing already opened buffers.

Fixes: https://github.com/protesilaos/denote/issues/652

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -2645,14 +2645,13 @@ Subroutine of `denote--file-with-temp-buffer'."
          (file-exists (file-exists-p file))
          (buffer-modified (buffer-modified-p buffer)))
     (cond
-     ((or (and file-exists
-               buffer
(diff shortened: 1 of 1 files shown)


protesilaos:
Thank you @kilesduli! This looks good. I notice you force pushed a change. I will wait a little bit longer before merging, in case you want to make any final tweaks. No pressure!

kilesduli:
Oops, you noticed me. Now you can merge.

PR #671, opened 2026-01-17
Fix docstring of denote-query-extract-title

mbatson:
Docstring says function returns `denote-grep-untitled-string`, but it actually returns `denote-query-untitled-string`. This commit corrects the name of the variable in the docstring.


protesilaos:
You are welcome!

Issue #670, opened 2026-01-16
Commit 8a88fcd breaks some renaming commands


jeanphilippegg:
Commit 8a88fcd causes issues with new unsaved notes.

Here is how to reproduce:

1. Create a new note and do not save it yet.
2. Modify the front matter.
3. Call `denote-rename-file-using-front-matter`.

Denote is now unable to retrieve the front matter and suggest a bad file name.

Maybe we should revert it for now. It was trying to fix #652. As I understand the original issue, it looks like front matter retrieval calls `insert-file-contents` multiple times instead of `insert-buffer`. This should only be possible if the file is `modified` or it is not viewed as a buffer by Emacs somehow.

Can you help me reproduce it, ie share the steps that lead to `insert-file-contents` being called multiple times in one command? Maybe I can help fix it? You can evaluate this modified `denote--file-with-temp-buffer-subr` for debugging:

```emacs-lisp
(defun denote--file-with-temp-buffer-subr (file)
  "Return path to FILE or its buffer together with the appropriate function.
Subroutine of `denote--file-with-temp-buffer'."
  (let* ((buffer (get-file-buffer file))
         (file-exists (file-exists-p file))
         (buffer-modified (buffer-modified-p buffer)))
    (cond
     ((or (and file-exists
               buffer
               (not buffer-modified)
               (not (eq buffer-modified 'autosaved)))
          (and file-exists (not buffer)))
      (cons #'insert-file-contents file)
      (print "insert-file-contents called"))
     (buffer
      (cons #'insert-buffer buffer))
     ;; (t
     ;;  (error "Cannot find anything about file `%s'" file))
     )))
```

If we are unable to get the message "insert-file-contents called" multiple times in one command, maybe the issue is in another function.

kilesduli:
Reproduced. I'll submit the fix later.

PR #672, opened 2026-01-17
Fix front-matter retrieving when buffer is unsaved or file missing

kilesduli:
This addresses a regression where
'denote-rename-file-using-front-matter' would fail to recognize changes
in an unsaved buffer, or fail entirely if the file had not yet been
created on disk (as described in
https://github.com/protesilaos/denote/pull/670).

The logic now consistently prioritizes the content of an open buffer
over the file on disk. Previously, the code would revert to reading the
file if the buffer was marked as modified; however, since we cannot
reliably detect if a modification affects the front-matter, it is safe
to always use the live buffer state.

* (denote--file-with-temp-buffer-subr): Remove.
(denote--file-with-temp-buffer): Refactor to prioritize buffer content
and remove the dependency on the internal subr.

Fixes: https://github.com/protesilaos/denote/pull/670

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -2624,30 +2624,17 @@ Also see `denote-extract-keywords-from-path' (alias
     (when (string-match denote-title-regexp filename)
       (match-string 1 filename))))
 
-(defun denote--file-with-temp-buffer-subr (file)
-  "Return path to FILE or its buffer together with the appropriate function.
-Subroutine of `denote--file-with-temp-buffer'."
-  (let* ((buffer (get-file-buffer file))
-         (file-exists (file-exists-p file))
-         (buffer-modified (buffer-modified-p buffer)))
-    (cond
-     ((and file-exists
-           buffer
-           (not buffer-modified))
-      (cons #'insert-buffer buffer))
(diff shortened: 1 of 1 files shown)


kilesduli:
There are two corner cases that my GPG fix does not cover:
 - The file does not exist, but a buffer with the same name exists.
 - The buffer is modified but not saved.
Both will cause denote-rename-file-using-front-matter to fail.

protesilaos:
Merged, thank you folks!

kilesduli:
> > I can reproduce this locally. When the front matter is modified but not autosaved/saved, denote-rename-file-using-front-matter does not rename the file.
> > 
> > ```
> > M-: (buffer-modified-p (current-buffer))) ;; check this, should be t
> > ```
> 
> I still cannot reproduce.
> 
> This is what I do:
> 1. Create and new note and save it.
> 2. Add a few characters to the front matter and do not save. `buffer-modified-p` returns `t`.
> 3. Call `denote-rename-file-using-front-matter.
> 
> The file is renamed as expected, without failing.


That's fine, it's been fixed. :)

Issue #317, opened 2024-04-11
org-display-inline-images doesn't work for denote: links


newmedicine:
Liking denote very much. 

```
[[denote:20180214T132312]]

[[file:20180214T132312--p1010849__image.JPG]]
``` 

If I call `org-display-inline-images`, the second (`file:`) link is replaced by an image, but the first (`denote:`) is not:

![image](https://github.com/protesilaos/denote/assets/22443747/2cb118df-9680-44b6-9d20-26adea936a34)

Using denote 2.3.5 and emacs 29.3 with org 9.6.15.

Thank you!

Nick

protesilaos:
Oh, this is interesting! I will check how Org handles images and report back to you.

protesilaos:
This is a bug in Org, as it hardcodes what type of link can preview an image. I modified the relevant function just to prove the concept (I SIMPLIFIED IT AND IT SHOULD NOT BE USED AS-IS):

```elisp
(defun org-display-inline-images (&optional include-linked refresh beg end)
  "Display inline images.

An inline image is a link which follows either of these
conventions:

  1. Its path is a file with an extension matching return value
     from `image-file-name-regexp' and it has no contents.

  2. Its description consists in a single link of the previous
     type.  In this case, that link must be a well-formed plain
     or angle link, i.e., it must have an explicit \"file\" or
     \"attachment\" type.

Equip each image with the key-map `image-map'.

When optional argument INCLUDE-LINKED is non-nil, also links with
a text description part will be inlined.  This can be nice for
a quick look at those images, but it does not reflect what
exported files will look like.

When optional argument REFRESH is non-nil, refresh existing
images between BEG and END.  This will create new image displays
only if necessary.

BEG and END define the considered part.  They default to the
buffer boundaries with possible narrowing."
  (interactive "P")
  (when (display-graphic-p)
    (when refresh
      (org-remove-inline-images beg end)
      (when (fboundp 'clear-image-cache) (clear-image-cache)))
    (let ((end (or end (point-max))))
      (org-with-point-at (or beg (point-min))
        (let* ((case-fold-search t)
               (file-extension-re (image-file-name-regexp))
               (link-abbrevs (mapcar #'car
                                     (append org-link-abbrev-alist-local
                                             org-link-abbrev-alist)))
               ;; Check absolute, relative file names and explicit
               ;; "file:" links.  Also check link abbreviations since
               ;; some might expand to "file" links.
               (file-types-re
[...]

PR #683, opened 2026-02-24
Enable preview of denote links

swflint:
Preview will only happen if there is one file with the denote ID/path, and will work with any format supported by `org-link-preview-file`.
 
See also #317.

Note, this currently assumes IDs are at the start of the filename.  Because of complications around copyright assignment, I will be unable to fix this at this time.


protesilaos:
I just made the tweaks. It seems to be working the same as before.

Note that since the beginning the behaviour is not 100% right. If I do `org-link-preview` with the cursor over a `denote:` link to an image file I get a message that there can be no preview at point. But it works if I do `C-u 11 org-link-preview`.

PR #697, opened 2026-04-20
Macro for denote-query-exclude-files and denote-query-only-include-files

gnuhack:
Replaced both functions with a new macro. Created the new functions through the macro.


protesilaos:
Merged, thank you!

PR #708, opened 2026-05-23
denote-open-or-create errors in directories with oddly-named files (#707)

### LINKED ISSUE #707: helm-M-x-execute-command: Wrong type argument: stringp, nil
Hey @protesilaos, first of all thank you for this wonderful package! ❤️ 

With one of the latest versions of denote when I invoke `denote-open-or-create` I have this error message

```elisp
helm-M-x-execute-command: Wrong type argument: stringp, nil
```

The version is `Version: 20260521.190922`

and I am on the latest *spacemacs*

Do you think it's a problem of *spacemacs* or something in `denote`?

Thank you so much!

segv:
Calling `denote-open-or-create` (or any command that uses the file prompt) can error with `Wrong type argument: stringp, nil` when the `denote-directory` contains files whose names happen to be a bare denote separator followed by an extension: `__.js`, `--.org`, `==.org`, `@@.org`. (lodash ships `node_modules/lodash/fp/__.js`, so an accidental lodash checkout under a denote-directory is enough to trip it)

Reported in #707, where it surfaces under helm as `helm-M-x-execute-command: Wrong type argument: stringp, nil`.

The fix is two-fold:

1. **Tighten the field regexps.** `denote-{identifier,signature,title,keywords}-regexp` used a zero-or-more lazy capture (`[^.]*?`), so `__.js` and friends matched as denote files with an empty field. Switching to `[^.]+?` requires at least one character in the field, so these names stop being recognised as denote notes and stop showing up as file-prompt candidates.

2. **Don't assume the field is present** in `denote-file-prompt--format-identifier`. Even with the regex tightened, the function shouldn't feed nil to `propertize` if the file has no identifier — it now returns nil, which `denote-file-prompt-affixate` already handles via `(or … "")`.

Fixes #707.

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -1492,10 +1492,9 @@ there.")
 
 (defun denote-file-prompt--format-identifier (file)
   "Return identifier of FILE for `denote-file-prompt-affixate'."
-  (let* ((identifier (denote-retrieve-filename-identifier file))
-         (date-or-id (or (ignore-errors (denote-id-to-date identifier)) identifier))
-         (propertized (propertize date-or-id 'face 'completions-annotations)))
-    (format "%s " propertized)))
+  (when-let* ((identifier (denote-retrieve-filename-identifier file))
+              (date-or-id (or (ignore-errors (denote-id-to-date identifier)) identifier)))
+    (format "%s " (propertize date-or-id 'face 'completions-annotations))))
 
 (defun denote-file-prompt--format-keywords-and-signature (file)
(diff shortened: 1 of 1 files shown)


protesilaos:
Hello @segv and thank you for doing this!

The `when-let*` part is the relevant fix. The change to the regular expressions makes sense but I am not sure is essential here. Anyway, I will merge this and then can think about it.

PR #711, opened 2026-06-11
Add a section on denote-paperless.el in the documentation

matthieumuller:
Following up on our mail discussion, here is an addition to the manual for the `denote-paperless` package: https://codeberg.org/matthieumuller/denote-paperless.


protesilaos:
Merged, thank you!

PR #722, opened 2026-07-31
Support Unix timestamps in date input

### LINKED ISSUE #721: Support for unix timestamp in `denote-date-prompt`
Hello @protesilaos, and many thanks for Denote!

I sometimes find it convenient to paste/yank a Unix timestamp when creating a new note. `denote-date-prompt` doesn't presently support dates in this format, so I advise the function as follows:

```elisp
(defun my/denote-date-prompt-advice (f &rest args)
  "Apply F to ARGS such that `denote-date-prompt' handles Unix timestamps.
If the value read by the prompt is a Unix timestamp, parse it to the
format expected by Denote. Otherwise, return the value as it was read."
  (let ((date (apply f args)))
    (if (string-match "[0-9]\\{10\\}" date) ; seconds since epoch
        (let ((secs (string-to-number (match-string 0 date))))
          (format-time-string "%Y-%m-%d %H:%M:%S" (seconds-to-time secs)))
      date)))

(advice-add 'denote-date-prompt :around #'my/denote-date-prompt-advice)
```

Might Unix timestamps make sense to support in Denote's date prompt? I would be happy to contribute a patch if you like.

jeffvalk:
This allows for other recognized date patterns to be supported as well, should this ever be desirable.

Closes #721

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -3674,8 +3674,18 @@ here for clarity."
         ((eq prefer-type :string)
          (if date (format-time-string "%F %T" date) ""))))
 
+(defun denote--date-parse-string (date)
+  "Convert DATE to Denote's preferred string format if a recognized pattern.
+Otherwise, return DATE unchanged.  Currently supports Unix timestamps."
+  (cond ((string-match "[0-9]\\{10\\}" date) ; seconds since epoch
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you!

PR #723, opened 2026-08-18
Add option to use simple heading links to denote-link-ol-store.

mkorat:
This pull request adds the new option to `denote-org-store-link-to-heading` to allow storing file local org links that will omit the `denote` specific part of the link.

For example, setting `denote-org-store-link-to-heading` to `context`  will currently produce the following link for a given heading:

`[[denote:20260818T222356::*This is a heading][This is a heading]]`

Setting it to the new option `local` will produce the following link for the same heading:

`[[*This is a heading][This is a heading]]`

If a `CUSTOM_ID` is set for the headline, storing the link will produce:

`[[#test][This is a heading]]`

### DIFF
--- a/denote.el
+++ b/denote.el
@@ -485,6 +485,10 @@ and the text of the current heading, like this:
 
     [[denote:20240118T060608::*Heading text][Some test::Heading text]].
 
+If the value is `local', the link consists of the text of the current heading, like this:
+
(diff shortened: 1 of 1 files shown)


protesilaos:
Merged, thank you @mkorat!

PR #726, opened 2026-08-24
Move Unix timestamps code in `denote-valid-date-p`

jeanphilippegg:
Hello Prot!

I moved the code about Unix timestamps in `denote-valid-date-p` as I think it makes more sense there. I have not tested, but it should also make `denote-use-date` variable accept Unix timestamps as well.

Have a good day!


protesilaos:
Merged, thank you!

